Issue #2, opened 2019-02-21
Fix server header version warning


kinow:
The Hub seems to expect a version reported by the spawned application. If no header, it displays the following warning in the logs.

```
[W 2019-02-22 11:26:33.705 JupyterHub _version:30] Single-user server has no version header, which means it is likely < 0.8. Expected 0.9.4
```

This can be easily fixed, apparently, by simply returning the JupyterHub version. IOW, if you have JupyterHub 0.9.4, and return 0.9.4, the warning is gone.

If you return something like 1.0, then the application is unavailable in the hub, and you will have an error in the logs.

```
[W 2019-02-22 11:34:37.321 JupyterHub web:1667] 500 GET /hub/user/kinow/?redirects=2 (::1): Redirect loop detected. Notebook has jupyterhub version 1.0, but the Hub expects 0.9.4. Try installing jupyterhub==0.9.4 in the user environment if you continue to have problems.
[E 2019-02-22 11:34:37.369 JupyterHub log:150] {
      "X-Forwarded-Host": "localhost:8000",
      "X-Forwarded-Proto": "http",
      "X-Forwarded-Port": "8000",
      "X-Forwarded-For": "::1",
      "Cookie": "jupyterhub-hub-login=\"2|1:0|10:1550786007|20:jupyterhub-hub-login|44:MzZkNzljMmE5ZDUwNGFiNGEwMGJlNmQ0ZjJjOTNiNDI=|fa6ff62d3476c652147c10138317e3bb6e65f7ca59049d6f6310c6be144d20db\"; _xsrf=2|9c1bb992|e073c22bafb9cc9ce3d33fcd34dc3814|1550716134; jupyterhub-session-id=bb91133a002142d8a3268b927e0fe1c4",
      "Accept-Language": "en-GB,en-US;q=0.9,en;q=0.8",
      "Accept-Encoding": "gzip, deflate, br",
      "Referer": "http://localhost:8000/hub/login",
      "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8",
      "User-Agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.109 Safari/537.36",
      "Upgrade-Insecure-Requests": "1",
      "Cache-Control": "no-cache",
      "Pragma": "no-cache",
      "Connection": "close",
      "Host": "localhost:8000"
    }
[...]

PR #3, opened 2019-02-21
Fix version warning

kinow:
This PR creates a class `CylcUIServer` (may be changed later) to wrap the logic for the UI server application create, start, etc.

Also sets the X-JupyterHub-Version header matching the JupyterHub version. This should remove the warning in the logs, and make our app to look more like a normal Notebook application for the hub.


kinow:
Fixes #2

PR #20, opened 2019-04-12
Use Python 3.7

### LINKED ISSUE #19: Use Python 3.7
No need to stay behind Cylc 8, I believe we can start using 3.7 right now, as the 1.0 release is still a little far away.

kinow:
Close #19 

To keep up with Cylc 8.


kinow:
Thanks @dwsutherland ! And feel free to do the honours if you'd like :D otherwise will merge it later.

PR #34, opened 2019-04-24
WS-UIS data and GraphQL integration

dwsutherland:
The sibling pull-request to the UI Server graphql protobuf feed:
https://github.com/cylc/cylc/pull/3122
And the Cylc8 UI Server component implementation of the [GraphQL PoC](https://github.com/cylc/cylc/issues/2900).

There will be many more improvements/additions to be incorporated both in this and subsequent pull requests, however, for the sake of UI development and review this needs to be exposed.

The extra dependencies are:
```
    'pyzmq',
    'protobuf',
    'python-jose',
    'colorama',
```
(`colorama` due to the `cylc scan` CL call, which will be deprecated as workflow discovery is already integrated)

The instructions on installation are included in the README.md, and for this change cylc workflow library needs to be put into the python path (or venv site packages), I just source:
```
source ~/.envs/cylc-uis/bin/activate
source ~/.npm_env
export PYTHONPATH="${PYTHONPATH}:/home/sutherlander/repos/cylc8/lib"
```
With `.npm_env` being my local node.js environment:
```
# Node environment
export NPM_PACKAGES="${HOME}/.npm-packages"
export NODE_PATH="$NPM_PACKAGES/lib/node_modules:$NODE_PATH"
export PATH="$NPM_PACKAGES/bin:$PATH"

unset MANPATH
export MANPATH="$NPM_PACKAGES/share/man:$(manpath)"
```

There are three main files in this change:
`schema.py` - GraphQL data structure and types definition (uses [graphene](https://graphene-python.org/))
`workflow_services_mgr.py` - Workflow discovery, REQ client established (and passed to data manager), and spawning (to come).
`data_mgr.py` - Uses established client/workflows to retrieve and process data structure (decode protobuf), and contains access & filtering methods used by the resolvers.

In order to pass in the uiserver/data_manager as context to the resolver's I had to subclass the `graphene-tornado` GraphQL handler:
```
class UIServerGraphQLHandler(TornadoGraphQLHandler):

    #Extra attributes
    uiserver = None

    def initialize(self, **kwargs):
        super(TornadoGraphQLHandler, self).initialize()
        for key, value in kwargs.items():
            if hasattr(self, key):
                if key == 'middleware' and value is not None:
                    setattr(self, key,
                        list(self.instantiate_middleware(value)))
                else:
                    setattr(self, key, value)

    @property
    def context(self):
        wider_context = {
            'uiserver': self.uiserver,
            'request': self.request,
        }
        return wider_context
```

The data update method is very crude at the moment; it pulls the full workflow data msg every 5 secs (which should be faster than CLI `$ cylc client --no-input baz pb_entire_workflow` as the client is already established by the WS manager)... This message is decoded/parsed as a protobuf object (class instance) and stored in a dictionary with "owner/workflow" as a key. 
One of the good features of the UI Server holding/maintaining a data structure is, without needing multiple requests, we can query and filter across workflows, for example:
**query**
```
query workflows($tIds: [ID]){
  workflows(ids: ["baz", "jin"]){
    id
    name
    status
    stateTotals {
      runahead
      waiting
      held
      queued
      expired
      ready
      submitFailed
      submitRetrying
      submitted
      retrying
      running
      failed
      succeeded
    }
    tasks(ids: $tIds){
      name
      meta {
        title
        description
        URL
        userDefined
      }
      proxies(ids: $tIds){
        cyclePoint
        state
      }
    }
  }
}
```
**variables**
```
{
  "tIds": ["foo", "baz"]
}
```
**results**
```
{
  "data": {
    "workflows": [
      {
        "id": "sutherlander/baz",
        "name": "baz",
        "status": "held",
        "stateTotals": {
          "runahead": 0,
          "waiting": 0,
          "held": 5,
          "queued": 0,
          "expired": 0,
          "ready": 0,
          "submitFailed": 0,
          "submitRetrying": 0,
          "submitted": 0,
          "retrying": 0,
          "running": 0,
          "failed": 0,
          "succeeded": 19
        },
        "tasks": [
          {
            "name": "foo",
            "meta": {
              "title": "Some Top family",
              "description": "some task foo",
              "URL": "https://github.com/dwsutherland/cylc",
              "userDefined": [
                "importance=Critical",
                "alerts=none"
              ]
            },
            "proxies": [
              {
                "cyclePoint": "20170301T0000+12",
                "state": "succeeded"
              },
              {
                "cyclePoint": "20170401T0000+12",
                "state": "succeeded"
              },
              {
                "cyclePoint": "20170501T0000+12",
                "state": "held"
              }
            ]
          }
        ]
      },
      {
        "id": "sutherlander/jin",
        "name": "jin",
        "status": "held",
        "stateTotals": {
          "runahead": 0,
          "waiting": 0,
          "held": 1,
          "queued": 0,
          "expired": 0,
          "ready": 0,
          "submitFailed": 0,
          "submitRetrying": 0,
          "submitted": 0,
          "retrying": 0,
          "running": 0,
          "failed": 0,
          "succeeded": 0
        },
        "tasks": [
          {
            "name": "baz",
            "meta": {
              "title": "",
              "description": "",
              "URL": "",
              "userDefined": []
            },
            "proxies": [
              {
                "cyclePoint": "1",
                "state": "held"
              }
            ]
          }
        ]
      }
    ]
  }
}
```
This applies to all fields (from root query), i.e. tasks, jobs, familes, edges... This could open up some possibilities for us.. (And obviously gives us the data for the __"gscan"__ type web view)

A lot more to do, but this is a start 😃 

**This Pull Request**
- [x] Workflow services discovery, management, ZeroMQ REQ/RES client
- [x] Add Data management (protobuf message request & processing) and access methods
- [x] Write/Integrate PoC graphql schema to represent UIS context.
- [x] Query types functioning.
- [x] Mutations types functioning.
- [ ] ? Perhaps some of the below additional

**Additional** (to come or future PR?)
- [ ] Incremental updates to data structure using deltas and checksum skeleton.
- [ ] Deltas received via PUB/SUB from Workflow Services UI Server and checksum skeleton data management.
- [ ] Tornado WebSockets and Subscription server implementation
- [ ] Pagination
- [ ] Workflow services spawning
- [ ] Testing

### DIFF
--- a/cylc_singleuser.py
+++ b/cylc_singleuser.py
@@ -21,13 +21,16 @@ import logging
 import os
 import signal
 
-from graphene_tornado.schema import schema
+from schema import schema
 from graphene_tornado.tornado_graphql_handler import TornadoGraphQLHandler
+from graphql import get_default_backend
 from jupyterhub.services.auth import HubOAuthCallbackHandler
 from jupyterhub.utils import url_path_join
 from tornado import web, ioloop
 
 from handlers import *
+from workflow_services_mgr import WorkflowServicesMgr
+from data_mgr import DataManager
 
 
 class MyApplication(web.Application):
@@ -44,6 +47,42 @@ class MyApplication(web.Application):
             logging.info('exit success')
 
 
+# This is needed in order to pass the server context in addition to existing.
+# It's possible to just overwrite TornadoGraphQLHandler.context but we would
+# somehow need to pass the request info (headers, username ...etc) in also
+class UIServerGraphQLHandler(TornadoGraphQLHandler):
+
+    # Declare extra attributes
+    uiserver = None
+
+    def initialize(self, schema=None, executor=None, middleware=None,
+                   root_value=None, graphiql=False, pretty=False,
+                   batch=False, backend=None, **kwargs):
+        super(TornadoGraphQLHandler, self).initialize()
+
+        self.schema = schema
+        if middleware is not None:
+            self.middleware = list(self.instantiate_middleware(middleware))
+        self.executor = executor
+        self.root_value = root_value
+        self.pretty = pretty
+        self.graphiql = graphiql
+        self.batch = batch
+        self.backend = backend or get_default_backend()
+        # Set extra attributes
+        for key, value in kwargs.items():
+            if hasattr(self, key):
+                setattr(self, key, value)
+
+    @property
+    def context(self):
+        wider_context = {
+            'uiserver': self.uiserver,
+            'request': self.request,
+        }
+        return wider_context
+
+
 class CylcUIServer(object):
 
     def __init__(self, port, static, jupyter_hub_service_prefix):
@@ -54,6 +93,8 @@ class CylcUIServer(object):
             script_dir = os.path.dirname(__file__)
             self._static = os.path.abspath(os.path.join(script_dir, static))
         self._jupyter_hub_service_prefix = jupyter_hub_service_prefix
+        self.ws_mgr = WorkflowServicesMgr()
+        self.data_mgr = DataManager(self.ws_mgr)
 
     def _make_app(self):
         """Crete a Tornado web application."""
@@ -92,16 +133,16 @@ class CylcUIServer(object):
                 # graphql
                 (url_path_join(self._jupyter_hub_service_prefix,
                                '/graphql'),
-                    TornadoGraphQLHandler,
-                    dict(graphiql=True, schema=schema)),
+                    UIServerGraphQLHandler,
+                    dict(schema=schema, uiserver=self)),
                 (url_path_join(self._jupyter_hub_service_prefix,
                                '/graphql/batch'),
-                    TornadoGraphQLHandler,
-                    dict(graphiql=True, schema=schema, batch=True)),
+                    UIServerGraphQLHandler,
+                    dict(schema=schema, uiserver=self, batch=True)),
                 (url_path_join(self._jupyter_hub_service_prefix,
                                '/graphql/graphiql'),
-                    TornadoGraphQLHandler,
-                    dict(graphiql=True, schema=schema))
+                    UIServerGraphQLHandler,
+                    dict(schema=schema, uiserver=self, graphiql=True)),
             ],
             # FIXME: decide (and document) whether cookies will be permanent
             # after server restart.
@@ -113,6 +154,14 @@ class CylcUIServer(object):
         signal.signal(signal.SIGINT, app.signal_handler)
         app.listen(self._port)
         ioloop.PeriodicCallback(app.try_exit, 100).start()
+        # Discover workflows on intiial start up.
+        ioloop.IOLoop.current().add_callback(self.ws_mgr.gather_workflows)
+        # Arbitrary intervals (msec) chosen for the two periodic callbacks
+        # below. These will likely change or be removed with PUB/SUB updates.
+        # Check for new workflows every 19sec.
+        ioloop.PeriodicCallback(self.ws_mgr.gather_workflows, 19000).start()
+        ioloop.PeriodicCallback(
+            self.data_mgr.entire_workflow_update, 5000).start()
         try:
             ioloop.IOLoop.current().start()
         except KeyboardInterrupt:
@@ -149,4 +198,4 @@ if __name__ == "__main__":
 
 
 __all__ = ['MainHandler', 'UserProfileHandler', 'MyApplication',
-           'CylcUIServer', 'main']
+           'UIServerGraphQLHandler', 'CylcUIServer', 'main']
--- a/cylc_singleuser.py
+++ b/cylc_singleuser.py
@@ -27,8 +27,9 @@ from jupyterhub.utils import url_path_join
 from tornado import web, ioloop
 
 from handlers import *
-from workflow_services_mgr import WorkflowServicesMgr
+from workflows_mgr import WorkflowsManager
 from data_mgr import DataManager
+from resolvers import Resolvers
 
 
 class MyApplication(web.Application):
@@ -55,8 +56,10 @@ class CylcUIServer(object):
             script_dir = os.path.dirname(__file__)
             self._static = os.path.abspath(os.path.join(script_dir, static))
         self._jupyter_hub_service_prefix = jupyter_hub_service_prefix
-        self.ws_mgr = WorkflowServicesMgr()
+        self.ws_mgr = WorkflowsManager()
         self.data_mgr = DataManager(self.ws_mgr)
+        self.resolvers = Resolvers(
+            self.ws_mgr, self.data_mgr)
 
     def _make_app(self):
         """Crete a Tornado web application."""
@@ -96,15 +99,16 @@ class CylcUIServer(object):
                 (url_path_join(self._jupyter_hub_service_prefix,
                                '/graphql'),
                     UIServerGraphQLHandler,
-                    dict(schema=schema, uiserver=self)),
+                    dict(schema=schema, resolvers=self.resolvers)),
                 (url_path_join(self._jupyter_hub_service_prefix,
                                '/graphql/batch'),
                     UIServerGraphQLHandler,
-                    dict(schema=schema, uiserver=self, batch=True)),
+                    dict(schema=schema, resolvers=self.resolvers, batch=True)),
                 (url_path_join(self._jupyter_hub_service_prefix,
                                '/graphql/graphiql'),
                     UIServerGraphQLHandler,
-                    dict(schema=schema, uiserver=self, graphiql=True)),
+                    dict(schema=schema, resolvers=self.resolvers,
+                         graphiql=True)),
             ],
             # FIXME: decide (and document) whether cookies will be permanent
             # after server restart.
--- a/cylc_singleuser.py
+++ b/cylc_singleuser.py
@@ -22,8 +22,6 @@ import os
 import signal
 
 from schema import schema
-from graphene_tornado.tornado_graphql_handler import TornadoGraphQLHandler
-from graphql import get_default_backend
 from jupyterhub.services.auth import HubOAuthCallbackHandler
 from jupyterhub.utils import url_path_join
 from tornado import web, ioloop
@@ -45,42 +43,6 @@ class MyApplication(web.Application):
             # clean up here
             ioloop.IOLoop.instance().stop()
             logging.info('exit success')
-
-
-# This is needed in order to pass the server context in addition to existing.
-# It's possible to just overwrite TornadoGraphQLHandler.context but we would
-# somehow need to pass the request info (headers, username ...etc) in also
-class UIServerGraphQLHandler(TornadoGraphQLHandler):
-
-    # Declare extra attributes
-    uiserver = None
-
-    def initialize(self, schema=None, executor=None, middleware=None,
-                   root_value=None, graphiql=False, pretty=False,
-                   batch=False, backend=None, **kwargs):
-        super(TornadoGraphQLHandler, self).initialize()
-
-        self.schema = schema
-        if middleware is not None:
-            self.middleware = list(self.instantiate_middleware(middleware))
-        self.executor = executor
-        self.root_value = root_value
-        self.pretty = pretty
-        self.graphiql = graphiql
-        self.batch = batch
-        self.backend = backend or get_default_backend()
-        # Set extra attributes
-        for key, value in kwargs.items():
-            if hasattr(self, key):
-                setattr(self, key, value)
-
-    @property
-    def context(self):
-        wider_context = {
-            'uiserver': self.uiserver,
-            'request': self.request,
-        }
-        return wider_context
 
 
 class CylcUIServer(object):
@@ -154,7 +116,7 @@ class CylcUIServer(object):
         signal.signal(signal.SIGINT, app.signal_handler)
         app.listen(self._port)
         ioloop.PeriodicCallback(app.try_exit, 100).start()
-        # Discover workflows on intiial start up.
+        # Discover workflows on initial start up.
         ioloop.IOLoop.current().add_callback(self.ws_mgr.gather_workflows)
         # Arbitrary intervals (msec) chosen for the two periodic callbacks
         # below. These will likely change or be removed with PUB/SUB updates.
@@ -198,4 +160,4 @@ if __name__ == "__main__":
 
 
 __all__ = ['MainHandler', 'UserProfileHandler', 'MyApplication',
-           'UIServerGraphQLHandler', 'CylcUIServer', 'main']
+           'CylcUIServer', 'main']
--- a/data_mgr.py
+++ b/data_mgr.py
@@ -17,16 +17,15 @@
 """Create and update the data structure for all workflow services."""
 
 import asyncio
-from fnmatch import fnmatchcase
 
 # from google.protobuf.json_format import MessageToDict
 
-from cylc.exceptions import ClientError, ClientTimeout
-from cylc.ws_messages_pb2 import PbEntireWorkflow
-from cylc.network.scan import MSG_TIMEOUT
+from cylc.flow.exceptions import ClientError, ClientTimeout
+from cylc.flow.ws_messages_pb2 import PbEntireWorkflow
+from cylc.flow.network.scan import MSG_TIMEOUT
 
 
-# TODO: define in and import this from cylc-flow
+# TODO: define in and import this from cylc.flow
 # maps methods in cylc-flow server to the returned bytes message
 METHOD_MAP = {
     'pb_entire_workflow': PbEntireWorkflow
@@ -74,170 +73,3 @@ class DataManager(object):
             self.data = new_data
         else:
             self.data.update(new_data)
-
-    # Data access filters
-    def get_workflow_msgs(self, args):
-        """Return list of workflows."""
-        result = []
-        for flow_msg in self.data.values():
-            if self._workflow_filter(flow_msg, args):
-                result.append(flow_msg)
-        # TODO: Sorting and Pagination
-        return result
-
-    def get_nodes_all(self, node_type, args):
-        """Return nodes from all workflows, filter by args."""
-        w_args = {
-            'ids': args['workflows'],
-            'exids': args['exworkflows'],
-        }
-        nodes = [
-            n for k in self.get_workflow_msgs(w_args)
-            for n in getattr(k, node_type)]
-        result = []
-        for node in nodes:
-            if self._node_filter(node, args):
-                result.append(node)
-        # TODO: Sorting and Pagination
-        return result
-
-    def get_nodes_by_id(self, node_type, args):
-        """Return protobuf node objects for given id."""
-        nat_ids = args.get('native_ids', [])
-        w_ids = []
-        for nat_id in nat_ids:
-            o_name, w_name, nid = nat_id.split('/', 2)
-            w_ids.append(f"{o_name}/{w_name}")
-        if node_type == 'proxy_nodes':
-            nodes = (
-                n for w_id in set(w_ids)
-                for n in (
-                    list(getattr(self.data[w_id], 'task_proxies', []))
-                    + list(getattr(self.data[w_id], 'family_proxies', []))))
-        else:
-            nodes = (
-                n for w_id in set(w_ids)
-                for n in getattr(self.data[w_id], node_type, []))
-        result = []
-        for node in nodes:
-            if ((not nat_ids or node.id in set(nat_ids)) and
-                    self._node_filter(node, args)):
-                result.append(node)
-        # TODO: Sorting and Pagination
-        return result
-
-    def get_node_by_id(self, node_type, args):
-        """Return protobuf node object for given id."""
-        n_id = args.get('id')
-        o_name, w_name, remainder = n_id.split('/', 2)
-        w_id = f"{o_name}/{w_name}"
-        if node_type == 'proxy_nodes':
-            nodes = (
-                list(getattr(self.data[w_id], 'task_proxies', []))
-                + list(getattr(self.data[w_id], 'family_proxies', [])))
-        else:
-            nodes = getattr(self.data[w_id], node_type, [])
-        for node in nodes:
-            if node.id == n_id:
-                return node
-        return None
-
-    def get_edges_all(self, args):
-        """Return edges from all workflows, filter by args."""
-        w_args = {
-            'ids': args['workflows'],
-            'exids': args['exworkflows'],
-        }
-        result = []
-        for edge in [
-                n for k in self.get_workflow_msgs(w_args)
-                for n in getattr(k, 'edges')]:
-            result.append(edge)
-        # TODO: Sorting and Pagination
-        return result
-
-    def get_edges_by_id(self, args):
-        """Return protobuf edge objects for given id."""
-        nat_ids = args.get('native_ids', [])
-        w_ids = []
-        for nat_id in nat_ids:
-            oname, wname, eid = nat_id.split('/', 2)
-            w_ids.append(f"{oname}/{wname}")
-        w_ids = list(set(w_ids))
-        result = []
-        for edge in [
-                e for k in w_ids
-                for e in getattr(self.data[k], 'edges')]:
-            result.append(edge)
-        # TODO: Sorting and Pagination
-        return result
-
-    def _node_filter(self, node, args):
-        """Filter nodes based on attribute arguments"""
-        natts = self._collate_node_atts(node)
-        return (
-                (not args.get('states') or node.state in args['states']) and
-                not (args.get('exstates') and
-                     node.state in args['exstates']) and
-                (args.get('mindepth', -1) < 0 or
-                    node.depth >= args['mindepth']) and
-                (args.get('maxdepth', -1) < 0 or
-                    node.depth <= args['maxdepth']) and
-                (not args.get('ids') or
-                    self._node_atts_filter(natts, args['ids'])) and
-                not (args.get('exids') and
-                     self._node_atts_filter(natts, args['exids'])))
-
-    def _workflow_filter(self, flow, args):
-        """Filter workflows based on attribute arguments"""
-        natts = [flow.workflow.owner, flow.workflow.name, flow.workflow.status]
-        return (
-                (not args.get('ids') or
-                    self._workflow_atts_filter(natts, args['ids'])) and
-                not (args.get('exids') and
-                     self._workflow_atts_filter(natts, args['exids'])))
-
-    @staticmethod
-    def _workflow_atts_filter(natts, items):
-        """Match components of id argument with those of workflow id."""
-        for owner, name, status in set(items):
-            if ((not owner or fnmatchcase(natts[0], owner)) and
-                    (not name or fnmatchcase(natts[1], name)) and
-                    (not status or natts[2] == status)):
-                return True
-        return False
-
-    @staticmethod
-    def _node_atts_filter(natts, items):
-        """Match node id argument with node attributes."""
-        for owner, workflow, cycle, name, submit_num, state in items:
-            if ((not owner or fnmatchcase(natts[0], owner)) and
-                    (not workflow or fnmatchcase(natts[1], workflow)) and
-                    (not cycle or fnmatchcase(natts[2], cycle)) and
-                    any(fnmatchcase(nn, name) for nn in natts[3]) and
-                    (not submit_num or
-                        fnmatchcase(str(natts[4]), submit_num.lstrip('0'))) and
-                    (not state or natts[5] == state)):
-                return True
-        return False
-
-    @staticmethod
-    def _collate_node_atts(node):
-        """Collate node filter attributes."""
-        node_id = getattr(node, 'id')
-        slash_count = node_id.count('/')
-        n_id = node_id.split('/', slash_count)
-        if slash_count == 2:
-            n_cycle = None
-            n_name = n_id[2]
-        else:
-            n_cycle = n_id[2]
-            n_name = n_id[3]
-        return [
-            n_id[0],
-            n_id[1],
-            n_cycle,
-            getattr(node, 'namespace', [n_name]),
-            getattr(node, 'submit_num', None),
-            getattr(node, 'state', None),
-        ]
--- a/data_mgr.py
+++ b/data_mgr.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
 # -*- coding: utf-8 -*-
 # Copyright (C) 2019 NIWA & British Crown (Met Office) & Contributors.
 #
@@ -18,17 +16,18 @@
 
 """Create and update the data structure for all workflow services."""
 
+import asyncio
 from fnmatch import fnmatchcase
 
-from tornado import gen
 # from google.protobuf.json_format import MessageToDict
 
-# from cylc import flags
 from cylc.exceptions import ClientError, ClientTimeout
 from cylc.ws_messages_pb2 import PbEntireWorkflow
 from cylc.network.scan import MSG_TIMEOUT
 
 
+# TODO: define in and import this from cylc-flow
+# maps methods in cylc-flow server to the returned bytes message
 METHOD_MAP = {
     'pb_entire_workflow': PbEntireWorkflow
 }
@@ -54,9 +53,7 @@ class DataManager(object):
         self.ws_mgr = ws_mgr
         self.data = {}
 
-    # **
     # Data syncing
-    # **
     async def entire_workflow_update(self, ids=None):
         if ids is None:
             ids = []
@@ -64,9 +61,11 @@ class DataManager(object):
         ws_args = (
             (w_id, info['req_client'], 'pb_entire_workflow')
             for w_id, info in self.ws_mgr.workflows.items())
-        items = await gen.multi(
-            get_workflow_data(*args) for args in ws_args
-            if not ids or args[0] in ids)
+        gathers = ()
+        for args in ws_args:
+            if not ids or args[0] in ids:
+                gathers += (get_workflow_data(*args),)
+        items = await asyncio.gather(*gathers)
         for w_id, result in items:
             if result is not None and result != MSG_TIMEOUT:
                 new_data[w_id] = result
@@ -74,23 +73,20 @@ class DataManager(object):
             # atomic update
             self.data = new_data
         else:
-            # apparently faster than update(new_data)
-            for w_id, flow in new_data.items():
-                self.data[w_id] = flow
+            self.data.update(new_data)
 
-    # **
     # Data access filters
-    # **
     def get_workflow_msgs(self, args):
         """Return list of workflows."""
         result = []
         for flow_msg in self.data.values():
             if self._workflow_filter(flow_msg, args):
                 result.append(flow_msg)
+        # TODO: Sorting and Pagination
         return result
 
     def get_nodes_all(self, node_type, args):
-        """Return list of nodes (defns/proxies/jobs)."""
+        """Return nodes from all workflows, filter by args."""
         w_args = {
             'ids': args['workflows'],
             'exids': args['exworkflows'],
@@ -102,9 +98,10 @@ class DataManager(object):
         for node in nodes:
             if self._node_filter(node, args):
                 result.append(node)
+        # TODO: Sorting and Pagination
         return result
 
-    def get_nodes_id(self, node_type, args):
+    def get_nodes_by_id(self, node_type, args):
         """Return protobuf node objects for given id."""
         nat_ids = args.get('native_ids', [])
         w_ids = []
@@ -126,9 +123,10 @@ class DataManager(object):
             if ((not nat_ids or node.id in set(nat_ids)) and
                     self._node_filter(node, args)):
                 result.append(node)
+        # TODO: Sorting and Pagination
         return result
 
-    def get_node_id(self, node_type, args):
+    def get_node_by_id(self, node_type, args):
         """Return protobuf node object for given id."""
         n_id = args.get('id')
         o_name, w_name, remainder = n_id.split('/', 2)
@@ -145,7 +143,7 @@ class DataManager(object):
         return None
 
     def get_edges_all(self, args):
-        """Return all workflows protobuf edge objects for given id."""
+        """Return edges from all workflows, filter by args."""
         w_args = {
             'ids': args['workflows'],
             'exids': args['exworkflows'],
@@ -155,9 +153,10 @@ class DataManager(object):
                 n for k in self.get_workflow_msgs(w_args)
                 for n in getattr(k, 'edges')]:
             result.append(edge)
+        # TODO: Sorting and Pagination
         return result
 
-    def get_edges_id(self, args):
+    def get_edges_by_id(self, args):
         """Return protobuf edge objects for given id."""
         nat_ids = args.get('native_ids', [])
         w_ids = []
@@ -170,27 +169,12 @@ class DataManager(object):
                 e for k in w_ids
                 for e in getattr(self.data[k], 'edges')]:
             result.append(edge)
+        # TODO: Sorting and Pagination
         return result
 
     def _node_filter(self, node, args):
         """Filter nodes based on attribute arguments"""
-        node_id = getattr(node, 'id')
-        slash_count = node_id.count('/')
-        n_id = node_id.split('/', slash_count)
-        if slash_count == 2:
-            n_cycle = None
-            n_name = n_id[2]
-        else:
-            n_cycle = n_id[2]
-            n_name = n_id[3]
-        natts = [
-            n_id[0],
-            n_id[1],
-            n_cycle,
-            getattr(node, 'namespace', [n_name]),
-            getattr(node, 'submit_num', None),
-            getattr(node, 'state', None),
-        ]
+        natts = self._collate_node_atts(node)
         return (
                 (not args.get('states') or node.state in args['states']) and
                 not (args.get('exstates') and
@@ -236,3 +220,24 @@ class DataManager(object):
                     (not state or natts[5] == state)):
                 return True
         return False
+
+    @staticmethod
+    def _collate_node_atts(node):
+        """Collate node filter attributes."""
+        node_id = getattr(node, 'id')
+        slash_count = node_id.count('/')
+        n_id = node_id.split('/', slash_count)
+        if slash_count == 2:
+            n_cycle = None
+            n_name = n_id[2]
+        else:
+            n_cycle = n_id[2]
+            n_name = n_id[3]
+        return [
+            n_id[0],
+            n_id[1],
+            n_cycle,
+            getattr(node, 'namespace', [n_name]),
+            getattr(node, 'submit_num', None),
+            getattr(node, 'state', None),
+        ]
--- a/data_mgr.py
+++ b/data_mgr.py
@@ -1,0 +1,238 @@
+#!/usr/bin/env python3
+
+# -*- coding: utf-8 -*-
+# Copyright (C) 2019 NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Create and update the data structure for all workflow services."""
+
+from fnmatch import fnmatchcase
+
+from tornado import gen
+# from google.protobuf.json_format import MessageToDict
+
+# from cylc import flags
+from cylc.exceptions import ClientError, ClientTimeout
+from cylc.ws_messages_pb2 import PbEntireWorkflow
+from cylc.network.scan import MSG_TIMEOUT
+
+
+METHOD_MAP = {
+    'pb_entire_workflow': PbEntireWorkflow
+}
+
+
+async def get_workflow_data(w_id, client, method):
+    # Use already established client
+    try:
+        pb_msg = await client.async_request(method)
+    except ClientTimeout as exc:
+        return (w_id, MSG_TIMEOUT)
+    except ClientError as exc:
+        return (w_id, None)
+    else:
+        ws_data = METHOD_MAP[method]()
+        ws_data.ParseFromString(pb_msg)
+    return (w_id, ws_data)
+
+
+class DataManager(object):
+
+    def __init__(self, ws_mgr):
+        self.ws_mgr = ws_mgr
+        self.data = {}
+
+    # **
+    # Data syncing
+    # **
+    async def entire_workflow_update(self, ids=None):
+        if ids is None:
+            ids = []
+        new_data = {}
+        ws_args = (
+            (w_id, info['req_client'], 'pb_entire_workflow')
+            for w_id, info in self.ws_mgr.workflows.items())
+        items = await gen.multi(
+            get_workflow_data(*args) for args in ws_args
+            if not ids or args[0] in ids)
+        for w_id, result in items:
+            if result is not None and result != MSG_TIMEOUT:
+                new_data[w_id] = result
+        if not ids:
+            # atomic update
+            self.data = new_data
+        else:
+            # apparently faster than update(new_data)
+            for w_id, flow in new_data.items():
+                self.data[w_id] = flow
+
+    # **
+    # Data access filters
+    # **
+    def get_workflow_msgs(self, args):
+        """Return list of workflows."""
+        result = []
+        for flow_msg in self.data.values():
+            if self._workflow_filter(flow_msg, args):
+                result.append(flow_msg)
+        return result
+
+    def get_nodes_all(self, node_type, args):
+        """Return list of nodes (defns/proxies/jobs)."""
+        w_args = {
+            'ids': args['workflows'],
+            'exids': args['exworkflows'],
+        }
+        nodes = [
+            n for k in self.get_workflow_msgs(w_args)
+            for n in getattr(k, node_type)]
+        result = []
+        for node in nodes:
+            if self._node_filter(node, args):
+                result.append(node)
+        return result
+
+    def get_nodes_id(self, node_type, args):
+        """Return protobuf node objects for given id."""
+        nat_ids = args.get('native_ids', [])
+        w_ids = []
+        for nat_id in nat_ids:
+            o_name, w_name, nid = nat_id.split('/', 2)
+            w_ids.append(f"{o_name}/{w_name}")
+        if node_type == 'proxy_nodes':
+            nodes = (
+                n for w_id in set(w_ids)
+                for n in (
+                    list(getattr(self.data[w_id], 'task_proxies', []))
+                    + list(getattr(self.data[w_id], 'family_proxies', []))))
+        else:
+            nodes = (
+                n for w_id in set(w_ids)
+                for n in getattr(self.data[w_id], node_type, []))
+        result = []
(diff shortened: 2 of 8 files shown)


dwsutherland:
Travis will likely fail, as cylc/cylc is not a requirement yet.

kinow:
>Travis will likely fail, as cylc/cylc-flow is not a requirement yet.

Indeed it failed with: `ModuleNotFoundError: No module named 'cylc'`

In this case I think once everybody is back from holidays, perhaps we could prioritize https://github.com/cylc/cylc/pull/2990. That one has become a blocker for a few issues now.

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+class DataManager(object):
+
+    def __init__(self, ws_mgr):
+        self.ws_mgr = ws_mgr
+        self.data = {}
+
+    # **
+    # Data syncing
+    # **
+    async def entire_workflow_update(self, ids=[]):
```
kinow: Mutable argument `ids`. Instead it should be `None` and then check its value... see [this](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) one for more.

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+        if ids == []:
+            self.data = data
+        else:
+            for id, flow_msg in data.items():
+                self.data[id] = flow_msg
+
+    # **
+    # Data access filters
+    # **
+    def get_workflow_msgs(self, args={}):
```
kinow: Mutable default argument.

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+        """Return list of workflows."""
+        ids = [self.parse_workflow_id(id) for id in args['ids']]
+        exids = [self.parse_workflow_id(id) for id in args['exids']]
+        result = []
+        for flow_msg in self.data.values():
+            if self._workflow_filter(flow_msg, ids, exids, args):
+                result.append(flow_msg)
+        return result
+
+    def get_workflows(self, args={}):
```
kinow: Mutable default argument.

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+        """Return list of workflows."""
+        ids = [self.parse_workflow_id(id) for id in args['ids']]
+        exids = [self.parse_workflow_id(id) for id in args['exids']]
+        result = []
+        for flow in self.data.values():
+            if self._workflow_filter(flow, ids, exids, args):
+                result.append(flow.workflow)
+        return result
+
+    def get_nodes_all(self, node_type, args={}):
```
kinow: Mutable default argument.

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+            for n in getattr(k, node_type)]
+        ids = [self.parse_node_id(id, node_type) for id in args['ids']]
+        exids = [self.parse_node_id(id, node_type) for id in args['exids']]
+        result = []
+        for node in nodes:
+            if self._node_filter(node, ids, exids, args):
+                result.append(node)
+        return result
+
+    def get_nodes_id(self, node_type, args={}):
```
kinow: Mutable default argument.

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+                n for w_id in set(w_ids)
+                for n in getattr(self.data[w_id], node_type, []))
+        result = []
+        for node in nodes:
+            if ((not nat_ids or node.id in set(nat_ids)) and
+                    self._node_filter(node, ids, exids, args)):
+                result.append(node)
+        return result
+
+    def get_node_id(self, node_type, args={}):
```
kinow: Mutable default argument.

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+                list(getattr(self.data[w_id], 'task_proxies', []))
+                + list(getattr(self.data[w_id], 'family_proxies', [])))
+        else:
+            nodes = getattr(self.data[w_id], node_type, [])
+        for node in nodes:
+            if node.id == nat_id:
+                return node
+        return None
+
+    def get_edges_all(self, args={}):
```
kinow: Mutable default argument.

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+            'exids': args['exworkflows'],
+        }
+        result = []
+        for edge in [
+                n for k in self.get_workflow_msgs(w_args)
+                for n in getattr(k, 'edges')]:
+            result.append(edge)
+        return result
+
+    def get_edges_id(self, args={}):
```
kinow: Mutable default argument.

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+            w_ids.append(f"{oname}/{wname}")
+        w_ids = list(set(w_ids))
+        result = []
+        for edge in [
+                e for k in w_ids
+                for e in getattr(self.data[k], 'edges')]:
+            result.append(edge)
+        return result
+
+    def _node_filter(self, node, ids=[], exids=[], args={}):
```
kinow: Mutable default arguments.

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+    def _node_atts_filter(natts, items):
+        for point, name, submit_num, state in items:
+            if ((not point or fnmatchcase(natts[0], point)) and
+                    any(fnmatchcase(nn, name) for nn in natts[1]) and
+                    (not submit_num or fnmatchcase(natts[2], submit_num)) and
+                    (not state or natts[3] == state)):
+                return True
+        return False
+
+    def _workflow_filter(self, flow, ids=[], exids=[], args={}):
```
kinow: Mutable default arguments. And `args` is never used?

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+        ids = [self.parse_node_id(id, node_type) for id in args['ids']]
+        exids = [self.parse_node_id(id, node_type) for id in args['exids']]
+        result = []
+        for node in nodes:
+            if self._node_filter(node, ids, exids, args):
+                result.append(node)
+        return result
+
+    def get_nodes_id(self, node_type, args={}):
+        """Return GrapphQL task object for given id."""
```
kinow: s/GrapphQL/GraphQL/

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+            owner, workflow = head.split('/', 1)
+        else:
+            # more common to filter on workflow (with owner constant)
+            workflow = head
+        return (owner, workflow, status)
+
+    @staticmethod
+    def parse_node_id(item, node_type=None):
+        """Parse node id argument.
+        Args:
```
kinow: The docstrings here are missing the arguments. I think the Google style docstrings do not support optional or tuples. So the `Items` entry below could be part of the documentation for `item` I think... or with PEP-484 tuple types can be specified too)

data_mgr.py
```diff
@@ -0,0 +1,296 @@
...
+            for id, info in self.ws_mgr.workflows.items())
+        # print(self.ws_mgr.workflows)
+        items = await gen.multi(
+            get_workflow_data(*args) for args in ws_args
+            if ids == [] or args[0] in ids)
+        for id, result in items:
+            if result is not None and result != MSG_TIMEOUT:
+                data[id] = result
+        # atomic update
+        if ids == []:
```
kinow: `if not ids:`

data_mgr.py
dwsutherland: Surprising behaviour.. Thanks!

data_mgr.py
dwsutherland: It's not used yet, however as soon as I add filters for filtering workflows by status or something, then it will be.. (will remove for now)

data_mgr.py
dwsutherland: I was just roughly following what Oliver did in the ZeroMQ (not that he did this), I need to read up on Google style docstrings...

data_mgr.py
dwsutherland: Yes in light of changing ids input arg to None, this makes sense.

kinow:
@dwsutherland see my last commit with the temporary fix for this build. Now Travis is [happy again](https://travis-ci.org/cylc/cylc-uiserver/builds/524791895).

cylc_singleuser.py
```diff
@@ -125,6 +153,10 @@ def start(self):
         signal.signal(signal.SIGINT, app.signal_handler)
         app.listen(self._port)
         ioloop.PeriodicCallback(app.try_exit, 100).start()
+        ioloop.IOLoop.current().add_callback(self.ws_mgr.gather_workflows)
+        ioloop.PeriodicCallback(self.ws_mgr.gather_workflows, 20000).start()
+        ioloop.PeriodicCallback(
+            self.data_mgr.entire_workflow_update, 5000).start()
```
matthewrmshin: May be worth commenting on the choice of these magic numbers?

data_mgr.py
```diff
@@ -0,0 +1,299 @@
...
+    # Data syncing
+    # **
+    async def entire_workflow_update(self, ids=None):
+        if ids is None:
+            ids = []
+        data = {}
+        method = 'pb_entire_workflow'
+        ws_args = (
+            (id, info['req_client'], method)
+            for id, info in self.ws_mgr.workflows.items())
```
matthewrmshin: Should use `id_` (or something else) as `id` is a Python built-in function.

data_mgr.py
```diff
@@ -0,0 +1,299 @@
...
+        w_ids = []
+        for nat_id in nat_ids:
+            oname, wname, nid = nat_id.split('/', 2)
+            w_ids.append(f"{oname}/{wname}")
+        if node_type == 'proxy_nodes':
+            nodes = (
+                n for w_id in set(w_ids)
+                for n in (
+                    list(getattr(self.data[w_id], 'task_proxies', []))
+                    + list(getattr(self.data[w_id], 'family_proxies', []))))
```
matthewrmshin: There is a bit of a logical pattern `list(getattr(self.data[w_id], node_type, []))` in here and below. Perhaps you can introduce a function or method to improve readability?

data_mgr.py
```diff
@@ -0,0 +1,299 @@
...
+                    args.get('exstates') and
+                    node.state in args['exstates']) and
+                (args.get('mindepth', -1) < 0 or
+                    node.depth >= args['mindepth']) and
+                (args.get('maxdepth', -1) < 0 or
+                    node.depth <= args['maxdepth']) and
+                (not ids or self._node_atts_filter(natts, ids)) and
+                not (exids and self._node_atts_filter(natts, exids))):
+            return True
+        return False
```
matthewrmshin: More comment needed for this!

You can just do:

```
return ...
```

Instead of:

```
if ...:
    return True
return False
```

dwsutherland: (>.<)!

data_mgr.py
dwsutherland: Oh wow never knew >.>

cylc_singleuser.py
dwsutherland: Yeah sure... No particular strategy here, chose 20sec and 5sec respectively, just for testing.. The will change or be removed once PUB/SUB goes in..

data_mgr.py
dwsutherland: No. Protobuf message lists don't support `+` ... Happy to implement something else if you have any suggestions..

data_mgr.py
```diff
@@ -0,0 +1,290 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Create and update the data structure for all workflow services."""
+
+from fnmatch import fnmatchcase
+
+from tornado import gen
+# from google.protobuf.json_format import MessageToDict
+
+# from cylc import flags
```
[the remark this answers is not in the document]
dwsutherland: Yeah, I use `MessageToDict` now and then for testing message content... And I haven't implemented any flags yet, so on the fence about this one.

workflow_services_mgr.py
```diff
@@ -0,0 +1,112 @@
...
+        self.workflows = {}
+
+    def spawn_workflow(self):
+        # TODO - Spawn workflows
+        pass
+
+    async def gather_workflows(self):
+        workflows = {}
+        cre_owner, cre_name = re_compile_filters(None, ['.*'])
+        timeout = CLIENT_TIMEOUT
```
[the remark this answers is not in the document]
dwsutherland: Sorry, oops! I've changed most of the methods/functions half a dozen times, and there's bound to be some dags hanging about..

workflow_services_mgr.py
```diff
@@ -0,0 +1,113 @@
...
+
+from cylc.hostuserutil import is_remote_host, get_host_ip_by_name
+from cylc.network.client import SuiteRuntimeClient
+from cylc.network.scan import (
+    get_scan_items_from_fs, re_compile_filters, MSG_TIMEOUT)
+
+INACTIVITY_TIMEOUT = 10.0
+MSG_QUIT = "QUIT"
+MSG_TIMEOUT = "TIMEOUT"
+SLEEP_INTERVAL = 0.01
```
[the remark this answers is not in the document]
dwsutherland: Agreed.

schema.py
```diff
@@ -0,0 +1,810 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""GraphQL API schema via Graphene implementation."""
+
+import graphene
```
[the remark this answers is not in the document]
dwsutherland: Yeah I've been tossing this one up.. All the docs have examples with `graphene` imported, so I thought I'd play it safe. However, given that I've isolated the schema from the server somewhat, it might make sense (definitely would de-clutter it)..

cylc_singleuser.py
```diff
@@ -125,6 +153,14 @@ def start(self):
...
         app.listen(self._port)
         ioloop.PeriodicCallback(app.try_exit, 100).start()
+        # Discover workflows on intiial start up.
+        ioloop.IOLoop.current().add_callback(self.ws_mgr.gather_workflows)
+        # Arbitrary intervals (msec) chosen for the two periodic callbacks
+        # below. These will likely change or be removed with PUB/SUB updates.
+        # Check for new workflows every 20sec.
+        ioloop.PeriodicCallback(self.ws_mgr.gather_workflows, 20000).start()
+        ioloop.PeriodicCallback(
+            self.data_mgr.entire_workflow_update, 5000).start()
```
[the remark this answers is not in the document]
dwsutherland: Yeah I could stick to prime numbers (i.e 3 & 19)... If the callbacks take longer than the specified time Tornado won't run the next one (BTW). However (might be a way!), as it is, we can't always guarantee that they won't clash because we don't know how long they both take to complete.. So, bar running them synchronously, we might just have to catch the error in the data manager if/when it ever occurs.

workflow_services_mgr.py
```diff
@@ -0,0 +1,113 @@
...
+    if is_remote_host(host):
+        try:
+            host = get_host_ip_by_name(host)  # IP reduces DNS traffic
+        except socket.error as exc:
+            if cylc.flags.debug:
+                raise
+            sys.stderr.write("ERROR: %s: %s\n" % (exc, host))
+            return (reg, host, port, None)
+
+    # NOTE: Connect to the suite by host:port, this was the
```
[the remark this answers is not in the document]
dwsutherland: I used @oliver-sanders' `scan_one` function `cylc.network.scan` and modified it to also return the client for use in the data manager.. I believe it should be `, this way the`..

data_mgr.py
```diff
@@ -0,0 +1,290 @@
...
+                    args.get('exstates') and
+                    node.state in args['exstates']) and
+                (args.get('mindepth', -1) < 0 or
+                    node.depth >= args['mindepth']) and
+                (args.get('maxdepth', -1) < 0 or
+                    node.depth <= args['maxdepth']) and
+                (not ids or self._node_atts_filter(natts, ids)) and
+                not (exids and self._node_atts_filter(natts, exids)))
+
+    @staticmethod
```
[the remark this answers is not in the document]
dwsutherland: I guess that would make sense if they weren't only used/had-utility as helper functions to other methods in the class and were not related to the theme of the class. If any of them qualify it would be `parse_workflow_id` and `parse_node_id`..

data_mgr.py
```diff
@@ -0,0 +1,290 @@
...
+
+                name
+                point/name:state
+                name.point:state
+                nam*.poin*
+                point/name/submit_num:state
+                name.point.submit_num:state
+
+        Returns:
+            A tuple of id components in repective order. For example:
```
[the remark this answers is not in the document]
dwsutherland: Done.

schema.py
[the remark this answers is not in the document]
dwsutherland: Done.

data_mgr.py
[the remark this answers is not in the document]
dwsutherland: Only the two mentioned.

data_mgr.py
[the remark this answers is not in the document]
dwsutherland: Leaving in for now.

schema.py
```diff
@@ -0,0 +1,813 @@
...
+    req_state = String()
+    satisfied = Boolean()
+    message = String()
+
+
+class Prerequisite(ObjectType):
+    """Task prerequisite."""
+    expression = String()
+    conditions = List(
+        Condition,
```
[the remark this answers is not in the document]
dwsutherland: The `lambda` are for fields of of an object type to lazy reference themselves or object types defined after it.
Document example: https://docs.graphene-python.org/en/latest/types/interfaces/#interfaces

workflow_services_mgr.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # NOTE: Connect to the suite by host:port. This way the
+    #       SuiteRuntimeClient will not attempt to check the contact file
+    #       which would be unnecessary as we have already done so.
+    # NOTE: This part of the scan *is* IO blocking.
+    client = SuiteRuntimeClient(reg, host=host, port=port, timeout=timeout)
+
+    result = {}
+    result['client'] = client
+    for method in methods:
```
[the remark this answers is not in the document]
dwsutherland: Yes true

dwsutherland:
@oliver-sanders 
> I'd kinda envisaged the GraphQL layer being in `cylc-flow` itself rather than in the UI-Server, 
Well we could have both, the data structure is present at both. They would be different though, because the `cylc-flow` one wouldn't be aware of other workflows. If we only had one GraphQL layer at the `cylc-flow` end then:
- There would be no point in maintaining a data structure at the UI-Server (which we kinda agreed on, and your ZeroMQ prototype addressed the update of this).
- We would lose the resolver ability to filter over data elements (flows, tasks, families, jobs, edges ...etc) from multiple workflows (without a huge mass of ZMQ requests routed according to the query args).
- Apart from the type checking on assignment, protobuf would be pointless.
- The demand of the web-client will directly effect the workflows.
- GraphQL Subscription haven't been developed for ZeroMQ PUB/SUB, not sure how much work this would be.
- ?

>this way could work fine too but we will need to have a quick think about:
> 
> * How the UI-Server will work when connecting to workflows of different versions.

Potential conflicts when:
- Protobuf message definition changes. Although the fields are optional and additions etc are fine. If other `cylc-flow` code changes, then protobuf field assignment and/or type checking will raise an exception.
- The `cylc-flow` API changes in some way between versions.. 
> I guess here we have a GraphQL-ZMQ adapter layer so it's about ensuring that the adapter isn't brittle to changes.

Exactly, protobuf helps alot (it was designed with these issues in mind!). But I can imagine similar issues even with only a `cylc-flow` GraphQL, some of which would be passed on to the web-client instead.

> * How to make this work with (API-On-The-Fly)[https://github.com/[cylc/cylc-flow/issues/2123](https://github.com/cylc/cylc-flow/issues/2123)].
> 

Will need to read up again, but GraphQL has introspection which can return all the information about the schema (including descriptions, which I need to do more of!), as shown in the `graphiql` interface.

schema.py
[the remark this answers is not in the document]
oliver-sanders: OK, so it's done for the benefit of delayed execution.

workflow_services_mgr.py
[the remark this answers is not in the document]
dwsutherland: Fixed (as in fixed method too)..

oliver-sanders:
I'm struggling quite a bit with this one trying to see how it can fit into my past visions of decoupling Cylc components. I've tried and kinda failed to outline some of it here, I might make more sense of things later...

**The issue I'm trying to get my head around is this:**

* Cylc-UI is tightly coupled to Cylc-UIServer by the GraphQL interface
* Cylc-UIServer is tightly coupled to Cylc-Flow by the ZMQ interface
* Therefore the Cylc-UI is tightly coupled to Cylc-Flow.

I don't know what the best way to decouple things is, it was all very clear to me once but things have evolved so far since then I can't see it any more. Back when things were simpler and we were talking about one workflow talking to one GUI (in Electron) it was easy to see how decoupling the Cylc Flow and Cylc UI versions might work. The GUI can just request the API from the workflow and present it in a nice form for the user (e.g. the stop suite panel) and GraphQL can handle most upgrade requirements simply enough.

Now we have one UIServer talking to multiple workflows things are more difficult as by defining the GraphQL schema at the UIServer we are defining the interface for *all* suites irrespective of their version. If we make changes to a Cylc suite (which we have form of) we will have to put a lot of adapter logic into the UIServer. GraphQL (and presumably ProtoBuff) can protect us from the most basic examples (fields added / removed) but the rest...

I'm struggling to see a way forward at the moment, @dwsutherland could you help me out by outlining how you envisage us handling some example API changes e.g. (adding fields, changing arguments in ZMQ endpoints, etc)?

**Further thoughts on upgrading Cylc components:** (a quick observation with no point to make really)

Cylc-UI is instantaneously upgradeable, that is to say you can change versions at the press of a button or just by refreshing the page. Handling change here is easy.

The Cylc-UIServer is much harder to upgrade, we have to wait for users to close their browser windows for the UIServer to become inactive for it to auto-shutdown (or have some sort of force upgrade system).

So we probably want the dynamic components isolated in Cylc-UI where they are easy to change?

matthewrmshin:
Is it possible to define multiple workflow API data models under a single GraphQL schema? (With luck, we'll only have to back support a small number of workflow server APIs for each UI/UI-Server version. So even if we have to spell out all the data models explicitly for the supported workflow API data models, it should not be a big deal. But the top level data model must be able to handle this.)

For upgrading a ui-server... I suppose it should be possible for use to configure the hub to tell all ui-servers running outdated software to shutdown and restart with the current software version. Browsers sessions can reconnect automatically. This does not sound any worse than what Outlook (web) does to my sessions everyday.

dwsutherland:
@oliver-sanders: Thanks for bring this up, yes there will be ways to break this PR data model..
I was a little bit taken back, not that it matters, as we had discussed this design before I implemented it (obviously not in enough detail!).

I guess before we posit any solutions, we need to describe problems are unique to the current _proposed_ data model...
 
> **The issue I'm trying to get my head around is this:**
> 
> * Cylc-UI is tightly coupled to Cylc-UIServer by the GraphQL interface
> * Cylc-UIServer is tightly coupled to Cylc-Flow by the ZMQ interface
> * Therefore the Cylc-UI is tightly coupled to Cylc-Flow.
> 

Firstly I'm not exactly sure what you mean by _tightly_, because I don't really see how 
> Back when things were simpler and we were talking about one workflow talking to one GUI (in Electron) it was easy to see how decoupling the Cylc Flow and Cylc UI versions might work. The GUI can just request the API from the workflow and present it in a nice form for the user (e.g. the stop suite panel) and GraphQL can handle most upgrade requirements simply enough.
>

doesn't apply to `UIServer => GUI`? ... To the GUI it's just a GraphQL endpoint in both designs, and I would have thought that in both cases the GUI would potentially be different repositories/versions than the API they connect to.

> Cylc-UI is instantaneously upgradeable, that is to say you can change versions at the press of a button or just by refreshing the page. Handling change here is easy.
>

The UIServer, of course, can ensure the version of GUI/`cylc-ui` is compatible on start up
as it has to serve the HTML via Tornado, so those versions are always in sync and like you said users can just refresh the page...

> The Cylc-UIServer is much harder to upgrade, we have to wait for users to close their browser windows for the UIServer to become inactive for it to auto-shutdown (or have some sort of force upgrade system).
> 

Maybe it's just Dev/Debug mode.. But during development you can have the UIServer running and change the source code, Tornado must reload or something because I never have to shut my browser down.
Even if we did have to shut the UIServer and/or JupyterHub, refreshing the browser usually just restarts the UIServer and/or reloads the page..


> Now we have one UIServer talking to multiple workflows things are more difficult as by defining the GraphQL schema at the UIServer we are defining the interface for _all_ suites irrespective of their version. 
> If we make changes to a Cylc suite (which we have form of) we will have to put a lot of adapter logic into the UIServer.
>

This setup is arguably how it should be! Why should the frontend have to deal with Workflow-GraphQL version inconsistencies in the backend?
Let me explain...

The UIServer will always be the one talking to all the user's workflows, because (and correct me if I'm wrong):
- [JupyterHub](https://zero-to-jupyterhub.readthedocs.io/en/stable/architecture.html) always redirects the signed in user to the UIServer
- The UIServer will monitor and have the ability to spawn the suites.
- ZeroMQ doesn't serve HTML and setup URLs, so there needs to be some translation between browser-proxy-jupyterhub and the workflow

We could still have graphql at the workflow, however:
- Even though queries/mutations could be sent as strings through ZeroMQ, there is no support/tools for it (i.e. no GraphiQL interface, we would have to cannibalize the graphene-tornado code for the interface)
- How do GraphQL Subscription Types work in ZeroMQ? We would have to write our own tool for that (not the same as ZeroMQ PUB/SUB even if it rides it).
- I think the ApolloGraphQL client expects to hook straight into a GraphQL endpoint for introspection (and possibly more...)
- How do you traverse across workflows in a single query? You couldn't really do it without routing your queries to every workflow
- We still have the issue with running multiple workflow versions of different graphql APIs (perhaps the queries differ, or input args or w/e), and all this does is pass a backend problem to the frontend developer...

BTW - GraphQL is designed/often-used to take multiple input sources of different types (SQLDB, ProtoBuf, REST, MongoDB ...etc) and provide a coherent data access point.
(also; I think protobuf is arguably a better data format then JSON, for transport over ZeroMQ)


So, I think we can narrow down the question to:  **How do we envisage `Cylc-Flow <=> Cylc-UIServer` handling some example API changes e.g. (adding fields, changing arguments in ZMQ endpoints, etc)?**

> GraphQL (and presumably ProtoBuff) can protect us from the most basic examples (fields added / removed) but the rest...
> 

Yes, ProtoBuf can easily handle adding more and more fields and I've made all the fields in the `.proto` definition optional (which is default and recommended) so you can remove them (within reason) as you please..
In fact the fields are only sent if they are populated, which means if we wanted then with incremental updates we could just populate the same protobuf objects with only the updated fields (i.e taskproxy status ...etc)...
This also means I don't do any filtering at the workflow, only the full protobuf objects are requested. This means we can design ZeroMQ endpoints with very few arguments, and those with them should have defaults (i.e. an argument might be a list of TaskProxy IDs found missing on checksum comparison, with the default sending all TaskProxy protobuf messages).

Leaving you with the question: How the hell will the UIServer handle this? 
I set defaults in the fields of the GraphQL schema, so if they are missing in the input data it returns that.. And in fact; if no default is set and required is not set, then graphql will return `null`.

If there are other breaking changes, then either the workflows or uiserver will have to contain back compatibility
As @matthewrmshin suggested; we could have the UIServer restart the workflow (by default or, with option in `.cylc/global.rc`, ignore suites with incompatible API version), and require the `identity` endpoint to always be there and always return the API version..


Just as a side note, we had the same issue with the REST Cherrypy API, we're we had to duplicate the URLs with `/id` (or something) as a prefix for back compatibility... So there's always going to be something


Please call me out if I'm missing something or plain wrong!

oliver-sanders:
OK, it looks like I'm the only one who's concerned so should probably push ahead whatever. Perhaps discuss at the next video meet.

Something else to think about which we came up with at a UK team meet was that we want out CLI client to have the ability to route via the UI server. I guess the UI server could just act as a TCP-Websocket buffer?

dwsutherland:
> 
> Something else to think about which we came up with at a UK team meet was that we want out CLI client to have the ability to route via the UI server. I guess the UI server could just act as a TCP-Websocket buffer?

Hmm... You need both, so how would the CLI discover a running UIServer?

At the moment, I import the workflow runtime client (from `network/client.py` ) into the UIServer and instantiate for every discovered workflow (on startup and refresh every 19sec (arbitrary)).. 

Perhaps for uniformity we could put a graphql schema the workflow, and import some of this to the graphql schema from there into the UIServer... But it would still have to be designed to handle multiple workflows (defaulting to itself when running in a single workflow)...

The UIServer is running with Tornado (which will be TCP-Websocket at some point, but for browser interaction), perharps we could start up a ZeroMQ stub in addition....
(I did see the use of Tornado in the pyzmq example folder)

dwsutherland:
> Hey @dwsutherland , any chance of including unit tests for the non-web related code? i.e. no need to bother with the Tornado handlers being tested.

@kinow - I'm a little reluctant to add tests while the code is in flux and under review... Perhaps it should be done last before merging. Thoughts?

kinow:
>@kinow - I'm a little reluctant to add tests while the code is in flux and under review... Perhaps it should be done last before merging. Thoughts?

+1, good idea.

data_mgr.py
```diff
@@ -0,0 +1,238 @@
...
+
+from fnmatch import fnmatchcase
+
+from tornado import gen
+# from google.protobuf.json_format import MessageToDict
+
+# from cylc import flags
+from cylc.exceptions import ClientError, ClientTimeout
+from cylc.ws_messages_pb2 import PbEntireWorkflow
+from cylc.network.scan import MSG_TIMEOUT
```
kinow: We will have to update to `cylc.flow` later.

cylc_singleuser.py
```diff
@@ -119,6 +160,14 @@ def start(self):
         signal.signal(signal.SIGINT, app.signal_handler)
         app.listen(self._port)
         ioloop.PeriodicCallback(app.try_exit, 100).start()
+        # Discover workflows on intiial start up.
```
kinow: s/intiial/initial

data_mgr.py
```diff
@@ -0,0 +1,238 @@
+#!/usr/bin/env python3
```
kinow: We do not need a shebang in this file, as it won't be executed as a script.

schema.py
```diff
@@ -0,0 +1,934 @@
+#!/usr/bin/env python3
```
kinow: We don't need a shebang here.

workflow_services_mgr.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
```
kinow: No need for shebang.

cylc_singleuser.py
```diff
@@ -119,6 +160,14 @@ def start(self):
...
+        ioloop.IOLoop.current().add_callback(self.ws_mgr.gather_workflows)
+        # Arbitrary intervals (msec) chosen for the two periodic callbacks
+        # below. These will likely change or be removed with PUB/SUB updates.
+        # Check for new workflows every 19sec.
+        ioloop.PeriodicCallback(self.ws_mgr.gather_workflows, 19000).start()
+        ioloop.PeriodicCallback(
+            self.data_mgr.entire_workflow_update, 5000).start()
```
kinow: These two `PeriodiCallback`'s are appearing with a warning on my IDE. Checking the signature of `PeriodicCallback` in Tornado source code it specifies the type as `Callable[[], None]`.

From [the typing docs about Callable](https://docs.python.org/3/library/typing.html#callable) 

>Frameworks expecting callback functions of specific signatures might be type hinted using Callable[[Arg1Type, Arg2Type], ReturnType].

So the Tornado function signature says `Callable[[], None]` which is equivalent to a `Callable` with no parameters, that does not return anything.

`async` methods are marked as [`Awaitable`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Awaitable)s.

Still trying to confirm that these are being executed correctly every 19 and 5 secs.

data_mgr.py
```diff
@@ -0,0 +1,238 @@
...
+    # Data syncing
+    # **
+    async def entire_workflow_update(self, ids=None):
+        if ids is None:
+            ids = []
+        new_data = {}
+        ws_args = (
+            (w_id, info['req_client'], 'pb_entire_workflow')
+            for w_id, info in self.ws_mgr.workflows.items())
+        items = await gen.multi(
```
kinow: I think we don't need to import `gen` and `multi` from tornado. Instead `asyncio.gather` should do it.

Tornado documentation suggests they have only a few coroutines that are still useful post Python 3, unless you want backward compatibility :+1: 

image


I think `gather` is simpler too, as you can just pass multiple arguments, instead of having to create a list.

data_mgr.py
```diff
@@ -0,0 +1,238 @@
...
+        for w_id, result in items:
+            if result is not None and result != MSG_TIMEOUT:
+                new_data[w_id] = result
+        if not ids:
+            # atomic update
+            self.data = new_data
+        else:
+            # apparently faster than update(new_data)
+            for w_id, flow in new_data.items():
+                self.data[w_id] = flow
```
kinow: Got interested in this one, as I would probably just use `.update(new_data)`. Testing with little data I think `dict.update` was faster?

image

Then with a dictionary with some [states and capitals](https://stackoverflow.com/questions/49193164/python-dictionary-states-and-capital-game) the difference for the time running `dict.update` was even bigger.

 
image

Even if I initialize the dictionaries with 15 items each, so that the dictionaries are not empty, `dict.update` still appears to be faster.

image

Do you have a link to some page explaining why iterating and accessing each item would be faster?

data_mgr.py
```diff
@@ -0,0 +1,238 @@
...
+    # **
+    # Data access filters
+    # **
+    def get_workflow_msgs(self, args):
+        """Return list of workflows."""
+        result = []
+        for flow_msg in self.data.values():
+            if self._workflow_filter(flow_msg, args):
+                result.append(flow_msg)
+        return result
```
kinow: Equivalent to `return [msg for msg in self.data.values() if self._workflow_filter(msg, args)]` ? Easier to read if we can go up to 86 characters here... otherwise breaking it into another line might still be easier to read?

data_mgr.py
```diff
@@ -0,0 +1,238 @@
...
+                e for k in w_ids
+                for e in getattr(self.data[k], 'edges')]:
+            result.append(edge)
+        return result
+
+    def _node_filter(self, node, args):
+        """Filter nodes based on attribute arguments"""
+        node_id = getattr(node, 'id')
+        slash_count = node_id.count('/')
+        n_id = node_id.split('/', slash_count)
```
kinow: I think it would be easier to test this part, and also reduce the complexity in `_node_filter` if we moved it to a `_get_node_id` function.

data_mgr.py
```diff
@@ -0,0 +1,238 @@
...
+                for n in getattr(self.data[w_id], node_type, []))
+        result = []
+        for node in nodes:
+            if ((not nat_ids or node.id in set(nat_ids)) and
+                    self._node_filter(node, args)):
+                result.append(node)
+        return result
+
+    def get_node_id(self, node_type, args):
+        """Return protobuf node object for given id."""
```
kinow: I thought this function would return the `node_id` of a node. But instead it appears to return a node for a given node id. So I think we could have a better name here - though I am horrible with names myself. Maybe `get_node_by_id`?

data_mgr.py
```diff
@@ -0,0 +1,238 @@
...
+                list(getattr(self.data[w_id], 'task_proxies', []))
+                + list(getattr(self.data[w_id], 'family_proxies', [])))
+        else:
+            nodes = getattr(self.data[w_id], node_type, [])
+        for node in nodes:
+            if node.id == n_id:
+                return node
+        return None
+
+    def get_edges_all(self, args):
```
kinow: Reading the documentation, it looks like it does not get all edges. But instead, all edges for a given ID (in `args` I assume). So maybe `get_edges_by_id`? Or just `get_edges`? Though I see `get_edges_id` too... not sure what's the difference?

kinow:
Enough of my comments. Feel free to mark them as resolved if you believe they are not pertinent @dwsutherland. I am now testing the GraphQL endpoints in the Vue.js app :+1:

setup.py
```diff
@@ -23,10 +23,13 @@
...
     long_description = f.read()
 
+# TODO: update cylc/cylc-flow dependency after cylc-flow PR has been merged
 install_requires = [
     'jupyterhub==1.0.*',
     'tornado==6.0.*',
-    'graphene-tornado==2.1.*'
+    'graphene-tornado==2.1.*',
+    ('cylc @ https://github.com/dwsutherland/cylc'
+     '/tarball/protobuf-uis-feed#egg=cylc-8.0a1.dev')
```
kinow: Sorry, there is one more thing necessary to get `setup.py` working.

If you install this branch with `pip install -e .` everything should work. But if you try `pip install .` it will fail to run. That's because editable installations are links back to the project source, which gets included in the `sys.path`.

Whereas installing via `pip install .`, it will read the package contents, and include only what's described there. In our case, because we started without a package/folder, we had to manually list what was included.

See further down in this file, and you should find a key in `setup(` similar to this one:

```
setup(
...
py_modules=['handlers', 'cylc_singleuser'],
...
)
```

We need to list each file that will be included in the package, manually, one by one there. Painful I know, but once the GraphQL work is merged we can now create `cylc.uiserver` structure, and re-organize files.

schema.py
```diff
@@ -0,0 +1,934 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""GraphQL API schema via Graphene implementation."""
+
+from graphene import (
+    Boolean, Field, Float, ID, InputObjectType, Int,
+    List, Mutation, ObjectType, Schema, String, Union
+)
+from graphene.types.resolver import dict_resolver
```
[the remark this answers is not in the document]
dwsutherland: Done

cylc_singleuser.py
```diff
@@ -44,6 +47,42 @@ def try_exit(self):
             logging.info('exit success')
 
 
+# This is needed in order to pass the server context in addition to existing.
+# It's possible to just overwrite TornadoGraphQLHandler.context but we would
+# somehow need to pass the request info (headers, username ...etc) in also
+class UIServerGraphQLHandler(TornadoGraphQLHandler):
```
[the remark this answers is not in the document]
dwsutherland: Done

cylc_singleuser.py
dwsutherland: I do some one initial start up, and yes still done every 19 and 5 secs respectively data manager one will likely change/be-removed with incremental updates.

data_mgr.py
```diff
@@ -0,0 +1,238 @@
...
+async def get_workflow_data(w_id, client, method):
+    # Use already established client
+    try:
+        pb_msg = await client.async_request(method)
+    except ClientTimeout as exc:
+        return (w_id, MSG_TIMEOUT)
+    except ClientError as exc:
+        return (w_id, None)
+    else:
+        ws_data = METHOD_MAP[method]()
```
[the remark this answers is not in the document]
dwsutherland: We may/will end up using more than just one method in the request:
```
async def get_workflow_data(w_id, client, method):
    # Use already established client
    try:
        pb_msg = await client.async_request(method)
    except ClientTimeout as exc:
        return (w_id, MSG_TIMEOUT)
    except ClientError as exc:
        return (w_id, None)
    else:
        ws_data = METHOD_MAP[method]()
        ws_data.ParseFromString(pb_msg)
    return (w_id, ws_data)
```

And I need to map the method to the protobuf message... I need this because I cannot read the bytes content without know what it is.

data_mgr.py
dwsutherland: Oppsie... I was looking at the speed of the individual action:
https://stackoverflow.com/a/15456218
Yes updating all at once would be better...

data_mgr.py
dwsutherland: Can't use generators within the gather, but works with:
```
        for args in ws_args:
            if not ids or args[0] in ids:
                gathers += (get_workflow_data(*args),)
        items = await asyncio.gather(*gathers)
```

data_mgr.py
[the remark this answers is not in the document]
kinow: Yup, but do we need the protobuf message type as text, instead of the Python object? i.e., instead of `get_workflow_data` receiving a `client` (which I believe is a `: cylc.flow.network.client.SuiteRuntimeClient` instance), and a `method: string`, this last one, could it be `method: object`? So instead of passing `"PbPoolTask"`, I would have to `from ws_messages_pb2 import PbPoolTask` and pass the object `PbPoolTask`. If so, we wouldn't need to dict with `text -> object` mapping?

cylc_singleuser.py
kinow: Ok, did some testing, and the function is called, but I believe it does not need to be `async`. It is not resolved with `await`, or executed in the main loop. It is [simply returned](https://github.com/tornadoweb/tornado/blob/f65b1802a549bab73b4a6721bbfc47f82954d156/tornado/ioloop.py#L907).

data_mgr.py
dwsutherland: All workflows.

data_mgr.py
dwsutherland: done.

data_mgr.py
dwsutherland: added # TODO  .. there might be further filtering on results.

kinow:
@dwsutherland with the following query

```graphql
query {
  tasks {
    id, name, meta {
      title
      description
      URL
    }, meanElapsedTime, namespace, proxies {
      id
      task {
        id
      }
      state
      cyclePoint
      spawned
      depth
      jobSubmits
      latestMessage
      outputs
      prerequisites {
        expression
        satisfied
      }
      jobs {
        id
      }
      parents {
        id
      }
    }, depth
  }
}
```

am getting the error

```graphql{
  "errors": [
    {
      "message": "Unknown server error"
    }
  ]
}
```

The console running `jupyterhub` with `cylc-uiserver` has:

```bash
ERROR:tornado.application:Error: Object of type RepeatedScalarContainer is not JSON serializable Traceback (most recent call last):
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/graphene_tornado/tornado_graphql_handler.py", line 95, in post
    yield self.run('post')
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/gen.py", line 729, in run
    value = future.result()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/gen.py", line 736, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/graphene_tornado/tornado_graphql_handler.py", line 112, in run
    result, status_code = yield self.get_response(data, method, show_graphiql)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/gen.py", line 729, in run
    value = future.result()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/gen.py", line 742, in run
    yielded = self.gen.send(value)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/graphene_tornado/tornado_graphql_handler.py", line 208, in get_response
    result = self.json_encode(response, pretty=self.pretty or show_graphiql)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/graphene_tornado/tornado_graphql_handler.py", line 272, in json_encode
    return json.dumps(d, separators=(',', ':'))
  File "/home/kinow/anaconda3/lib/python3.7/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/home/kinow/anaconda3/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/home/kinow/anaconda3/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/home/kinow/anaconda3/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type RepeatedScalarContainer is not JSON serializable

ERROR:tornado.access:500 POST /user/kinow/graphql/graphiql? (127.0.0.1) 13.41ms
```

Is it something you are aware that could happen? I built the query in the browser, using GraphiQL's auto complete feature.

Thanks!

kinow:
Appears to be the `outputs` field that is causing the error on the server side.

kinow:
Another dummy user support question @dwsutherland (sorry!). At the moment we have 2 REST calls to populate the Suite Tasks.

One to retrieve the current Suite, then one more to retrieve the Tasks in the suite. But as we are using GraphQL, I decided to try and use a single query now too.

```graphql
query ($name: String!) {
  workflows (name: $name) {
    id 
    name 
    owner 
    host 
    port 
    tasks {
      id, name, meta {
        title
        description
        URL
      }, meanElapsedTime, namespace, proxies {
        id
        task {
          id
        }
        state
        cyclePoint
        spawned
        depth
        jobSubmits
        latestMessage
      }, depth
    }
  }
}
```

However, I am not able to pass the `name` variable. It fails with

```
{
  "errors": [
    {
      "message": "Unknown argument \"name\" on field \"workflows\" of type \"Query\".",
      "locations": [
        {
          "line": 2,
          "column": 14
        }
      ]
    }
  ]
}
```

Just wondering if there is anything special that needs to be done on the Resolver to accept variables? Or am I able to use a variable for any field in the Query?

image

dwsutherland:
@kinow 
> ```graphql
> query ($name: String!) {
>   workflows (name: $name) {
>     id 
>     name 
>     owner 
>     host 
>     port 
>     tasks {
>       id, name, meta {
>         title
>         description
>         URL
>       }, meanElapsedTime, namespace, proxies {
>         id
>         task {
>           id
>         }
>         state
>         cyclePoint
>         spawned
>         depth
>         jobSubmits
>         latestMessage
>       }, depth
>     }
>   }
> }
> ```

Two things:

a) Like the error suggests. there is no argument `name`, you could use  `workflows (ids: [$name])`.. as the ID `owner/name:state` argument will work for globs or just name.  (this is the cause of the error)
b) `workflows{tasks{proxies{task}}}}` is circular, so no point in getting task from within tasks.

With cylc-ui use `/graphql`  not `/graphql/graphiql`

I'll try the other error you presented.. I often get browser cache issues when restarting the ui-server..

dwsutherland:
outputs bug fixed.

kinow:
Thanks a lot @dwsutherland ! Will update the code with your suggestions and rebase the code to try it again.

dwsutherland:
Mutations added last week. Some examples:
# ---------------------------------------------------
```
mutation {
  stopWorkflow(
      command: "set_stop_after_clock_time",
      workflows: ["*/baz"],
      args: {datetimeString: "2019-05-30T21:24:06+12:00"}) {
    result
  }
}
```
# ---------------------------------------------------
```
mutation{
  holdWorkflow(
      command: "hold_after_point_string",
      workflows: ["baz"],
      pointString: "20171201T0000+12") {
    result
  }
}
```
```
mutation {
  releaseWorkflow(workflows: ["baz"]){
    result
  }
}
```
# ---------------------------------------------------
```
mutation{
  putMessages(
      workflows: ["baz"],
      ids: ["20170101T0000+12/foo/01"],
      messages: [["CRITICAL", "HELLO, IS THERE ANYONE OUT THERE?"]]) {
    result
  }
}
```
# ---------------------------------------------------
```
mutation {
  taskActions(
      command: "insert_tasks",
      workflows: ["baz"],
      ids: ["20180501T0000+12/foo"]) {
    result
  }
}
```

```
mutation {
  taskActions(
      command: "trigger_tasks",
      workflows: ["baz"],
      ids: ["2017*/foo:failed"]) {
    result
  }
}
```
# ---------------------------------------------------
```
mutation{
  putBroadcast(
    workflows: ["baz"],
    namespaces: ["foo"],
    settings: [{environment: {DANGER: "Dr Robertson"}}]) {
    result
  }
}
```

```
mutation{
  clearBroadcast(
      workflows: ["baz"],
      namespaces: ["foo"],
      cancelSettings: [{environment: {DANGER: "Dr Robertson"}}]) {
    result
  }
}
```

data_mgr.py
kinow: Is it solved in your local branch @dwsutherland ? Just synced branches and noticed I had to alter the imports locally again to get it working.

matthewrmshin:
HI @dwsutherland We are going to try to crowd review and merge this and cylc/cylc-flow#3122 while @hjoliver and @kinow are in UK. Please resolve the conflict of this branch when you have the time to do so.

dwsutherland:
> HI @dwsutherland We are going to try to crowd review and merge this and [cylc/cylc-flow#3122](https://github.com/cylc/cylc-flow/pull/3122) while @hjoliver and @kinow are in UK. Please resolve the conflict of this branch when you have the time to do so.

Done.

I've got a few important changes inbound for both PR, will try and get them in before.

dwsutherland:
Travis failing with:
```
Collecting cylc-flow==8.* (from cylc-uiserver==1.0)
  Could not find a version that satisfies the requirement cylc-flow==8.* (from cylc-uiserver==1.0) (from versions: 8.0a0)
No matching distribution found for cylc-flow==8.* (from cylc-uiserver==1.0)
The command "pip install .[all]" failed and exited with 1 during .
```
@kinow confirmed it will be like this until both PRs are merged.. (right?)

setup.py
```diff
@@ -23,10 +23,12 @@
 with open(path.join(here, 'README.md'), encoding='utf-8') as f:
     long_description = f.read()
 
+# TODO: update cylc/cylc-flow dependency after cylc-flow PR has been merged
 install_requires = [
     'jupyterhub==1.0.*',
     'tornado==6.0.*',
-    'graphene-tornado==2.1.*'
+    'graphene-tornado==2.1.*',
+    'cylc-flow==8.*'
```
kinow: Sorry David, @oliver-sanders had a good idea to actually make `cylc-uiserver` to always point to `cylc-flow`'s `master` branch. Could we change it back?

kinow: I think it needs to be 

```python
    ('cylc @ https://github.com/cylc/cylc-flow'
     '/tarball/master#egg=cylc-8.0a1.dev')
```

The egg part with the version does not really matter much

dwsutherland: Install fails
```
Collecting cylc@ https://github.com/cylc/cylc-flow/tarball/master#egg=cylc_flow-8.0a1.dev from https://github.com/cylc/cylc-flow/tarball/master#egg=cylc_flow-8.0a1.dev (from cylc-uiserver==1.0)
  Downloading https://github.com/cylc/cylc-flow/tarball/master (699kB)
     |████████████████████████████████| 706kB 329kB/s 
  WARNING: Generating metadata for package cylc produced metadata for project name cylc-flow. Fix your #egg=cylc fragments.
```
```
ERROR: cylc-uiserver 1.0 requires cylc@ https://github.com/cylc/cylc-flow/tarball/master#egg=cylc_flow-8.0a1.dev, which is not installed.
```

kinow: I think maybe

```
('cylc @ https://github.com/cylc/cylc-flow'
     '/tarball/master#egg=cylc-flow-8.0a1.dev')
```

The egg name was cylc... I think it should work if we just add the cylc-flow part?

dwsutherland: The following seems to work:
```
('cylc-flow @ https://github.com/cylc/cylc-flow'
     '/tarball/master#egg=cylc-8.0a1.dev')
```

dwsutherland:
Ok pushed, however, Travis may still break as it imports `cylc-flow` modules not yet in master..

hjoliver:
MERGING! :boom:  thanks @dwsutherland :bouquet: We'll put up a bunch of Issues for urgent follow-up...

dwsutherland:
:beers:  Thanks everyone for your input and critique!

kinow:
:clap: :clap: :clap: Here's to you @dwsutherland :beers: ! Well done!

PR #35, opened 2019-04-24
Specify the versions in setup.py

### LINKED ISSUE #9: Upgrade to Tornado 6
https://www.tornadoweb.org/en/stable/releases/v6.0.0.html

Might be a good idea to freeze pip and use specific versions in the dependencies list too.

kinow:
Close #9 

This pull request makes sure we have the requirements versions well defined in our setuptools configuration.

Notice that `jupyterhub` has a requirement on `tornado >= 5.0`. Which means that if you have `tornado==5.0`, and happens to also have `jupyterhub==0.9.6` (their latest release), the environment requirements would be OK. Even though there is no guarantee JupyterHub 0.9.6 works with Tornado 5.x any more (they upgraded to 6 and even reported back bugs in the WebSockets features, so they may have changed their API).

We are doing the same in the pull request to add `setup.py` to Cylc, with the argument that having a clear picture of what are our requirements, should gives us less headaches for maintaining the project (e.g. users reporting that it is not working for them, but in reality they have an outdated dependency).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,9 +24,9 @@ with open(path.join(here, 'README.md'), encoding='utf-8') as f:
     long_description = f.read()
 
 install_requires = [
-    'jupyterhub==0.9.*',
+    'jupyterhub==1.0.*',
     'tornado==6.0.*',
-    'graphene-tornado==2.0.*'
+    'graphene-tornado==2.1.*'
 ]
 
 setup_requires = [
--- a/setup.py
+++ b/setup.py
@@ -24,19 +24,19 @@ with open(path.join(here, 'README.md'), encoding='utf-8') as f:
     long_description = f.read()
 
 install_requires = [
-    'jupyterhub',
-    'tornado',
-    'graphene-tornado'
+    'jupyterhub==0.9.*',
+    'tornado==6.0.*',
+    'graphene-tornado==2.0.*'
 ]
 
 setup_requires = [
-    'pytest-runner'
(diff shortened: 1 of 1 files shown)


kinow:
We can do this before or after #34 if it's easier. I'd be happy to update the versions for the new dependencies 👍

kinow:
Versions updated, rebased PR as well. Thanks!!!

image

setup.py
```diff
@@ -24,19 +24,19 @@
     long_description = f.read()
 
 install_requires = [
-    'jupyterhub',
-    'tornado',
-    'graphene-tornado'
+    'jupyterhub==1.0.*',
+    'tornado==6.0.*',
+    'graphene-tornado>=2.0.3,<2.1.0'
```
dwsutherland: Wouldn't that be `>=2.0.3,<=2.1.0`?

kinow: I thought our current code was not ready for 2.1.0 yet, and that maybe you could bump it up to 2.1.* or so in your branch

dwsutherland: Yes, could do. I think it might work as is in this branch though, because it was just my subclassing of the handler that was the issue.

kinow: Let's bump up to 2.1.* here then, and either merge your branch or cherry-pick the changes and merge it earlier if necessary 🚀

dwsutherland: 👍

PR #37, opened 2019-05-08
Simplify headers, with a common class, and ensure jupyterhub version header is present

kinow:
Moved the JupyterHub handlers to the `handlers.py` file (which is common in Tornado apps, e.g. see JupyterHub and Notebook).

Then also fixed an old warning about missing JupyterHub version.

```
[W 2019-05-09 10:50:58.516 JupyterHub _version:34] Single-user server has no version header, which means it is likely < 0.8. Expected 1.0.0
```

This was fixed in the main handler. But as we added more handlers in an ad-hoc way (myself, mainly :grimacing:), we ended up with some duplicated code for CORS headers, and the jupyterhub header was set only in the main handler (not in the scan, or user profile).

This PR moves all the CORS and JupyterHub headers to `BaseHandler`, which is used by the children handlers. Local manual tests passed fine.

### DIFF
--- a/cylc_singleuser.py
+++ b/cylc_singleuser.py
@@ -17,49 +17,17 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import argparse
-import json
 import logging
 import os
 import signal
 
 from graphene_tornado.schema import schema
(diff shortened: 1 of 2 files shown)


matthewrmshin:
Looks good by :eyes:.

PR #39, opened 2019-05-10
Fix 404 after jupyterhub 1.0.0

kinow:
Finally found a way to fix this one. Tested locally.

It worked before when the button was pointing to `/user/$USER/` with the trailing slash. Now it copies the behaviour of jupyterhub's [`AddSlashHandler`](https://github.com/jupyterhub/jupyterhub/commit/ad7867ff11426755ec752ed5b5fbfabd1d2462c3).

### DIFF
--- a/cylc_singleuser.py
+++ b/cylc_singleuser.py
@@ -23,11 +23,11 @@ import signal
 
 from graphene_tornado.schema import schema
 from graphene_tornado.tornado_graphql_handler import TornadoGraphQLHandler
-from jupyterhub.services.auth import HubOAuthCallbackHandler
-from jupyterhub.utils import url_path_join
 from tornado import web, ioloop
 
 from handlers import *
+from jupyterhub.services.auth import HubOAuthCallbackHandler
+from jupyterhub.utils import url_path_join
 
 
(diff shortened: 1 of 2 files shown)


kinow:
And later if we keep using JupyterHub as a dependency in the ui server (again, there is work-in-progress that may eliminate that dependency by @MartinRyan), we can remove our implementation and simply import theirs.

sadielbartholomew:
Sorry @kinow, I am too late to the game on this one as I see it has already been merged, buy FYI & for future, there is [a useful Tornado decorator, ``@tornado.web.addslash``](https://www.tornadoweb.org/en/stable/web.html?highlight=slash#tornado.web.addslash), which will automatically add in a ``/`` to the URL if not present.

(I discovered it & used it in my recent work for a Rose service (PR metomi/rose#2288 if you are interested, see https://github.com/metomi/rose/blob/2dd607f6ca8ed251c2559265a9e58a7fed2fe9f0/lib/python/rosie/ws.py#L166-L168.)

Neat custom solution for the problem though!

kinow:
I think that worked @sadielbartholomew ! Much simpler, thanks!

PR #40, opened 2019-05-10
Use tornado's addslash decorator

kinow:
@sadielbartholomew found a neater solution for the missing slashes instead of the one implemented in #39. One that if that works we can even suggest to JupyterHub (which is a nice thing to do, as we are using so much of their work for free).

[Docs here](https://www.tornadoweb.org/en/stable/web.html#tornado.web.addslash) for the decorator. @sadielbartholomew , could you review this one? Here are the steps to reproduce:

- Check out this pull request
- Remove the annotation
- Create a venv/conda environment
- `pip install -e .` (editable so you don't have to re-install later)
- Start `cylc/cylc-ui` in another directory with `npm run build:watch`, copy the directory where `cylc-ui` is now running
- Edit `jupyterhub_config.py`, looking for `c.Spawner.args` and specify the location of the `cylc-ui/dist` directory (absolute path might be easier)
- Start `cylc-uiserver` by running `jupyterhub`
- Go to `http://localhost:8000` and log in with your local user (it will use PAM to authenticate). You should either see a new page that is either a page with two buttons ("Stop My Server", "My Server"), or a page with one button ("Launch Server"), or the `cylc-ui` landing page.
- If you are on the `cylc-ui` landing page, there should be a link to the "Hub" on the menu to the left, at the bottom
- From the Hub page, after logged in, you should see a blue button that says "My Server". Hover over the link, and you should notice that the link is something like `/user/$yourusernoslash`
- Click on that button, and confirm it works. If you would like to try, you can remove the decorator `addslash` from the `handlers.py` `MainHandler` class. Restarting `jupyterhub`, and re-doing the previous steps, instead of showing the `cylc-ui` landing page, you should see either a blank page or a 404 (it might be blank, because now instead of `/`, we followed the docs of `addslash` and used `/?`, so the `/` is optional, but then it is also matching static paths, causing a major confusion to the roles).
- If you reached this point, you are ready to review this PR with aye/nay :) one review should be enough for this one for now.

Cheers
Bruno

### DIFF
--- a/cylc_singleuser.py
+++ b/cylc_singleuser.py
@@ -86,7 +86,7 @@ class CylcUIServer(object):
                     self._jupyter_hub_service_prefix, 'suites'),
                     CylcScanHandler),
 
-                (self._jupyter_hub_service_prefix,
+                (rf"{self._jupyter_hub_service_prefix}?",
                     MainHandler, {"path": self._static}),
 
                 # graphql
@@ -101,13 +101,7 @@ class CylcUIServer(object):
                 (url_path_join(self._jupyter_hub_service_prefix,
                                '/graphql/graphiql'),
                     TornadoGraphQLHandler,
-                    dict(graphiql=True, schema=schema)),
-
-                # since jupyterhub 1.0.0, the "My Server link the hub page
-                # removed the last slash from the URL. This handler is based
-                # on the handler with the same name in jupyterhub 1.0.0,
-                # and adds back the last `/` to every URL in the app.
-                (r".*", AddSlashHandler)
+                    dict(graphiql=True, schema=schema))
             ],
             # FIXME: decide (and document) whether cookies will be permanent
             # after server restart.
--- a/handlers.py
+++ b/handlers.py
@@ -19,8 +19,6 @@ import os
 import re
 from subprocess import Popen, PIPE
 from typing import List, Union
-from urllib.parse import urlparse
-from urllib.parse import urlunparse
 
 from jupyterhub import __version__ as jupyterhub_version
 from jupyterhub.services.auth import HubOAuthenticated
@@ -52,6 +50,7 @@ class MainHandler(HubOAuthenticated, BaseHandler):
     def initialize(self, path):
         self._static = path
 
+    @web.addslash
     @web.authenticated
     def get(self):
         """Render the UI prototype."""
@@ -102,17 +101,8 @@ class CylcScanHandler(HubOAuthenticated, APIHandler):
         self.write(json.dumps(suites))
 
 
-class AddSlashHandler(web.RequestHandler):
-
-    def get(self, *args):
-        src = urlparse(self.request.uri)
-        dest = src._replace(path=src.path + '/')
-        self.redirect(urlunparse(dest))
-
-
 __all__ = [
     "MainHandler",
     "UserProfileHandler",
-    "CylcScanHandler",
-    "AddSlashHandler"
+    "CylcScanHandler"
 ]


kinow:
I used incognito mode as the browser seemed to be getting 30x requests and using cached responses instead of rendering the blank page or 404 page. Just in case that helps too.

sadielbartholomew:
Hi @kinow, I'm very happy to review (& generally to start getting involved on the UI- & server- side).

Thank you for providing such detailed instructions for recreating the issue at hand. Unfortunately, I've started reviewing the PR this afternoon & I'm struggling with some of the steps, very likely because I have mis-interpreted a step or something similar meaning there are differences which make it hard to understand the context of later steps.

I'll try to cover the main points where I might be going wrong, as  follows. Sorry for all the questions - if you can see why I might not be setting up correctly & being unsure about how to follow your steps, please feel free to skip to the relevant questions & ignore the others!

****

> Create a venv/conda environment

In case it is relevant to any possible differences we are seeing, I am using a Conda Python-3 environment, & using ``npm`` via ``nodejs`` (version 11.14.0 on the ``conda-forge`` channel).

> Remove the annotation

I skipped this one as I am not sure what annotation you mean? I suppose you mean the decorator that is the key bit to the PR? In which case this shouldn't be relevant to the set-up difficulty at hand; of course I need to eventually test both ``master`` & this branch to compare results for reviewing.

> Start cylc/cylc-ui in another directory with npm run build:watch, copy the directory where cylc-ui is now running

I can only get that ``npm`` command to work from the ``cylc-ui`` repository. Perhaps this is because I used ``npm install`` there? Do both these commands need to be run in the same directory? And by 'in another directory' do you mean any old directory created to contain the build or serve some similar purpose?

> Edit jupyterhub_config.py, looking for c.Spawner.args and specify the location of the cylc-ui/dist directory (absolute path might be easier)

As described in #41, which I opened in parallel to this reviewing, I am unsure as to what the ``dist/`` directory is. Please can you explain a bit further (either on that or this PR)?

> Go to http://localhost:8000 and log in with your local user (it will use PAM to authenticate). You should either see a new page that is either a page with two buttons ("Stop My Server", "My Server"), or a page with one button ("Launch Server"), or the cylc-ui landing page.
>
> If you are on the cylc-ui landing page, there should be a link to the "Hub" on the menu to the left, at the bottom
>
> From the Hub page, after logged in, you should see a blue button that says "My Server". Hover over the link, and you should notice that the link is something like /user/$yourusernoslash

After logging in, no port ``8000`` I get the 'Launch Server' page and when I launch it after the progress bar gets to the end I briefly see a ``user/sbarth`` item, but after less than a second or so it loads a blank page (probably I imagine because I am pointing at the wrong place for the ``cylc-ui/dist`` as above, I am using ``<...full-path...>/cylc-ui-test/`` which is the copied directory from where ``cylc-ui`` was running).

> If you are on the cylc-ui landing page, there should be a link to the "Hub" on the menu to the left, at the bottom

I didn't get to see this, so I guess I have gone astray somewhere in the above...

Possibly relevant is that there is another Hub 'route' created automatically on port ``8081``:

```console
[I 2019-05-13 19:24:10.084 JupyterHub proxy:645] Starting proxy @ http://:8000
19:24:11.357 [ConfigProxy] info: Proxying http://*:8000 to (no default)
19:24:11.364 [ConfigProxy] info: Proxy API at http://127.0.0.1:8001/api/routes
19:24:11.855 [ConfigProxy] info: 200 GET /api/routes 
[I 2019-05-13 19:24:11.855 JupyterHub proxy:319] Checking routes
[I 2019-05-13 19:24:11.856 JupyterHub proxy:399] Adding default route for Hub: / => http://127.0.0.1:8081
19:24:11.860 [ConfigProxy] info: Adding route / -> http://127.0.0.1:8081
19:24:11.861 [ConfigProxy] info: Route added / -> http://127.0.0.1:8081
19:24:11.862 [ConfigProxy] info: 201 POST /api/routes/ 
```

 When I try the same steps for the equivalent URL endpoints of that route, noting I am using your branch here not master, I get a ``500 : Internal Server Error; Redirect loop detected.`` message on the rendered page, & a repeated (with ``redirects=N`` increasing from ``N=1 to 4``) log message to STDOUT of:

```console
[I 2019-05-13 19:33:08.933 JupyterHub log:174] 302 GET /hub/user/sbarth/?redirects=3 -> /user/sbarth/?redirects=N (sbarth@127.0.0.1) 8036.32ms
[I 2019-05-13 19:33:08.944 JupyterHub log:174] 302 GET /user/sbarth/?redirects=4 -> /hub/user/sbarth/?redirects=N (@127.0.0.1) 2.07ms
[W 2019-05-13 19:33:09.002 JupyterHub web:1782] 500 GET /hub/user/sbarth/?redirects=N (127.0.0.1): Redirect loop detected.
```
which does suggest to me there might be an issue on this branch. But I need to make sure my set-up is correct & I understand it enough before I can judge that.

****

Any help on the above is very appreciated. As a side note, while troubleshooting I am enjoying getting truly stuck in with the nitty-gritty of the UI & UI Server!

kinow:
>of course I need to eventually test both master & this branch to compare results for reviewing.

That was myself being lazy, and testing by commenting out the `addslash` decorator :grimacing: much better if you can switch `master` and the branch for this PR instead.

>I can only get that npm command to work from the cylc-ui repository. Perhaps this is because I used npm install there? Do both these commands need to be run in the same directory? And by 'in another directory' do you mean any old directory created to contain the build or serve some similar purpose?

Sorry, I think it was a typo. Probably I should have said "start cylc-ui in another terminal". You are doing it correctly.

>As described in #41, which I opened in parallel to this reviewing, I am unsure as to what the dist/ directory is. Please can you explain a bit further (either on that or this PR)?

Sure. When you run `npm run build` or `npm run build:watch`, it will build the Vue.js project, and output its contents into the `dist` folder. That is specified in the [`vue.config.js`](https://github.com/cylc/cylc-ui/blob/d9c63955b688bd5b19b7ac97016a8f26e6fb2206/vue.config.js#L5).

So in your `jupyterhub_config.py`, you will have to set the `c.Spawner.args` to point to the location of the Vue.js/WebPack generated `dist` folder.

Could you check if doing this you are able to progress a bit more and fix the blank page, please? The nice thing is that once you review this PR, looks like you will have a complete environment for testing other PR's or even changing the code :grimacing: 

Thanks @sadielbartholomew !

kinow:
PR rebased onto `master` to fix conflicts :+1:

sadielbartholomew:
Thanks for fixing the conflicts @kinow. Just to give you an indication of timescale, I have been quite busy lately with user support & reviewing (etc.) but I will try to get this reviewed tomorrow. Sorry for the delay, & thanks for your patience :+1:

kinow:
No hurry on this one @sadielbartholomew . I had finished reviewing GraphQL, so was just going through my pull requests to repos in github and making sure they were ready for review :) Thanks!

kinow:
No problems, and thanks a lot @sadielbartholomew !

kinow:
I think the same approach can be used in JupyterHub too: https://github.com/jupyterhub/jupyterhub/pull/2609

PR #49, opened 2019-05-29
Use argparse parse_known_args to allow extra params/args

### LINKED ISSUE #48: Implement remaining parameters available for the singleuser app
When the hub starts the `singleuser` app, it has a series of parameters that it may send, e.g. `debug`.

Right now, we are only supporting a few (port, static folder), and raising an issue if any other parameter is provided.

We must ensure it works even if we don't use the values (i.e. probaly use args/kwargs).

kinow:
Close #48 

Tested by running `$ cylc-singleuser -s $(pwd)/../cylc-ui/dist/ -p 9999`, which works fine.

The tested the scenario that is broken:

```bash
$ cylc-singleuser -s $(pwd)/../cylc-ui/dist/ -p 9999 --debug
usage: cylc-singleuser [-h] [-p PORT] -s STATIC
cylc-singleuser: error: unrecognized arguments: --debug
```

With the fix in this PR, the `--debug` is simply ignored (or any other value passed by the `jupyterhub` command, after reading the `jupyterhub_config.py` settings).

Cheers
Bruno

### DIFF
--- a/cylc_singleuser.py
+++ b/cylc_singleuser.py
@@ -133,7 +133,7 @@ def main():
                         default=8888)
     parser.add_argument('-s', '--static', action="store", dest="static",
                         required=True)
-    args = parser.parse_args()
+    args = parser.parse_known_args()[0]
 
     jupyterhub_service_prefix = os.environ.get(
         'JUPYTERHUB_SERVICE_PREFIX', '/')


cylc_singleuser.py
```diff
@@ -133,7 +133,7 @@ def main():
                         default=8888)
     parser.add_argument('-s', '--static', action="store", dest="static",
                         required=True)
-    args = parser.parse_args()
+    args, _ = parser.parse_known_args()
```
matthewrmshin: Is `_` assigned to the unknown arguments here?

kinow: Yes. Used it as otherwise the IDE linter complains about an unused variable. But happy to change it to a more descriptive value if you think it's better

matthewrmshin: Can you do this instead?

```python
args = parser.parse_known_args()[0]
```

On the other note, in your original description, you mentioned the `--debug` option. I guess this is not used yet? Ever?

kinow: >Can you do this instead?

Done! Good idea!

>On the other note, in your original description, you mentioned the --debug option. I guess this is not used yet? Ever?

Not yet. There are other settings in the `jupyterhub_config.py` like `c.JupyterHub.hub_connect_ip` (if I recall correctly) that cause the spawner to pass `--ip ?.?.?.?` as parameter.

The current `cylc_singleuser.py` code started with no settings I think. Then we added one for the port, then another for the static content. So we may use debug later when we start doing more logging in the UI Server.

matthewrmshin: Understood.

kinow:
Merging as we are still doing 1 approval in cylc-uiserver & cylc-ui :tada:

PR #53, opened 2019-06-28
WS network imports

dwsutherland:
By importing more WS network code, this pull request syncs the `cylc-flow` and `cylc-uiserver` GraphQL schema and resolver filters.. This will make divergence impossible, and pave the way for a CLI unified across WS and UIServer.

Task:
- [x] Import GraphQL, resolver filters, and other related network code.

(also fixed a workflow gathers client swapping risk)


dwsutherland:
> This looks good to me, but I don't have an environment to test (or time, about to go to airport). @hjoliver would you like to try this and see if the end to end still works?

BTW - The schema was out of date, so this will fix that at least.. (ghost true/false arg, and tree starts from within the workflow root query (if we use that))

dwsutherland:
🎉

dwsutherland:
Thanks!

PR #55, opened 2019-07-08
Use Cylc logo in JupyterHub.

hjoliver:
The logo is configurable in JupyterHub.

(Currently assumes relative path to cylc-ui project, as we already do elsewhere in the config file; this might not be the permanent solution to that problem?)

We might have to contribute a small change to JupyterHub to make the X'd words similarly configurable:

jhub-logo

### DIFF
--- a/jupyterhub_config.py
+++ b/jupyterhub_config.py
@@ -306,7 +306,7 @@
 #c.JupyterHub.load_groups = {}
 
 ## Specify path to a logo image to override the Jupyter logo in the banner.
-#c.JupyterHub.logo_file = ''
+c.JupyterHub.logo_file = '../cylc-ui/dist/img/logo.png'
 
 ## File to write PID Useful for daemonizing JupyterHub.
 #c.JupyterHub.pid_file = ''


kinow:
Should we ship a copy of the logo in `cylc-uiserver` instead of loading it from `cylc-ui`? I think that path may fail if the user has cylc-ui somewhere else, or if the virtual env directory doesn't match?

>We might have to contribute a small change to JupyterHub to make the X'd words similarly configurable:

+1, here's the file where the text resides. There is a simple JS code at the bottom of that file, that makes it visible when the URL is not using https.

https://github.com/jupyterhub/jupyterhub/blob/276ef261613400ff75c370d50160c3cc88017ff7/share/jupyterhub/templates/login.html#L28

jupyterhub_config.py
```diff
@@ -306,7 +306,7 @@
 #c.JupyterHub.load_groups = {}
 
 ## Specify path to a logo image to override the Jupyter logo in the banner.
-#c.JupyterHub.logo_file = ''
+c.JupyterHub.logo_file = '../cylc-ui/dist/img/logo.png'
```
kinow: Should we ship a copy of the logo in cylc-uiserver instead of loading it from cylc-ui? I think that path may fail if the user has cylc-ui somewhere else, or if the virtual env directory doesn't match?

(just finishing another code, will give it a try in a bit)

hjoliver: Yes we probably should.  I just did it this way as a quick test, as we already refer to `../cylc-ui/dist` in one other place in `jupyterhub_config.py`.

kinow:
Looks really neat!!!

image

If the image is not found, nothing crashes, which is good. It just displays the `alt` (alternative) text.

image

>Should we ship a copy of the logo in cylc-uiserver instead of loading it from cylc-ui? I think that path may fail if the user has cylc-ui somewhere else, or if the virtual env directory doesn't match?

Scratch that idea. We would have to package that too, and JupyterHub loads from a location, not from a Python module. So in the end this would probably be more complicated.

Much simpler to just include as is now, and let users choose whatever logo they'd like (and we may be able to set the logo in the Conda/Ansible/Docker/etc when we have automation for the complete Cylc 8 system)

PR #58, opened 2019-07-10
Fix one bug and add a few minor improvements

kinow:
The line `if cylc.flags.debug:` would result in a runtime error. It does not cause any issue, unless you have a certain exception (or if you pass through that code in a unit/functional test).

The other changes are all minor improvements, like removing unused imports or unused variables.


kinow:
Did a quick manual test, and found no issues after this change :+1:

Issue #47, opened 2019-05-27
user of jupyterhub_config.py


hjoliver:
We've only made minimal changes to `jupyterhub_config.py`, but added our own copyright header.  It seems the default config file generated by `jupyterhub --generate-config` has changed quite a bit at the latest version (of jupyterhub).  Would it be better for us not to include the config file in the repository, but tell users to generate one, and what in it to modify to spawn our UI Server?

hjoliver:
(I see our README does say to generate one).

kinow:
>but added our own copyright header.

Oh, I think I went too far when I added the copyright headers.

> Would it be better for us not to include the config file in the repository, but tell users to generate one, and what in it to modify to spawn our UI Server?

Might be a good idea. Perhaps include that to `.gitignore` too (as it can be convenient to have the file in the project folder).

>(I see our README does say to generate one).

I **think** the README predates the configuration file. I have this vague memory of adding the config during some PoC...

hjoliver:
I wasn't complaining about the copyright header, just noticing that our config file was out of date compared with a new one generated by the latest Jupyterhub release :grin: (so maybe better not to store a copy of the config file?)

oliver-sanders:
#61 has taken out the comments, with a little bit of documentation might that be enough to close this?

hjoliver:
OK, so #61 takes everything out of JHub config except for the absolute minimum that we need.  And given that in releases we will be depending on specific versions of JHub, I think that is the best solution.  So, closing this Issue...

PR #60, opened 2019-07-17
List other UI servers

### LINKED ISSUE #27: List other UI servers running
We will have one UI server per user. And each UI server will look after one or more workflows.

We want to allow users to access each other's UI server and see their running workflows - subjected to proper authorization, of course.

We need to define what is the workflow going to look like. 

- Is user A going to configure his/her workflow to authorize that user B can access it?
- Is user A going to share the link of his UI server with user B, or will user B have a way to automatically see what suites s/he has access to?

kinow:
Closes #27 

This PR adds a new Tornado handler that retrieves the list of HTTP Routes currently available in the Hub's proxy. The data returned is exactly the same data returned by the Configurable HTTP Proxy API.

```json
{"/":{"hub":true,"target":"http://127.0.0.1:8081","jupyterhub":true,"last_activity":"2019-07-17T04:56:59.242Z"},"/user/kinow":{"user":"kinow","server_name":"","target":"http://127.0.0.1:53125","jupyterhub":true,"last_activity":"2019-07-17T04:57:01.021Z"}}
```

I've added some TODO's to the code. One in special is to decide whether we should restrict what information is returned... for now I think we will need the key in the dictionary (i.e. `/user/kinow`). But not sure if we will need the remaining info? I don't even know what `jupyterhub: false` would mean.


kinow:
(the other commit simplifies the configuration, related to #47, but also fixes the annoying relative directory, so others don't have to change it. I am thinking about putting another PR to get it merged quickly so others don't have to keep changing that file for development)

kinow:
In order to fetch the users, first we had to work around the proxy's HTTP auth. The request to the proxy API looks something like:

```bash
$ curl -XGET -H "Authorization: token b9cc537dbe8d8fb4919da43bf429fa4e15498cacc123143eb4ef549858ea" http://localhost:8081/api/routes
```

The authorization token, if not provided in `jupyterhub_config.py`, is generated automatically during the `jupyterhub` application initialization.

And that token is available only in the proxy, as an environment variable of that process (not same process as `jupyterhub` or even `cylc-singleuser), so we can't use that information. It is also available as a property of the Proxy object, but that object is detached from the UI Server (which is spawned via the command argument settings in `jupyterhub_config.py`).

So after thinking for a while this afternoon, I thought that the only other alternative would be to a) define the token in `jupyterhub_config.py` (which means `jupyterhub` will pass it to the proxy when it is started), b) expose the TOKEN as an environment variable to `cylc-singleuser`, and c) also expose the proxy URL (we can't assume it will always be http://localhost:8001, using TraefikProxy or containers normally means a remote URL, and possibly a different port).

Not the best of the alternatives, as now that auth token is available as an environment variable... though JupyterHub exposes its own HTTP auth API already... so maybe that's OK?

kinow:
Quickly hacked `cylc-ui` to use axios to request `window.location.pathname + '/routes` in the `UserProfile` view, and display every user logged in.

kinow:
image

kinow:
And just for myself in case I need any of this, current diff in `cylc-ui`:

```diff
diff --git a/src/views/UserProfile.vue b/src/views/UserProfile.vue
index 7a643c8..6988df0 100644
--- a/src/views/UserProfile.vue
+++ b/src/views/UserProfile.vue
@@ -2,6 +2,14 @@
   <v-container>
     <v-layout justify-center wrap>
       <v-flex xs12 md12>
+        <v-list>
+          <template>
+            <v-subheader>Other users with active UI Servers</v-subheader>
+            <v-list-tile>
+              <v-list-tile-title v-for="userhub in users" :key="userhub"><a :href="userhub">{{ userhub }}</a></v-list-tile-title>
+            </v-list-tile>
+          </template>
+        </v-list>
         <material-card
             color="green"
             title="Your Profile"
@@ -71,16 +79,29 @@
 <script>
 import { UserService } from '@/services/user.service'
 import { mapState } from 'vuex'
+import axios from 'axios'
 export default {
   computed: {
     ...mapState('user', ['user'])
   },
+  data: () => ({
+    users: []
+  }),
   beforeCreate () {
     this.$store
       .dispatch('user/setUser', null)
       .then(() => {
         UserService.getUserProfile()
-      })
+      }).then(() => {
+        axios.get(window.location.pathname + '/routes').then((response) => {
+          for (var key in response.data) {
+            if (key !== '/') {
+              this.users.push(key)
+            }
+          }
+        })
+      }
+      )
   },
   metaInfo () {
     return {
```

Git commit: aa6e75244dd4add7537552e4baf7fe9d904ebe27

kinow:
Closing as I hadn't really understood the issue. See linked issue for comments.

PR #61, opened 2019-07-31
Update config

oliver-sanders:
The jupyterhub_config no-longer works for us as the relative path seems to be causing issues, switch to using an absolute path (calculated from the relative path).

* Relative -> absolute path
* Tidy config file (comments get out of date, we can get commented lines back by re-generating the file)
* Fix README


oliver-sanders:
Need to move the SVG to a file in cylc-ui.

kinow:
Yup. Let's do that after this one gets merged :)

sadielbartholomew:
> Tidy config file

:+1: for that, I would prefer to add in such comments as necessary if they are relevant or useful to our needs, rather than dumping everything that might be so just in case.

kinow:
>+1 for that, I would prefer to add in such comments as necessary if they are relevant or useful to our needs, rather than dumping everything that might be so just in case.

I think that's the idea in #47

Issue #71, opened 2019-09-03
Investigate how to prevent the Tornado app of showing exception traceback info to users


kinow:
Before going to production, we need to make sure users have a way to configure Cylc UI Server so that traceback are not displayed in the browser to users, as it may leak sensitive information.

- https://www.owasp.org/index.php/Exception_Handling#Displaying_Errors
- https://www.owasp.org/index.php/Error_Handling#Generic_error_messages

At the moment the user gets the following if the HTML assets are not found:

![image](https://user-images.githubusercontent.com/304786/64216038-4f16e080-cf0b-11e9-8972-2f9706752587.png)

PR #70, opened 2019-09-01
n-distance edges & nodes, BaseResolvers import

dwsutherland:
This is the accompanying PR to https://github.com/cylc/cylc-flow/pull/3337

**Changes:**
- **Reformed data-store:** This was done to have a uniform access for the resolvers at both WS an UIS.
- **`BaseResolvers` imported:** Contains all resolvers for use by both WS an UIS, with local `Resolvers` class inheriting this.

`BaseResolvers` contains the `nodesEdges` resolver... That and other inherited improvements are detailed in the [parent](https://github.com/cylc/cylc-flow/pull/3337) cylc-flow PR.

PR #73, opened 2019-09-09
Add logging

### LINKED ISSUE #67: Configure logging for the Cylc UI Server app
I added some logging in `main.py` some time ago, while working on the initial prototype. But there was nothing important in those logging messages.

While looking at #65 , I wondered where exceptions were logged. Looks like the GraphQL client code is not logging exceptions (neither is the rest of the application to be honest).

We need to look at how JupyterHub and its spawned applications handle logging, and use it properly in Cylc UI Server. A quick look at the JupyterHub code, looks like they [use](https://github.com/jupyterhub/jupyterhub/blob/915664ede2081838ae18f19806d7b8ecc422eae6/jupyterhub/log.py#L10) some code from [Tornado logging](https://www.tornadoweb.org/en/stable/log.html).

But I am not sure if that's the only way they are handling logging.

Cheers
Bruno

kinow:
Closes #67 

Uses the Python `logging` module. The configuration is kept in an INI file, natively supported. By default there is a `logging_config.ini` with the logging configuration, instead of initializing in the Python code.

This way, in `argparse` we use a `FileType` parameter that defaults to the module's own `logging_config.ini`, but the user should be able to customize the logging via `--logging-config /opt/test/logging.ini` or any other location. The parameter is defined in:

```py
    parser.add_argument('--logging-config', type=argparse.FileType('r'),
                        help='path to logging configuration file',
                        action="store", dest="logging_config",
                        default=join(here, 'logging_config.ini'))
```

We have a default handler outputting to the console, with the common formatter. And loggers for:

- `tornado.access` set to `WARN` level
- `tornado.application` set to `WARN` level
- `cylc.flow` set to `DEBUG` level
- `cylc.uiserver` set to `DEBUG` level

Python code uses the `logging` module by creating loggers with `__name__`, so we can control the logging down to the level of a single file. For example, here's the output after this PR (before it was blank):

```bash
$ cylc-uiserver -s /home/kinow/Development/python/workspace/cylc-ui/dist/ -p 9999
2019-09-09 14:07:03,134 cylc.uiserver.main INFO     JupyterHub Service Prefix: /
2019-09-09 14:07:03,134 cylc.uiserver.main INFO     Listening on 9999 and serving static content from /home/kinow/Development/python/workspace/cylc-ui/dist/
2019-09-09 14:07:03,134 cylc.uiserver.main INFO     Starting Cylc UI
2019-09-09 14:07:03,134 cylc.uiserver.main INFO     /home/kinow/Development/python/workspace/cylc-ui/dist/
^C2019-09-09 14:07:06,887 cylc.uiserver.main INFO     exiting...
2019-09-09 14:07:06,939 cylc.uiserver.main INFO     exit success
```

We could make any log entry from `main.py` to be ignored, unless it is a `WARN` (or higher like `ERROR` etc)

```ini
diff --git a/cylc/uiserver/logging_config.ini b/cylc/uiserver/logging_config.ini
index 8704a32..0c000fb 100644
--- a/cylc/uiserver/logging_config.ini
+++ b/cylc/uiserver/logging_config.ini
@@ -5,7 +5,7 @@ keys=basic
 keys=console
 
 [loggers]
-keys=root, tornado_access, tornado_application, cylc_flow, cylc_uiserver
+keys=root, tornado_access, tornado_application, cylc_flow, cylc_uiserver, cylc_uiserver_main
 
 # Formatters
 [formatter_basic]
@@ -45,3 +45,9 @@ level=DEBUG
 handlers=console
 propagate=0
 qualname=cylc.uiserver
+
+[logger_cylc_uiserver_main]
+level=WARN
+handlers=console
+propagate=0
+qualname=cylc.uiserver.main
```

The Python `logging` module respects the hierarchy, so in `main.py`, creating a `logger` with `__name__`, means that it will use the qualified name `cylc.uiserver.main`, which will match the newly created logger from the example above.

Cheers
Bruno


kinow:
And an example of what happens when we have the default configuration, and something goes wrong in the Tornado application (e.g. invalid request, without the OAuth information):

```bash
$ cylc-uiserver -s /home/kinow/Development/python/workspace/cylc-ui/dist/ -p 9999
2019-09-09 14:25:53,136 cylc.uiserver.main INFO     JupyterHub Service Prefix: /
2019-09-09 14:25:53,137 cylc.uiserver.main INFO     Listening on 9999 and serving static content from /home/kinow/Development/python/workspace/cylc-ui/dist/
2019-09-09 14:25:53,137 cylc.uiserver.main INFO     Starting Cylc UI
2019-09-09 14:25:53,137 cylc.uiserver.main INFO     /home/kinow/Development/python/workspace/cylc-ui/dist/
2019-09-09 14:25:58,799 tornado.application ERROR    Uncaught exception GET / (127.0.0.1)
HTTPServerRequest(protocol='http', host='localhost:9999', method='GET', uri='/', version='HTTP/1.1', remote_ip='127.0.0.1')
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/traitlets/traitlets.py", line 528, in get
    value = obj._trait_values[self.name]
KeyError: 'oauth_client_id'

  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/web.py", line 1697, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/web.py", line 1908, in wrapper
    return method(self, *args, **kwargs)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/web.py", line 3160, in wrapper
    if not self.current_user:
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/web.py", line 1337, in current_user
    self._current_user = self.get_current_user()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/jupyterhub/services/auth.py", line 894, in get_current_user
    user_model = self.hub_auth.get_user(self)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/jupyterhub/services/auth.py", line 496, in get_user
    user_model = self._get_user_cookie(handler)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/jupyterhub/services/auth.py", line 542, in _get_user_cookie
    token = handler.get_secure_cookie(self.cookie_name)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/jupyterhub/services/auth.py", line 531, in cookie_name
    return self.oauth_client_id
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/traitlets/traitlets.py", line 556, in __get__
    return self.get(obj, cls)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/traitlets/traitlets.py", line 535, in get
    value = self._validate(obj, dynamic_default())
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/traitlets/traitlets.py", line 593, in _validate
    value = self._cross_validate(obj, value)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/traitlets/traitlets.py", line 599, in _cross_validate
    value = obj._trait_validators[self.name](obj, proposal)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/traitlets/traitlets.py", line 907, in __call__
    return self.func(*args, **kwargs)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/jupyterhub/services/auth.py", line 568, in _ensure_not_empty
    raise ValueError("%s cannot be empty." % proposal.trait.name)
ValueError: oauth_client_id cannot be empty.
2019-09-09 14:25:58,804 tornado.access ERROR    500 GET / (127.0.0.1) 7.86ms
2019-09-09 14:25:58,922 tornado.access WARNING  404 GET /favicon.ico (127.0.0.1) 1.11ms
```

At the moment the stack trace is printed only in the browser (which I learned how to fix while working on this issue, so will send another PR for #71). After this PR, the exception is logged to the console, because we have a logger in the default file for `tornado.application`, and because it is higher or equal than `WARN`.

cylc/uiserver/main.py
```diff
@@ -22,6 +22,8 @@
 import signal
 
 from cylc.flow.network.schema import schema
+from logging.config import fileConfig
```
matthewrmshin: Just noting the from:

https://docs.python.org/3/library/logging.config.html#configuration-file-format

> The fileConfig() API is older than the dictConfig() API and does not provide functionality to cover certain aspects of logging.

Is it worth using `dictConfig` for a better future?

cylc/uiserver/workflows_mgr.py
```diff
@@ -62,7 +65,7 @@
         except socket.error as exc:
             if flags.debug:
                 raise
-            sys.stderr.write("ERROR: %s: %s\n" % (exc, host))
+            logger.error("ERROR: %s: %s\n" % (exc, host))
```
matthewrmshin:
```suggestion
            logger.error("%s: %s", exc, host)
```

(Remove `ERROR: ` and don't format the string upfront for lazy logging.)

cylc/uiserver/main.py
kinow: My initial implementation was with `dictConfig`. I stopped when I realized that it wouldn't be easy to allow users to extend with their own logging configuration.

I would probably have to do something like

```
if user_provided_config:
    fileConfig(...)
else:
    dictConfig(...)
```

But reading the rest of the docs, looks like `dictConfig` is the way to go. Though, looks like we are actually free to load the `dict` from any resource, even [YAML](https://stackoverflow.com/a/49013493).

I will come up with a different approach with `dictConfig`, but without YAML (as this would imply adding a new dependency to the project). Quite sure we can load a dict from an INI file? Or maybe JSON as that's part of Python stdlib I think.

kinow: Changed to load from a JSON file. We could hardcode the dictionary as Python code instead (i.e. a Python dictionary). But I thought perhaps this way we can link the file from the docs, saying something like "for logging in Cylc UI Server, you can change its default configuration with a JSON file. See the project default JSON for an example, and the Python logging page for more..."?

matthewrmshin:
Note discussion at cylc/cylc-flow#2505.

matthewrmshin:
Just noticed that the `jupyterhub` command also has a `--log-level=<Enum>` option. I wonder if we can make use of that when we have the hub spawning the UIS.

dwsutherland:
@kinow - Conflicts with newly merged #70  I think...

kinow:
Thanks @dwsutherland , conflicts solved.

kinow:
Brilliant! We can change the error log to debug or something if that's better, but much quicker to work on this in another issue, you are right. Thanks!!

PR #74, opened 2019-09-09
Add command line option for the Tornado app debug

### LINKED ISSUE #71: Investigate how to prevent the Tornado app of showing exception traceback info to users

kinow:
Closes #71 

The exceptions were logged to the browser due to the [debug mode](https://www.tornadoweb.org/en/stable/guide/running.html#debug-mode-and-automatic-reloading) being turned on for our Tornado application.

This PR simply adds an `argparse` option to enable or disable it (disabled by default). After this, instead of printing the traceback to the user browser, the user instead will see:

image

The traceback won't be lost once we have #73 merged (i.e. just look at the console). This will prevent any security issues during deployment by default.

Cheers
Bruno

### DIFF
--- a/cylc/uiserver/main.py
+++ b/cylc/uiserver/main.py
@@ -61,12 +61,16 @@ class CylcUIServer(object):
         self.resolvers = Resolvers(
             self.ws_mgr, self.data_mgr)
 
-    def _make_app(self):
-        """Crete a Tornado web application."""
+    def _make_app(self, debug: bool):
+        """Crete a Tornado web application.
+
+        Args:
+            debug (bool): flag to set debugging in the Tornado application
+        """
         logging.info(self._static)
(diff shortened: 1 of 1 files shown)


kinow:
To enable in the `jupyterhub_config.py`:

```py
c.Spawner.args = ['-s', str(DIST), '--debug']
```

kinow:
@dwsutherland got a minute to do the final review on this one? Please :grimacing:

PR #76, opened 2019-09-09
Set version back to 0.1

kinow:
In JS normally I see `0.0.0`. But in Python I am not so sure whether 0.1 or 0.1.0 would be the best option. Happy to update the PR if necessary.

Also thought about raising the issue of using a `-dev` or `-something` suffix too, to display a different version for using installing from github (but I didn't do that for the cylc-ui project, so I think we can do it in a next PR)


kinow:
One review should do, if you have time @hjoliver :pray:

PR #79, opened 2019-09-19
setup made dependent on cylc-flow alpha 1

wxtim:


setup.py
```diff
@@ -41,9 +41,7 @@ def find_version(*file_paths):
     'jupyterhub==1.0.*',
     'tornado==6.0.*',
     'graphene-tornado==2.1.*',
-    ('cylc-flow @ https://github.com/cylc/cylc-flow'
-     '/tarball/master#egg=cylc-8.0a1.dev')
-    # 'cylc-flow==8.*'
+    'cylc-flow==8.0a1'
```
kinow: For one moment I thought about having `cylc-flow==8.*`, but for this release of `cylc-uiserver` I think we need to lock on `8.0a1` as that's the only one which the protobuf/graphql/nodes/etc will work. So :+1: from me

PR #81, opened 2019-09-26
Add subscriptions

### LINKED ISSUE #14: Add GraphQL subscriptions support via WebSocket
See #11 for previous discussion. The UI Server is a Tornado project, that inherits classes from JupyterHub Tornado project (for things like authentication).

In #11 and #13 it's getting an extra layer with Graphene, an open source library for Python that implements GraphQL.

However, GraphQL subscriptions require the base layer of Graphene for GraphQL, plus an extra library to handle the query subscription (which is basically a mechanism for sending to clients the output of a query whenever if changed), the normal way being via WebSockets.

WebSockets are supported by Tornado out of the box. What we are missing, is actually the extra layer of integration with Tornado. That is done via the Graphele `graphql-ws` project, which has one Pull Request to add support to tornado still [pending](https://github.com/graphql-python/graphql-ws/pull/25).

Once that pull request is merged, we can add that library in our `setup.py`, and fix this issue. An alternative in case we need that before the pull request is merged, is simply fork it and use our own until it is merged.

kinow:
Closes #14 

- Will use a file that is still under review, but we have tested and it works for what we need (and looks like the code review will take much longer in `graphql-ws`)
- Will use the simplest approach available, of just looping and `asyncio.sleep`ing for a few seconds, then sending more data. This can be improved later to send data only when there is data available, but this will be an improvement for later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ python: 3.7
 
 before_install:
   - pip install pycodestyle
-  - pycodestyle --exclude=".git,.tox,__pycache,venv,.eggs,build,jupyterhub_config.py"
+  - pycodestyle --exclude=".git,.tox,__pycache,venv,.eggs,build,jupyterhub_config.py,cylc/uiserver/tornado_ws.py"
 install:
   - pip install .[all]
 script:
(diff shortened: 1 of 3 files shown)


matthewrmshin:
Moved to #82.

PR #82, opened 2019-09-26
Add subscriptions

### LINKED ISSUE #14: Add GraphQL subscriptions support via WebSocket

kinow:
Closes #14 

- Will use a file that is still under review, but we have tested and it works for what we need (and looks like the code review will take much longer in `graphql-ws`)
- Will use the simplest approach available, of just looping and `asyncio.sleep`ing for a few seconds, then sending more data. This can be improved later to send data only when there is data available, but this will be an improvement for later.

Sibling PRs:
- https://github.com/cylc/cylc-ui/pull/280
- https://github.com/cylc/cylc-flow/pull/3390


kinow:
Work in progress, will start adding now the code to have subscriptions. But this was what was more obvious to me that was necessary for this change. More to come in the next days :+1:

dwsutherland:
I think we can add the UIS Subscriptions machinery (modules, endpoints ...etc) to this PR, but develop the GraphQL schema in `cylc-flow`.. 
(even for the simplest case because we can just test/dev locally with `cylc-uiserver` & `cylc-flow`)

This way we design it from the beginning, to be able to plug into either `WS` or `UIS` subscription resolvers.

kinow:
Oh that's helpful. I saw you define resolvers in cylc flow, then extend it here and add only the specific for uiserver. Clever.

So this ia pretty much it for uiserver for now? We leave it like this and now start working on chema changes in cylc flow??

dwsutherland:
`cylc-flow` might need allow_subscriptions sorted in the handler.

dwsutherland:
> Oh that's helpful. I saw you define resolvers in cylc flow, then extend it here and add only the specific for uiserver. Clever.
> 
> So this ia pretty much it for uiserver for now? We leave it like this and now start working on chema changes in cylc flow??

I think so, once you have this one working locally with the simple case in `cylc-flow` (and GraphiQL working).

cylc/uiserver/main.py
```diff
@@ -119,6 +120,10 @@ def _make_app(self, debug: bool):
                     UIServerGraphQLHandler,
```
dwsutherland: Was GraphiQL was using the it's own handler in our example?
https://github.com/kinow/tornado-sandbox/blob/f1b6e75be2081b12025c045d26350801ab65a511/web4/app.py#L14-L16

And I think this can be just `/graphiql` now (as I believe the JS uses both endpoints)

kinow: It was. This way it would serve the custom template for the graphiql endpoint. And this custom template has the React app to consume websockets (which was re-rendering the query whenever a websocket message arrived... in our case a date/time and a fruit I think haha).

I wasn't going to add this to avoid bloating cylc-uiserver unnecessarily (we will have cylc-ui to test it) but maybe it's a good idea, so that we can test the subscriptions without the cylc-ui as well?

kinow:
@dwsutherland I've added the template to render websocket subscriptions in GraphiQL.

Moved everything websockets related under the `cylc.uiserver.websockets`. This package is temporary, excluded in `pycodestyle`, and should either disappear or be simplified once we have the PR merged (could have called it `subscriptions`, but not good naming things :).

The UI is still working, tested with a running workflow. Then going to `/user/kinow/subscriptions` I get the normal message that an HTTP GET is not able to talk WebSockets. And submitting a query like

```graphql
subscription {
  workflows {
    id
  }
}
```
I can see that the message was submitted to the backend, parsed, interpreted, and rejected because the schema is not ready to handle it yet.

image

I think we are done here for now. And now the only missing parts are in the cylc-flow repository. What do you think @dwsutherland ?

cylc/uiserver/main.py
dwsutherland: > I wasn't going to add this to avoid bloating cylc-uiserver unnecessarily (we will have cylc-ui to test it) but maybe it's a good idea, so that we can test the subscriptions without the cylc-ui as well?

Maybe we could migrate it to Apollo? (or maybe there's one out there already?)

dwsutherland:
> I think we are done here for now. And now the only missing parts are in the cylc-flow repository. What do you think @dwsutherland ?

Yes, although perhaps we could leave this one open until we have put a minimal Subscription into `cylc-flow`?

cylc/uiserver/main.py
kinow: Yeah, for now keeping it to minimum changes required, so simply copied the template from the PR for now. Leaving the change of `/graphql/graphiql` to `/graphiql` for later too :+1:

kinow:
> > I think we are done here for now. And now the only missing parts are in the cylc-flow repository. What do you think @dwsutherland ?
> 
> Yes, although perhaps we could leave this one open until we have put a minimal Subscription into `cylc-flow`?

Yup, I'm planning to mark it ready for review once we get at least one subscription working, so that we can test it from point (ui + hub) to point (cylc-flow with running workflow data).

.travis.yml
```diff
@@ -24,7 +24,7 @@ python: 3.7
 
 before_install:
   - pip install pycodestyle
-  - pycodestyle --exclude=".git,.tox,__pycache,venv,.eggs,build,jupyterhub_config.py"
+  - pycodestyle --exclude=".git,.tox,__pycache,venv,.eggs,build,jupyterhub_config.py,cylc/uiserver/websockets/"
```
kinow: This is due to the `graphene-tornado` class that we copied from an open PR. Mainly due to line number that is greater than 80 there :+1: so we need to ignore that for now.

cylc/uiserver/handlers.py
```diff
@@ -145,9 +150,58 @@ def prepare(self):
         super().prepare()
 
 
+class SubscriptionHandler(websocket.WebSocketHandler):
```
kinow: Note: copied from the PR's example.

cylc/uiserver/handlers.py
```diff
@@ -145,9 +150,58 @@ def prepare(self):
...
+    def check_origin(self, origin: str) -> bool:
+        return True
+
+    @property
+    def context(self):
+        wider_context = {
+            'request': self.request,
+            'resolvers': self.resolvers,
+        }
+        return wider_context
```
kinow: This was necessary but took me a while to understand why. Thanks @dwsutherland for saving me a lot of time! Note for others: the schema and base resolvers are defined in `cylc-flow`. But in `cylc-uiserver` we **use** the schema, but **extend** the resolvers in cylc-uiserver (I believe it's to augment the resolvers with extra data/functionality). So we need to pass the resolvers from cylc-uiserver to cylc-flow... so that when the schema resolves it it uses the right resolvers :grimacing: (feel free to correct me if I said something wrong @dwsutherland)

cylc/uiserver/handlers.py
```diff
@@ -145,9 +150,58 @@ def prepare(self):
...
+
+
+class GraphiQLHandler(UIServerGraphQLHandler):
```
kinow: From the PR's example. This handler renders a custom React app that talks WebSockets. So that the query is reloaded automatically :+1:

cylc/uiserver/websockets/template.py
```diff
@@ -0,0 +1,130 @@
+# This file is a temporary solution for subscriptions with graphql_ws and
+# Tornado, from the following pending PR to graphql-ws:
+# https://github.com/graphql-python/graphql-ws/pull/25/files
+# The file was copied from this revision:
+# https://github.com/graphql-python/graphql-ws/blob/cf560b9a5d18d4a3908dc2cfe2199766cc988fef/examples/tornado/template.py
+
+from string import Template
+
+
+def render_graphiql(base_url: str) -> str:
```
kinow: From the PR's example.

cylc/uiserver/websockets/tornado.py
```diff
@@ -0,0 +1,124 @@
...
+from graphql_ws.constants import (
+    GQL_CONNECTION_ACK,
+    GQL_CONNECTION_ERROR,
+    GQL_COMPLETE
+)
+
+setup_observable_extension()
+
+
+class TornadoConnectionContext(BaseConnectionContext):
```
kinow: From the PR's example, with one modification.

cylc/uiserver/websockets/tornado.py
```diff
@@ -0,0 +1,124 @@
...
+        try:
+            await self.on_connect(connection_context, payload)
+            await self.send_message(connection_context, op_type=GQL_CONNECTION_ACK)
+        except Exception as e:
+            await self.send_error(connection_context, op_id, e, GQL_CONNECTION_ERROR)
+            await connection_context.close(1011)
+
+    def execute(self, request_context, params):
+        params['context_value'] = request_context
+        return super().execute(request_context, params)
```
kinow: @dwsutherland here is where I got stuck yesterday (good thing I didn't create the PR yesterday).

Reading the existing tickets/PR's this morning, I found this one: https://github.com/graphql-python/graphql-ws/pull/19 which is pretty much the same issue I was having. I created the `context` in the Tornado handler, but it would not be accessible to the `cylc-flow` schema. Turns out it was not being set, so first I had to add the `context` in the Websockets handler, and then override the `execute` method in this class, as per this comment: https://github.com/graphql-python/graphql-ws/issues/7

cylc/uiserver/handlers.py
```diff
@@ -145,9 +150,58 @@ def prepare(self):
...
+    def initialize(self, sub_server, resolvers):
+        self.subscription_server = sub_server
+        self.resolvers = resolvers
+
+    def select_subprotocol(self, subprotocols):
+        return GRAPHQL_WS
+
+    def open(self, *args, **kwargs):
+        IOLoop.current().spawn_callback(self.subscription_server.handle, self,
+                                        self.context)
```
kinow: @dwsutherland when the Tornado handler is created, it adds a callback to the existing loop. The example from the PR passes just `(self.subscription_server.handle, self)`, but I had to include the `self.context` as well. Then in another part of the code had to extend `execute()` method to retrieve the value passed here and propagate it to the cylc-flow schema :+1:

cylc/uiserver/handlers.py
dwsutherland: > I believe it's to augment the resolvers with extra data/functionality

Yes, the mutations resolvers are different (at the moment) but query resolver's identical.. `BaseResolvers` is extended in both `cylc-flow` and `cylc-uiserver` .. just in different ways..

> So we need to pass the resolvers from cylc-uiserver to cylc-flow... so that when the schema resolves it it uses the right resolvers 😬 (feel free to correct me if I said something wrong @dwsutherland)

Not sure I follow here.. WRT code, nothings passed from `cylc-uiserver` to `cylc-flow` (only the other way round.. The GraphQL end points at UIS and WS are completely independent of each other, except for Mutations where the UIS passes the request doc directly to the respective WS GraphQLs..
The schema resolves from the same named methods at both UIS & WS, this is to keep the schema identical for both end-points.. 
The main purposes are free upgrades of `cylc-uiserver` when `cylc-flow` is upgraded, and less code to maintain. (keeps the two in sync).. `BaseResolvers` allows us to do the same with what resolvers we are capable of.

kinow: >Not sure I follow here.. WRT code, nothings passed from cylc-uiserver to cylc-flow (only the other way round)

That's what I am not understanding. Probably I am missing something here. But bear with me a bit more. Let me use some code from `master` to try to explain again what I was trying to say.

- This is the Tornado handler for `/graphql`: https://github.com/cylc/cylc-uiserver/blob/453688c023e39a030c987fd9b4fb172852140fe5/cylc/uiserver/main.py#L109-L111
- It passes a `resolvers=` parameter to the Handler.
- That `resolvers` will be set in the Handler `initialize` method: https://github.com/cylc/cylc-uiserver/blob/453688c023e39a030c987fd9b4fb172852140fe5/cylc/uiserver/handlers.py#L132
- Later, the request will have a `.context` property defined here (still in `cylc-uiserver`): https://github.com/cylc/cylc-uiserver/blob/453688c023e39a030c987fd9b4fb172852140fe5/cylc/uiserver/handlers.py#L135-L141

From what I understood about the design in place, `cylc-flow` has the schema that is used by Cylc UI Server. So Cylc UI Server is independent of Cylc Flow, and just a consumer of its schema. I like this approach.

But now here's the part I found strange. In Cylc Flow, the `schema.py` receives an `info` object, and uses the `context` to retrieve the resolvers: https://github.com/cylc/cylc-flow/blob/9ff5b7719fe924a9a2c8c1b3446faaaf47a4c5b2/cylc/flow/network/schema.py#L275

That resolver is not `cylc.flow.network.resolvers`, but `cylc.uiserver.resolvers`. **So Cylc UI Server has a dependency on the schema of Cylc Flow; and Cylc Flow has a dependency on the resolvers of Cylc UI Server**.

From what I understood, if I made a change in the resolvers of Cylc UI Server that impacted the schema in Cylc Flow, we could end up having to patch/release Cylc Flow. Which is going in the wrong dependency direction? (i.e. I thought Cylc UI Server would be a consumer of Cylc Flow, not the other way around)

But again, I could be missing something here, and saying something wrong - so sorry in case it's just gibberish and also for the long reply :+1:

dwsutherland: > That resolver is not cylc.flow.network.resolvers, but cylc.uiserver.resolvers. So Cylc UI Server has a dependency on the schema of Cylc Flow; and Cylc Flow has a dependency on the resolvers of Cylc UI Server.

Here's the missing pieces:
https://github.com/cylc/cylc-flow/blob/9ff5b7719fe924a9a2c8c1b3446faaaf47a4c5b2/cylc/flow/network/server.py#L36
Instantiated here:
https://github.com/cylc/cylc-flow/blob/9ff5b7719fe924a9a2c8c1b3446faaaf47a4c5b2/cylc/flow/network/server.py#L268-L270
And passed in here:
https://github.com/cylc/cylc-flow/blob/9ff5b7719fe924a9a2c8c1b3446faaaf47a4c5b2/cylc/flow/network/server.py#L342-L350

So nothing from `cylc-uiserver` is imported or passed to `cylc-flow`

kinow: AaaaaaaaaahhhhhhhH! Now that makes sense! hahaha. Thanks @dwsutherland !

cylc/uiserver/websockets/tornado.py
```diff
@@ -0,0 +1,210 @@
...
+
+    async def send_execution_result(self, connection_context, op_id,
```
kinow: @dwsutherland the solution you linked was added to the cylc-uiserver code from Tornado PR, not to cylc-flow :+1:

dwsutherland: Nice 👍  ... Should really be part of the code base, as it overwrites:
https://github.com/graphql-python/graphql-ws/blob/523dd1516d8709149a475602042d6a3c5eab5a3d/graphql_ws/base.py#L113-L115

Although it would have to handle more than just asyncio I suppose (probably why)

kinow: But it does give users a nice feature of being able to re-use the Query schema (even if partially). Otherwise we would have to simply duplicate all those fields that have resolvers :open_mouth:

kinow:
Added some unit tests. Should be ready for review now.

kinow:
Juts added a new commit with the changelog.

kinow:
Oh, got one conflict in the handlers, will fix it in a few minutes :+1:

cylc/uiserver/websockets/template.py
```diff
@@ -0,0 +1,130 @@
...
+      }),
+      document.body
+    );
+  </script>
+</body>
+</html>''').substitute(
+        GRAPHIQL_VERSION='0.10.2',
+        SUBSCRIPTIONS_TRANSPORT_VERSION='0.7.0',
+        subscriptionsEndpoint=f'ws://localhost:8000/{base_url}subscriptions',
+        # subscriptionsEndpoint='ws://localhost:8000/',
```
kinow: This doesn't work when you deploy JupyterHub to a different port! :angry: 

Realized that I couldn't use GraphiQL when deployed it behind a Nginx rev proxy (following JupyterHub documentation). We can fix it later maybe... though I am not sure how this could be fixed in this template...

dwsutherland: Maybe turn it into a Jinja2 (or other) template? then the handler can render?

dwsutherland: Ah.. It's a python script.. Maybe directly via handler?

kinow: We had the same line of thinking! I started by thinking that I could template'ize.. then thought about doing it in Python... then JS... and then I remember that I actually tried it, but I think it failed because - for some reason - the relative path (i.e.  or `/subscriptions`) or the abbreviated form `ws://:8000/subscriptions` did not work in my browser.

And I couldn't quickly find a way to inject the current server name. Gotta have another go with more calm/time over the next week :+1: thanks!

kinow:
I think I found one possible solution. A mix of more JS, and just a little Python. We still replace the `base_url` (required as we need to know what value goes in `/user/$USER/`), but the protocol+host+port can be done in JS.

The WebSocket client is created on the client-side, so it should be safe to ignore JupyterHub configurations like `c.JupyterHub.port` and `...host`, and just grab whatever is specified in the browser.

Test with default port 8000 after running `jupyterhub`.

image

Then the result after running `jupyterhub --ip=10.0.2.15 --port=4444`:

image

cylc/uiserver/websockets/template.py
kinow: Modified to create the WebSocket subscription URL using JS for hostname and port, and Python for the base URL.

kinow:
Rebased.

PR #94, opened 2019-10-14
Remove old CylcScan tornado handler

kinow:

This was our first handler before GraphQL, implemented with the help of @dwsutherland . It worked at the time, but now we have GraphQL and we are not using it. So probably safer to remove (we can always find it again using `git` if necessary).

**Requirements check-list**
- [x] Does not need tests (why? code removal).


kinow:
One review should do IMO.

PR #95, opened 2019-10-18
WFS Data Store Sync

dwsutherland:
These changes partially address https://github.com/cylc/cylc-flow/issues/3207

This branch is the companion of and needs to be merged **after** https://github.com/cylc/cylc-flow/pull/3389
(CI will fail until the [PR](https://github.com/cylc/cylc-flow/pull/3389) is merged)

**Functional/Done:**
- [x] Updates to the data-store via subscriptions to workflow published deltas.
- [x] Verification of  data-store, and reconciliation with the workflow data-store.

**To Do:** (excluding below)
- [ ] Investigate alternative to threading the subscription/sync loops?

The sync appears solid; verification seldom fails (I haven't seen one yet), and there's reconciliation on failure..
- The data-store is split-up-into/published-as "topics" (i.e. `task_proxies`, `jobs`, `workflow`...etc), and these topic messages/deltas/updates are subscribed-to/received, processed, and verified independently. This means update frequency can differ between topics, while order and content is orchestrated by the workflow service (WFS).
- The subscription is started first (in a separate thread), and messages wait in the zmq Queue for initial data request (for ~3sec, then moves onto the next and waits again).
- The entire workflow is requested in the main thread.
- The subscription message/delta handler then works through the queue (via recv_multi ZeroMQ holds messages up to the High Water Mark (HWM)), only applying deltas that were created **after** the initial data and then **after** last applied delta.
- The local data-store topic elements (i.e. `task_proxies`) are verified against the respective WFS, via a checksum sent with the delta, and a new set is requested on failure.

(deltas only contain the elements and fields that were updated)

We need to figure out if creating threads for each subscription loop (i.e. one per workflow), is scalable..  As in, do they yield execution to other threads while `await`ing a publish..  I did it to avoid blocking the main asyncio loop with subscription loops.. Sockets can't be passed between threads, but I think it's fine for them to exist in their lifetime on one thread. (will see what happens with a large load of workflows)
    Perhaps if we can dynamically update (`add/remove`) an `asyncio.gather` (or something), then we could avoid threads?...

But for now it's fully functional! 🎉 

**Requirements check-list**
- [x] Contains logically grouped changes.

### DIFF
--- a/cylc/uiserver/data_mgr.py
+++ b/cylc/uiserver/data_mgr.py
@@ -19,39 +19,133 @@
 import asyncio
 import logging
 
-from cylc.flow.exceptions import ClientError, ClientTimeout
+from concurrent.futures import ThreadPoolExecutor
+from functools import partial
+from time import sleep
+
 from cylc.flow.network.server import PB_METHOD_MAP
 from cylc.flow.network.scan import MSG_TIMEOUT
+from cylc.flow.network.subscriber import WorkflowSubscriber, process_delta_msg
 from cylc.flow.ws_data_mgr import (
-    EDGES, FAMILIES, FAMILY_PROXIES, JOBS, TASKS, TASK_PROXIES, WORKFLOW
+    EDGES, FAMILIES, FAMILY_PROXIES, JOBS, TASKS, TASK_PROXIES, WORKFLOW,
+    DELTAS_MAP, apply_delta, generate_checksum
 )
+from .workflows_mgr import workflow_request
 
 logger = logging.getLogger(__name__)
-
-
-async def get_workflow_data(w_id, client, method):
-    """Call WS endpoint for entire workflow protobuf message."""
-    # Use already established client
-    try:
-        pb_msg = await client.async_request(method)
-    except ClientTimeout as exc:
-        logger.exception(exc)
-        return (w_id, MSG_TIMEOUT)
-    except ClientError as exc:
-        logger.exception(exc)
-        return (w_id, None)
-    else:
-        ws_data = PB_METHOD_MAP[method]()
-        ws_data.ParseFromString(pb_msg)
-    return (w_id, ws_data)
 
 
 class DataManager:
     """Manage the local data-store acquisition/updates from all workflows."""
 
-    def __init__(self, ws_mgr):
-        self.ws_mgr = ws_mgr
+    def __init__(self, workfloworkflows_mgr):
+        self.workflows_mgr = workfloworkflows_mgr
         self.data = {}
+        self.w_subs = {}
+        self.topics = {topic.encode('utf-8') for topic in DELTAS_MAP}
+        self.loop = None
+
+    async def sync_workflows(self):
+        """Run data store sync with workflow services.
+
+        Subscriptions and sync management is instantiated and run in
+        a separate thread for each workflow. This is to avoid the sync loop
+        blocking the main loop.
+        """
+        self.loop = asyncio.get_running_loop()
+        # Might be options other than threads to achieve
+        # non-blocking subscriptions, but this works.
+        with ThreadPoolExecutor() as pool:
+            while True:
+                for w_id, info in self.workflows_mgr.workflows.items():
+                    if w_id not in self.w_subs:
+                        pool.submit(
+                            partial(
+                                self._start_subscription,
+                                w_id,
+                                info['host'],
+                                info['pub_port']))
+                        await self.entire_workflow_update(ids=[w_id])
+                for w_id in list(self.w_subs):
+                    if w_id not in self.workflows_mgr.workflows:
+                        self.w_subs[w_id].stop()
+                        del self.w_subs[w_id]
+                        if w_id in self.data:
+                            del self.data[w_id]
+
+                await asyncio.sleep(1.0)
+
+    def _start_subscription(self, w_id, host, port):
+        """Instatiate and run subscriber and data-store management."""
+        self.w_subs[w_id] = WorkflowSubscriber(
+            host, port,
+            context=self.workflows_mgr.context, topics=self.topics)
+        self.w_subs[w_id].loop.run_until_complete(
+            self.w_subs[w_id].subscribe(
+                process_delta_msg,
+                func=self.update_workflow_data,
+                w_id=w_id))
+
+    def update_workflow_data(self, topic, delta, w_id):
+        """Manage and apply incomming data-store deltas."""
+        loop_cnt = 0
+        while loop_cnt < 5:
+            if w_id not in self.data:
+                sleep(0.5)
+                loop_cnt += 1
+                continue
+            break
+        if w_id not in self.data:
+            return
+        delta_time = getattr(
+                delta, 'time', getattr(delta, 'last_updated', 0.0))
+        # If the workflow has reloaded recreate the data
+        if delta.reloaded:
+            self.data[w_id][topic] = {ele.id: ele for ele in delta.deltas}
+            self.data[w_id]['delta_times'][topic] = delta_time
+        elif delta_time >= self.data[w_id]['delta_times'][topic]:
+            apply_delta(topic, delta, self.data[w_id])
+            self.data[w_id]['delta_times'][topic] = delta_time
+            self.reconcile_update(topic, delta, w_id)
+
+    def reconcile_update(self, topic, delta, w_id):
+        """Reconcile local with workflow data-store.
+
+        Verify data-store is in sync by topic/element-type
+        and on failure request entire set of respective data elements.
+
+        """
+        if topic == WORKFLOW:
+            return
+        if topic == EDGES:
+            s_att = 'id'
+        else:
+            s_att = 'stamp'
+        local_checksum = generate_checksum(
+            [getattr(e, s_att)
+             for e in self.data[w_id][topic].values()])
+        if local_checksum != delta.checksum:
+            future = asyncio.run_coroutine_threadsafe(
+                workflow_request(
+                    self.workflows_mgr.workflows[w_id]['req_client'],
+                    'pb_data_elements',
+                    args={'element_type': topic}
+                ),
+                self.loop
+            )
+            try:
+                _, new_delta_msg = future.result(5.0)
+            except asyncio.TimeoutError:
+                logger.info(f'The reconcile update coroutine {w_id} {topic}'
+                            f'took too long, cancelling the task...')
+                future.cancel()
+            except Exception as exc:
+                logger.exception(exc)
+            else:
+                new_delta = DELTAS_MAP[topic]()
+                new_delta.ParseFromString(new_delta_msg)
+                apply_delta(topic, new_delta, self.data[w_id])
+                self.data[w_id]['delta_times'][topic] = new_delta.time
 
     # Data syncing
     async def entire_workflow_update(self, ids=None):
@@ -61,33 +155,41 @@ class DataManager:
 
         # Prune old data
         for w_id in list(self.data):
-            if w_id not in self.ws_mgr.workflows:
+            if w_id not in self.workflows_mgr.workflows:
                 del self.data[w_id]
 
-        # Fetch new data
-        ws_args = (
-            (w_id, info['req_client'], 'pb_entire_workflow')
-            for w_id, info in self.ws_mgr.workflows.items())
+        # Request new data
+        req_method = 'pb_entire_workflow'
+        req_kwargs = (
+            {'client': info['req_client'],
+             'command': req_method,
+             'req_context': w_id}
+            for w_id, info in self.workflows_mgr.workflows.items())
         gathers = ()
-        for args in ws_args:
-            if not ids or args[0] in ids:
-                gathers += (get_workflow_data(*args),)
+        for kwargs in req_kwargs:
+            if not ids or kwargs['req_context'] in ids:
+                gathers += (workflow_request(**kwargs),)
         items = await asyncio.gather(*gathers)
         new_data = {}
         for w_id, result in items:
             if result is not None and result != MSG_TIMEOUT:
+                pb_data = PB_METHOD_MAP[req_method]()
+                pb_data.ParseFromString(result)
                 new_data[w_id] = {
-                    EDGES: {e.id: e for e in getattr(result, EDGES)},
-                    FAMILIES: {f.id: f for f in getattr(result, FAMILIES)},
+                    EDGES: {e.id: e for e in getattr(pb_data, EDGES)},
+                    FAMILIES: {f.id: f for f in getattr(pb_data, FAMILIES)},
                     FAMILY_PROXIES: {
                         n.id: n
-                        for n in getattr(result, FAMILY_PROXIES)},
-                    JOBS: {j.id: j for j in getattr(result, JOBS)},
-                    TASKS: {t.id: t for t in getattr(result, TASKS)},
+                        for n in getattr(pb_data, FAMILY_PROXIES)},
+                    JOBS: {j.id: j for j in getattr(pb_data, JOBS)},
+                    TASKS: {t.id: t for t in getattr(pb_data, TASKS)},
                     TASK_PROXIES: {
                         n.id: n
-                        for n in getattr(result, TASK_PROXIES)},
-                    WORKFLOW: getattr(result, WORKFLOW),
+                        for n in getattr(pb_data, TASK_PROXIES)},
+                    WORKFLOW: getattr(pb_data, WORKFLOW),
+                    'delta_times': {
+                        topic: getattr(pb_data, WORKFLOW).last_updated
+                        for topic in DELTAS_MAP.keys()}
                 }
 
         self.data.update(new_data)
--- a/cylc/uiserver/data_mgr.py
+++ b/cylc/uiserver/data_mgr.py
@@ -13,8 +13,24 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Manage a local data-store replica of all workflow service data-stores.
 
-"""Create and update the data structure for all workflow services."""
+A local data-store is created and synced for all workflows established by
+the workflow service manager.
+
+The workflows publish the updated fields of the updated data elements (deltas),
+and these elements are grouped by type/topic. Once subscribed to, the publisher
+queues these messages until the are received, if the delta creation time is
+newer than that of the last update then it is applied (updates merged, pruned
+deleted) then a checksum is generated from the time stamped IDs and compared to
+the published one.
+
+Reconciliation on failed verification is done by requesting all elements of a
+topic, and replacing the respective data-store elements with this.
+
+Subscriptions are currently run in a different thread (via ThreadPoolExecutor).
+
+"""
 
 import asyncio
 import logging
@@ -36,10 +52,10 @@ logger = logging.getLogger(__name__)
 
 
 class DataManager:
-    """Manage the local data-store acquisition/updates from all workflows."""
+    """Manage the local data-store acquisition/updates for all workflows."""
 
-    def __init__(self, workfloworkflows_mgr):
-        self.workflows_mgr = workfloworkflows_mgr
+    def __init__(self, workflows_mgr):
+        self.workflows_mgr = workflows_mgr
         self.data = {}
         self.w_subs = {}
         self.topics = {topic.encode('utf-8') for topic in DELTAS_MAP}
@@ -76,7 +92,14 @@ class DataManager:
                 await asyncio.sleep(1.0)
 
     def _start_subscription(self, w_id, host, port):
-        """Instatiate and run subscriber and data-store management."""
+        """Instatiate and run subscriber and data-store management.
+
+        Args:
+            w_id (str): Workflow external ID.
+            host (str): Hostname of target workflow.
+            port (int): Port of target workflow.
+
+        """
         self.w_subs[w_id] = WorkflowSubscriber(
             host, port,
             context=self.workflows_mgr.context, topics=self.topics)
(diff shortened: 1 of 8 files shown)


cylc/uiserver/data_mgr.py
```diff
@@ -19,39 +19,133 @@
...
+        """Run data store sync with workflow services.
+
+        Subscriptions and sync management is instantiated and run in
+        a separate thread for each workflow. This is to avoid the sync loop
+        blocking the main loop.
+        """
+        self.loop = asyncio.get_running_loop()
+        # Might be options other than threads to achieve
+        # non-blocking subscriptions, but this works.
+        with ThreadPoolExecutor() as pool:
```
kinow: Ah! I hadn't seen this thread pool @dwsutherland , so my comment about the zmq context is probably invalid, sorry

cylc/uiserver/data_mgr.py
```diff
@@ -19,39 +19,133 @@
...
+    async def sync_workflows(self):
+        """Run data store sync with workflow services.
+
+        Subscriptions and sync management is instantiated and run in
+        a separate thread for each workflow. This is to avoid the sync loop
+        blocking the main loop.
+        """
+        self.loop = asyncio.get_running_loop()
+        # Might be options other than threads to achieve
+        # non-blocking subscriptions, but this works.
```
kinow: Looks really strange, with a ThreadPool within an async function. It wouldn't work if we used the same ioloop instance?

dwsutherland:
> @dwsutherland so far all good, I just need a bit more of time with `data_mgr.py` to understand how it is going to work (might need it for subscriptions I guess). So will finish review either tonight or tomorrow 👍

There's not a lot of comments or docstrings in this one yet.. Hopefully I'll push that tonight.

cylc/uiserver/data_mgr.py
dwsutherland: How do you run a loop and carry on in parent function otherwise?

dwsutherland: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor

kinow: Oh yeah. Interesting. Ioloop's run_in_executor appears to use a threadpool too

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor

dwsutherland: Yeah, and not only is waiting for a publish blocking, but it’s in an infinite loop so when it receives and processes one publish it waits for the next publish... For each workflow...

The other thing; I thought about using asyncio.gather(...), however I don’t think you can dynamically add and remove subscriptions to the coros iterable argument (and you’d still need a different executor)

kinow: Well, it works as-is, and asyncio uses threading when necessary too. So I think the TODO marker is more than enough for now :-) Let's mark this as resolved :tada:

kinow: (plus, I use `jupyterhub` + UI Server every day, for several hours, if there's any performance bottleneck, bugs, etc, they will eventually appear and be easy to investigate - our code base is really quite small in ui server for now)

cylc/uiserver/data_mgr.py
```diff
@@ -19,39 +19,133 @@
...
-        ws_data.ParseFromString(pb_msg)
-    return (w_id, ws_data)
-
-
 class DataManager:
     """Manage the local data-store acquisition/updates from all workflows."""
 
-    def __init__(self, ws_mgr):
-        self.ws_mgr = ws_mgr
+    def __init__(self, workfloworkflows_mgr):
```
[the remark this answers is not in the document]
dwsutherland: my bad! find and replace strikes back

cylc/uiserver/data_mgr.py
dwsutherland: I ran this with 32 workflows on my VM running on two cores of my Surface Pro lol! (the workflows slowed it down)

dwsutherland: I ran this with 32 workflows on my VM running on two cores of my Surface Pro lol! (the workflows slowed it down)

kinow:
Thanks for the code comments you just added @dwsutherland ! Had a quick read, but will read it with :coffee: tomorrow.

dwsutherland:
> Thanks for the code comments you just added @dwsutherland ! Had a quick read, but will read it with ☕️ tomorrow.

I have no idea how to unit test the data sync... would have to recreate a workflow, or create a lot of fake data (which would be annoying to update) and do a publish of this...

kinow:
> > Thanks for the code comments you just added @dwsutherland ! Had a quick read, but will read it with coffee tomorrow.
> 
> I have no idea how to unit test the data sync... would have to recreate a workflow, or create a lot of fake data (which would be annoying to update) and do a publish of this...

This one might be easier to leave with no test for now, or wait until we add either some sort of functional tests here or a cross project functional test (I think we have an issue somewhere).

kinow:
ps: I think Travis will keep failing until cylc-flow's PR is merged: `E   ImportError: cannot import name 'ZMQSocketBase' from 'cylc.flow.network' (/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/cylc/flow/network/__init__.py)`

kinow:
Oh, actually had something new in the console with this branch:

```bash
zmq.error.ZMQError: Too many open files
2019-11-13 12:32:47,467 tornado.application ERROR    Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7f490bece080>>, <Task finished coro=<WorkflowsManager.gather_workflows() done, defined at /home/kinow/Development/python/workspace/cylc-uiserver/cylc/uiserver/workflows_mgr.py:95> exception=ZMQError('Too many open files')>)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 105, in gather_workflows
    items = await asyncio.gather(*gathers)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 77, in est_workflow
    context=context, timeout=timeout)
  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/network/client.py", line 129, in __init__
    self.start(host, port)
  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/network/__init__.py", line 91, in start
    self._start_sequence(*args, **kwargs)
  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/network/__init__.py", line 105, in _start_sequence
    self._socket_connect(*args, **kwargs)
  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/network/__init__.py", line 137, in _socket_connect
    self.socket = self.context.socket(self.pattern)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/zmq/sugar/context.py", line 146, in socket
    s = self._socket_class(self, socket_type, **kwargs)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/zmq/_future.py", line 134, in __init__
    super(_AsyncSocket, self).__init__(context, socket_type, **kwargs)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/zmq/sugar/socket.py", line 59, in __init__
    super(Socket, self).__init__(*a, **kw)
  File "zmq/backend/cython/socket.pyx", line 328, in zmq.backend.cython.socket.Socket.__init__
zmq.error.ZMQError: Too many open files
```

I do have a few too many things running in parallel, but not different than what I normally have everyday. I don't remember seeing a "Too many open files" problem before. Will take a look at my OS limits.

kinow:
```bash
$ ulimit
unlimited
$ ulimit -Hn
1048576
```

kinow:
Hmmm, restarted the UI Server (while the two workflows are still running), then went to run more benchmarking tests on Cylc UI, then noticed this later in the console:

```
[I 2019-11-13 12:55:05.620 JupyterHub log:174] 200 GET /hub/api/authorizations/token/[secret] (kinow@127.0.0.1) 34.16ms
2019-11-13 12:55:11,842 tornado.application ERROR    Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7fd249f486a0>>, <Task finished coro=<WorkflowsManager.gather_workflows() done, defined at /home/kinow/Development/python/workspace/cylc-uiserver/cylc/uiserver/workflows_mgr.py:95> exception=TypeError("stop() got an unexpected keyword argument 'stop_loop'")>)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 126, in gather_workflows
    client.stop(stop_loop=False)
TypeError: stop() got an unexpected keyword argument 'stop_loop'
``

**Comment by kinow** [comment]
Yup, all good now. No more errors so far :grimacing: sorry for the noise!

**Comment by dwsutherland** [comment]
> Yup, all good now. No more errors so far 😬 sorry for the noise!

All good. Cylc does raise the open file count, I remember having to address this with Cylc7 in NIWA operations..

Thanks 👍

**Comment by kinow** [comment]
> > Yup, all good now. No more errors so far grimacing sorry for the noise!
> 
> All good. Cylc does raise the open file count, I remember having to address this with Cylc7 in NIWA operations..
> 
> Thanks +1

Makes sense. The only other thing I noticed is that stopping the hub takes a bit longer now. I assume it's because the UI Server is waiting for the thread or tasks to finish. But that's expected too I believe.

**Comment by dwsutherland** [comment]
> > > Yup, all good now. No more errors so far grimacing sorry for the noise!
> > 
> > 
> > All good. Cylc does raise the open file count, I remember having to address this with Cylc7 in NIWA operations..
> > Thanks +1
> 
> Makes sense. The only other thing I noticed is that stopping the hub takes a bit longer now. I assume it's because the UI Server is waiting for the thread or tasks to finish. But that's expected too I believe.

Yes, correct. I think it's the thread join, but the amount of time seems to be consistent (might be a way to daemonise the threads so we can exit more harshly and quicker).

**Comment by dwsutherland** [comment]
The UI Server now uses the published `shutdown` topic to stop the subscriber and signal the workflow manager to clear the workflow.

File: cylc/uiserver/data_mgr.py [hunk 4]
```diff
@@ -13,81 +13,231 @@
...
+
+        Args:
+            topic (str): topic of published data.
+            delta (object): Published protobuf message data container.
+            w_id (str): Workflow external ID.
+
+        """
+        if topic == WORKFLOW:
+            return
+        if topic == EDGES:
```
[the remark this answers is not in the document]
dwsutherland: Because edges don't have dynamic fields (no plan to change that (yet!)). Once created they don't change, a suite reload replaces existing, and they only get deleted.  (guess for consistency we could still do it the same way...)

cylc/uiserver/data_mgr.py
```diff
@@ -13,81 +13,231 @@
...
+                                info['pub_port']))
+                        await self.entire_workflow_update(ids=[w_id])
+                for w_id in list(self.w_subs):
+                    if w_id not in self.workflows_mgr.workflows:
+                        self.w_subs[w_id].stop()
+                        del self.w_subs[w_id]
+                        if w_id in self.data:
+                            del self.data[w_id]
+
+                await asyncio.sleep(1.0)
```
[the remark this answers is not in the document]
dwsutherland: Yes, will make this event driven (along with the clean-up). This is only when it starts syncing, I still have a loop delay on discovery in the workflow manager.

cylc/uiserver/data_mgr.py
```diff
@@ -13,81 +13,231 @@
...
+
+        Args:
+            topic (str): topic of published data.
+            delta (object): Published protobuf message data container.
+            w_id (str): Workflow external ID.
+
+        """
+        # wait until data-store is populated for this workflow
+        loop_cnt = 0
+        while loop_cnt < 5:
```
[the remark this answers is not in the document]
dwsutherland: Maybe (?).. Subscription happens first, but needs to wait around for initial data payload from REQ client... Publishes are queued until the subscriber recv them, earlier than payload deltas are ignored.

cylc/uiserver/data_mgr.py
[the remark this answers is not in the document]
dwsutherland: Done.

cylc/uiserver/data_mgr.py
```diff
@@ -13,81 +13,231 @@
...
+        # If the workflow has reloaded recreate the data
+        # otherwise apply the delta if it's newer than the previously applied.
+        if delta.reloaded:
+            self.data[w_id][topic] = {ele.id: ele for ele in delta.deltas}
+            self.data[w_id]['delta_times'][topic] = delta_time
+        elif delta_time >= self.data[w_id]['delta_times'][topic]:
+            apply_delta(topic, delta, self.data[w_id])
+            self.data[w_id]['delta_times'][topic] = delta_time
+            self.reconcile_update(topic, delta, w_id)
+
```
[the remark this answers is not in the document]
dwsutherland: Unless I missed your point:  The `elif delta_time >= self.data[w_id]['delta_times'][topic]:` statement filters for valid deltas to apply, so you don't really want to do anything to the store with an invalid  delta (delta creation time < latest applied delta creation time).

cylc/uiserver/data_mgr.py
```diff
@@ -13,81 +13,231 @@
...
+                    args={'element_type': topic}
+                ),
+                self.loop
+            )
+            try:
+                _, new_delta_msg = future.result(5.0)
+            except asyncio.TimeoutError:
+                logger.info(f'The reconcile update coroutine {w_id} {topic}'
+                            f'took too long, cancelling the task...')
+                future.cancel()
```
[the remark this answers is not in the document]
dwsutherland: Done, the sync is stopped.. And the workflow purged from the data-store.. The workflow manager can add it again if it becomes responsive again.

cylc/uiserver/data_mgr.py
```diff
@@ -13,81 +13,231 @@
...
+            future = asyncio.run_coroutine_threadsafe(
+                workflow_request(
+                    self.workflows_mgr.workflows[w_id]['req_client'],
+                    'pb_data_elements',
+                    args={'element_type': topic}
+                ),
+                self.loop
+            )
+            try:
+                _, new_delta_msg = future.result(5.0)
```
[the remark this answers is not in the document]
dwsutherland: Done.

cylc/uiserver/data_mgr.py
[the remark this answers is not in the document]
oliver-sanders: I was wondering about the impact of out of order messages and whether there is the potential for the data store to become corrupted but for us not to notice until the next message comes in.

I think you've got this covered with the topic/checksum system so probably irrelevant:

* If the workflow sends messages: 1, 2, 3
* And the UI Server receives messages: 1, 3, 2
* Then we know something is out-of-wack, we don't need to wait for message 4 to find that out.

I think this system should pick up on the error, if messages 1, 2 and 3 relate to different topics is the the potential for anything to fall down the cracks?

cylc/uiserver/data_mgr.py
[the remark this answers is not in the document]
oliver-sanders: Cool

cylc/uiserver/data_mgr.py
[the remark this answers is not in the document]
oliver-sanders: Thanks.

cylc/uiserver/workflows_mgr.py
```diff
@@ -42,69 +43,84 @@
...
+        """Scan, establish, and discard workflows."""
         scanflows = {}
         cre_owner, cre_name = re_compile_filters(None, ['.*'])
         scan_args = (
-            (reg, host, port, CLIENT_TIMEOUT)
-            for reg, host, port in
-            get_scan_items_from_fs(cre_owner, cre_name))
+            (reg, host, port, pub_port, self.context, CLIENT_TIMEOUT)
+            for reg, host, port, pub_port in
+            get_scan_items_from_fs(cre_owner, cre_name)
```
oliver-sanders: Suites launched pre https://github.com/cylc/cylc-flow/pull/3389 will generate quite a lot of traceback somewhere around here. Might be worth suppressing it, the contact file has an `API` field which we can use to filter out incompatible suites.

```
2019-11-28 11:34:22,270 tornado.application ERROR    Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7fb5a252b400>>, <Task finished coro=<WorkflowsManager.gather_workflows() done, defined at ~/cylc-uiserver/cylc/uiserver/workflows_mgr.py:99> exception=KeyError('CYLC_SUITE_PUBLISH_PORT')>)
  File "/home/h06/osanders/miniconda3/envs/cylc8/lib/python3.7/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/home/h06/osanders/miniconda3/envs/cylc8/lib/python3.7/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "~/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 112, in gather_workflows
    for arg in scan_args:
  File "~/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 104, in <genexpr>
    (reg, host, port, pub_port, self.context, CLIENT_TIMEOUT)
  File "~/cylc/cylc/flow/network/scan.py", line 280, in get_scan_items_from_fs
    contact_data[ContactFileFields.PUBLISH_PORT]
KeyError: 'CYLC_SUITE_PUBLISH_PORT'
```

dwsutherland: True.. I might have to increment the `cylc-flow` API version in the publish PR.

cylc/uiserver/data_mgr.py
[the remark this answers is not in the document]
dwsutherland: Not sure if the ZMQ message queue can be out of order? and if you hit `recv` on the subscriber you'd just get the oldest first.. (unless HWM has been hit, then you'd miss any pruned)

Different topics hit different topics of the data store. There's no confusion of topic, as the topic and respective data arrive together.

cylc/uiserver/workflows_mgr.py
dwsutherland: This is fixed, however, in order to get the API version without connecting a client I added `api` to the result of `get_scan_items_from_fs`. So there will still be errors until this is in, but as long as scan items remain the same thereafter, then the UI Server will only talk to those workflows with the same version (and should (optionally) restart workflows to version, otherwise).

oliver-sanders: Nice, thanks.

cylc/uiserver/data_store_mgr.py
```diff
@@ -0,0 +1,254 @@
...
+            future = asyncio.run_coroutine_threadsafe(
+                workflow_request(
+                    self.workflows_mgr.workflows[w_id]['req_client'],
+                    'pb_data_elements',
+                    args={'element_type': topic}
+                ),
+                self.loop
+            )
+            try:
+                _, new_delta_msg = future.result(5.0)
```
hjoliver:
```suggestion
                _, new_delta_msg = future.result(RECONCILE_TIMEOUT)
```

hjoliver: Jeez, finally found something :grin:

cylc/uiserver/data_store_mgr.py
```diff
@@ -0,0 +1,254 @@
...
+    def __init__(self, workflows_mgr):
+        self.workflows_mgr = workflows_mgr
+        self.data = {}
+        self.w_subs = {}
+        self.topics = {topic.encode('utf-8') for topic in DELTAS_MAP}
+        self.topics.add(b'shutdown')
+        self.loop = None
+        # Might be options other than threads to achieve
+        # non-blocking subscriptions, but this works.
+        self.executor = ThreadPoolExecutor()
```
hjoliver: I think this is probably the right way to go, as the subscriptions are presumably "I/O bound".

dwsutherland: Yeah, blocking.. However, there are some ways to avoid blocking.. i.e. Like not waiting around if there's no message(s) using the `flag=NOBLOCK` [recv](https://pyzmq.readthedocs.io/en/latest/api/zmq.html#zmq.Socket.recv) arg, or using a [poller](https://pyzmq.readthedocs.io/en/latest/api/zmq.html#zmq.Poller) .. but in that case you'd need to create a common list of sockets to loop over in a separate thread or as a timed callback in the main thread asyncio thread.. 
The current implementation is probably more event driven, and less burden/delay on the main thread and/or other subscriptions?

Open for other ideas.. But I think it should be a future improvement PR, if a better approach emerges.

hjoliver: Not sure if we're on the same wavelength - I was just trying to say that threads is correct IMO, for IO-bound callables, as opposed to process pool - the other option with concurrent.futures. (Or asyncio? but yes, we can look later if needed...).

dwsutherland: Definitely on board.. I was just saying the blocking bit is receiving messages, but you can make it not hang-around/block while waiting for the next publish (don't block if no messages have been published). This could be an idea, even on a separate thread.

hjoliver: Ah, roger that. :+1:

hjoliver:
I'll merge this, for compatibility with the back end; and @oliver-sanders comments above suggest he's as good as approved it.

PR #118, opened 2020-02-08
UIS Delta subscriptions & GraphQL null stripping back-end

dwsutherland:
These changes partially address: 
https://github.com/cylc/cylc-admin/blob/master/docs/proposal-subscriptions.md
As a back-end solution required to bring the proposal to fruition. 

Depends on:  https://github.com/cylc/cylc-flow/pull/3500
(They should be merge together, as there is one change that is breaking to the `BaseResolvers`)

Doesn't break Web UI.

**Features:**
- Published deltas of a single main loop iteration grouped into a single message are received (as topic `all`) and applied to the local data-store (instead of the same multi-topic ones).
- Use of Imported Cylc-Flow GraphQL backend and middleware to set undesired/empty field values to null, and strip these out of the execution result.
- New/Working GraphQL subscription of received Workflow Scheduler deltas:
uis_delta_subs_final

**Requirements check-list**

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -44,7 +44,7 @@ from cylc.flow.network.scan import MSG_TIMEOUT
 from cylc.flow.network.subscriber import WorkflowSubscriber, process_delta_msg
 from cylc.flow.data_store_mgr import (
     EDGES, FAMILIES, FAMILY_PROXIES, JOBS, TASKS, TASK_PROXIES, WORKFLOW,
-    DELTAS_MAP, apply_delta, generate_checksum
+    ALL_DELTAS, DELTAS_MAP, apply_delta, generate_checksum
 )
 from .workflows_mgr import workflow_request
 
@@ -62,10 +62,10 @@ class DataStoreMgr:
         self.workflows_mgr = workflows_mgr
         self.data = {}
         self.w_subs = {}
-        self.topics = {topic.encode('utf-8') for topic in DELTAS_MAP}
-        self.topics.add(b'shutdown')
+        self.topics = {ALL_DELTAS.encode('utf-8'), b'shutdown'}
         self.loop = None
         self.executors = {}
+        self.delta_queues = {}
 
     async def sync_workflow(self, w_id, *args, **kwargs):
         """Run data store sync with workflow services.
@@ -80,12 +80,16 @@ class DataStoreMgr:
         if w_id in self.w_subs:
             return
 
+        self.delta_queues[w_id] = {}
+
         # Might be options other than threads to achieve
         # non-blocking subscriptions, but this works.
         self.executors[w_id] = ThreadPoolExecutor()
         self.executors[w_id].submit(
+        self.executor.submit(
             partial(self.start_subscription, w_id, *args, **kwargs)
         )
+
         await self.entire_workflow_update(ids=[w_id])
 
     def purge_workflow(self, w_id):
@@ -95,6 +99,8 @@ class DataStoreMgr:
             del self.w_subs[w_id]
         if w_id in self.data:
             del self.data[w_id]
+        if w_id in self.delta_queues:
+            del self.delta_queues[w_id]
         self.executors[w_id].shutdown(wait=True)
         del self.executors[w_id]
 
@@ -130,6 +136,8 @@ class DataStoreMgr:
             w_id (str): Workflow external ID.
 
         """
+        for delta_queue in self.delta_queues[w_id].values():
+            delta_queue.put((w_id, topic, delta))
         # wait until data-store is populated for this workflow
         if w_id not in self.data:
             loop_cnt = 0
@@ -143,23 +151,24 @@ class DataStoreMgr:
             self.workflows_mgr.stopping.add(w_id)
             self.w_subs[w_id].stop()
             return
-        delta_time = getattr(
-            delta, 'time', getattr(delta, 'last_updated', 0.0))
-        # If the workflow has reloaded recreate the data
-        # otherwise apply the delta if it's newer than the previously applied.
-        if delta.reloaded:
-            if topic == WORKFLOW:
-                self.data[w_id][topic].CopyFrom(delta)
-            else:
-                self.data[w_id][topic] = {
-                    ele.id: ele
-                    for ele in delta.deltas
-                }
-            self.data[w_id]['delta_times'][topic] = delta_time
-        elif delta_time >= self.data[w_id]['delta_times'][topic]:
-            apply_delta(topic, delta, self.data[w_id])
-            self.data[w_id]['delta_times'][topic] = delta_time
-            self.reconcile_update(topic, delta, w_id)
+        for field, sub_delta in delta.ListFields():
+            delta_time = getattr(
+                sub_delta, 'time', getattr(sub_delta, 'last_updated', 0.0))
+            # If the workflow has reloaded recreate the data
+            # otherwise apply the delta if newer than the previously applied.
+            if sub_delta.reloaded:
+                if field.name == WORKFLOW:
+                    self.data[w_id][field.name].CopyFrom(sub_delta)
+                else:
+                    self.data[w_id][field.name] = {
+                        ele.id: ele
+                        for ele in sub_delta.deltas
+                    }
+                self.data[w_id]['delta_times'][field.name] = delta_time
+            elif delta_time >= self.data[w_id]['delta_times'][field.name]:
+                apply_delta(field.name, sub_delta, self.data[w_id])
+                self.data[w_id]['delta_times'][field.name] = delta_time
+                self.reconcile_update(field.name, sub_delta, w_id)
 
     def reconcile_update(self, topic, delta, w_id):
         """Reconcile local with workflow data-store.
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -43,8 +43,8 @@ from cylc.flow.network.server import PB_METHOD_MAP
 from cylc.flow.network.scan import MSG_TIMEOUT
 from cylc.flow.network.subscriber import WorkflowSubscriber, process_delta_msg
 from cylc.flow.data_store_mgr import (
-    EDGES, FAMILIES, FAMILY_PROXIES, JOBS, TASKS, TASK_PROXIES, WORKFLOW,
-    DATA_TEMPLATE, ALL_DELTAS, DELTAS_MAP, apply_delta, generate_checksum
+    EDGES, DATA_TEMPLATE, ALL_DELTAS, DELTAS_MAP, WORKFLOW,
+    apply_delta, generate_checksum, create_delta_store
 )
 from .workflows_mgr import workflow_request
 
@@ -86,10 +86,8 @@ class DataStoreMgr:
         # non-blocking subscriptions, but this works.
         self.executors[w_id] = ThreadPoolExecutor()
         self.executors[w_id].submit(
-        self.executor.submit(
             partial(self.start_subscription, w_id, *args, **kwargs)
         )
-
         await self.entire_workflow_update(ids=[w_id])
 
     def purge_workflow(self, w_id):
@@ -146,8 +144,8 @@ class DataStoreMgr:
                 loop_cnt += 1
                 continue
         if topic == 'shutdown':
+            self.delta_store_to_queues(w_id, topic, delta)
             self.workflows_mgr.stopping.add(w_id)
-            self.w_subs[w_id].stop()
             return
         for field, sub_delta in delta.ListFields():
             delta_time = getattr(sub_delta, 'time', 0.0)
@@ -164,9 +162,15 @@ class DataStoreMgr:
                 apply_delta(field.name, sub_delta, self.data[w_id])
                 self.data[w_id]['delta_times'][field.name] = delta_time
                 self.reconcile_update(field.name, sub_delta, w_id)
+
+        self.delta_store_to_queues(w_id, topic, delta)
+
+    def delta_store_to_queues(self, w_id, topic, delta):
         # Queue delta for graphql subscription resolving
-        for delta_queue in self.delta_queues[w_id].values():
-            delta_queue.put((w_id, topic, delta))
+        if self.delta_queues[w_id]:
+            delta_store = create_delta_store(delta, w_id)
+            for delta_queue in self.delta_queues[w_id].values():
+                delta_queue.put((w_id, topic, delta_store))
 
     def reconcile_update(self, topic, delta, w_id):
         """Reconcile local with workflow data-store.
@@ -202,13 +206,12 @@ class DataStoreMgr:
             try:
                 _, new_delta_msg = future.result(self.RECONCILE_TIMEOUT)
             except asyncio.TimeoutError:
-                logger.info(
+                logger.debug(
                     f'The reconcile update coroutine {w_id} {topic}'
                     f'took too long, cancelling the subscription/sync.'
                 )
                 future.cancel()
                 self.workflows_mgr.stopping.add(w_id)
-                self.w_subs[w_id].stop()
             except Exception as exc:
                 logger.exception(exc)
             else:
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -33,10 +33,10 @@ Subscriptions are currently run in a different thread (via ThreadPoolExecutor).
 """
 
 import asyncio
-import logging
-
 from concurrent.futures import ThreadPoolExecutor
+from copy import deepcopy
 from functools import partial
+import logging
 from time import sleep
 
 from cylc.flow.network.server import PB_METHOD_MAP
@@ -44,7 +44,7 @@ from cylc.flow.network.scan import MSG_TIMEOUT
 from cylc.flow.network.subscriber import WorkflowSubscriber, process_delta_msg
 from cylc.flow.data_store_mgr import (
     EDGES, FAMILIES, FAMILY_PROXIES, JOBS, TASKS, TASK_PROXIES, WORKFLOW,
-    ALL_DELTAS, DELTAS_MAP, apply_delta, generate_checksum
+    DATA_TEMPLATE, ALL_DELTAS, DELTAS_MAP, apply_delta, generate_checksum
 )
 from .workflows_mgr import workflow_request
 
@@ -136,8 +136,6 @@ class DataStoreMgr:
             w_id (str): Workflow external ID.
 
         """
-        for delta_queue in self.delta_queues[w_id].values():
-            delta_queue.put((w_id, topic, delta))
         # wait until data-store is populated for this workflow
         if w_id not in self.data:
             loop_cnt = 0
@@ -152,23 +150,23 @@ class DataStoreMgr:
             self.w_subs[w_id].stop()
             return
         for field, sub_delta in delta.ListFields():
-            delta_time = getattr(
-                sub_delta, 'time', getattr(sub_delta, 'last_updated', 0.0))
-            # If the workflow has reloaded recreate the data
-            # otherwise apply the delta if newer than the previously applied.
+            delta_time = getattr(sub_delta, 'time', 0.0)
+            # If the workflow has reloaded clear the data before
+            # delta application.
             if sub_delta.reloaded:
                 if field.name == WORKFLOW:
-                    self.data[w_id][field.name].CopyFrom(sub_delta)
+                    self.data[w_id][field.name].Clear()
                 else:
-                    self.data[w_id][field.name] = {
-                        ele.id: ele
-                        for ele in sub_delta.deltas
-                    }
-                self.data[w_id]['delta_times'][field.name] = delta_time
-            elif delta_time >= self.data[w_id]['delta_times'][field.name]:
+                    self.data[w_id][field.name].clear()
+                self.data[w_id]['delta_times'][field.name] = 0.0
+            # Apply the delta if newer than the previously applied.
+            if delta_time >= self.data[w_id]['delta_times'][field.name]:
                 apply_delta(field.name, sub_delta, self.data[w_id])
                 self.data[w_id]['delta_times'][field.name] = delta_time
                 self.reconcile_update(field.name, sub_delta, w_id)
+        # Queue delta for graphql subscription resolving
+        for delta_queue in self.delta_queues[w_id].values():
+            delta_queue.put((w_id, topic, delta))
 
     def reconcile_update(self, topic, delta, w_id):
         """Reconcile local with workflow data-store.
@@ -247,26 +245,18 @@ class DataStoreMgr:
             if not ids or kwargs['req_context'] in ids:
                 gathers += (workflow_request(**kwargs),)
         items = await asyncio.gather(*gathers)
-        new_data = {}
         for w_id, result in items:
             if result is not None and result != MSG_TIMEOUT:
                 pb_data = PB_METHOD_MAP[req_method]()
                 pb_data.ParseFromString(result)
-                new_data[w_id] = {
-                    EDGES: {e.id: e for e in getattr(pb_data, EDGES)},
-                    FAMILIES: {f.id: f for f in getattr(pb_data, FAMILIES)},
-                    FAMILY_PROXIES: {
-                        n.id: n
-                        for n in getattr(pb_data, FAMILY_PROXIES)},
-                    JOBS: {j.id: j for j in getattr(pb_data, JOBS)},
-                    TASKS: {t.id: t for t in getattr(pb_data, TASKS)},
-                    TASK_PROXIES: {
-                        n.id: n
-                        for n in getattr(pb_data, TASK_PROXIES)},
-                    WORKFLOW: getattr(pb_data, WORKFLOW),
-                    'delta_times': {
-                        topic: getattr(pb_data, WORKFLOW).last_updated
-                        for topic in DELTAS_MAP.keys()}
-                }
-
-        self.data.update(new_data)
+                new_data = deepcopy(DATA_TEMPLATE)
+                for field, value in pb_data.ListFields():
+                    if field.name == WORKFLOW:
+                        new_data[field.name].CopyFrom(value)
+                        new_data['delta_times'] = {
+                            key: value.last_updated
+                            for key in DATA_TEMPLATE
+                        }
+                        continue
+                    new_data[field.name] = {n.id: n for n in value}
+                self.data[w_id] = new_data
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -99,6 +99,7 @@ class UIServerGraphQLHandler(HubOAuthenticated, TornadoGraphQLHandler):
         self.graphiql = graphiql
         self.batch = batch
         self.backend = backend or get_default_backend()
+        self.strip_null = True
         # Set extra attributes
         for key, value in kwargs.items():
             if hasattr(self, key):
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -99,7 +99,6 @@ class UIServerGraphQLHandler(HubOAuthenticated, TornadoGraphQLHandler):
         self.graphiql = graphiql
         self.batch = batch
         self.backend = backend or get_default_backend()
-        self.strip_null = True
         # Set extra attributes
         for key, value in kwargs.items():
             if hasattr(self, key):
@@ -117,6 +116,16 @@ class UIServerGraphQLHandler(HubOAuthenticated, TornadoGraphQLHandler):
     @web.authenticated
     def prepare(self):
         super().prepare()
+
+    async def execute(self, *args, **kwargs):
+        # Use own backend, and TornadoGraphQLHandler already does validation.
+        return await self.schema.execute(
+            *args,
+            backend=self.backend,
+            variable_values=kwargs.get('variables'),
+            validate=False,
+            **kwargs,
+        )
 
 
 class SubscriptionHandler(websocket.WebSocketHandler):
--- a/cylc/uiserver/main.py
+++ b/cylc/uiserver/main.py
@@ -28,6 +28,9 @@ from typing import Any, Tuple, Type
 
 from tornado import web, ioloop
 
+from cylc.flow.network.graphql import (
+    GraphQLCoreBackend, IgnoreFieldMiddleware
+)
 from cylc.flow.network.schema import schema
 from graphene_tornado.tornado_graphql_handler import TornadoGraphQLHandler
 from jupyterhub.services.auth import HubOAuthCallbackHandler
@@ -39,6 +42,9 @@ from .websockets.tornado import TornadoSubscriptionServer
 from .workflows_mgr import WorkflowsManager
 
 logger = logging.getLogger(__name__)
+
+# Avoid re-instantiation
+MIDDLEWARE = [IgnoreFieldMiddleware()]
 
 
 class MyApplication(web.Application):
@@ -76,7 +82,7 @@ class CylcUIServer(object):
         self.workflows_mgr = WorkflowsManager(self)
         self.data_store_mgr = DataStoreMgr(self.workflows_mgr)
         self.resolvers = Resolvers(
-            self.data_store_mgr.data,
+            self.data_store_mgr,
             workflows_mgr=self.workflows_mgr)
 
     def _create_static_handler(
@@ -140,6 +146,8 @@ class CylcUIServer(object):
             clazz,
             schema=schema,
             resolvers=self.resolvers,
+            backend=GraphQLCoreBackend(),
+            middleware=MIDDLEWARE,
             **kwargs
         )
 
@@ -151,7 +159,11 @@ class CylcUIServer(object):
         """
         logger.info(self._static)
         # subscription/websockets server
-        subscription_server = TornadoSubscriptionServer(schema)
+        subscription_server = TornadoSubscriptionServer(
+            schema,
+            backend=GraphQLCoreBackend(),
+            middleware=MIDDLEWARE,
+        )
         return MyApplication(
             static_path=self._static,
             debug=debug,
@@ -166,14 +178,20 @@ class CylcUIServer(object):
                 self._create_handler("userprofile",
                                      UserProfileHandler),
                 # graphql handlers
-                self._create_graphql_handler("graphql",
-                                             UIServerGraphQLHandler),
-                self._create_graphql_handler("graphql/batch",
-                                             UIServerGraphQLHandler,
-                                             batch=True),
-                self._create_graphql_handler("graphql/graphiql",
-                                             GraphiQLHandler,
-                                             graphiql=True),
+                self._create_graphql_handler(
+                    "graphql",
+                    UIServerGraphQLHandler,
+                ),
+                self._create_graphql_handler(
+                    "graphql/batch",
+                    UIServerGraphQLHandler,
+                    batch=True
+                ),
+                self._create_graphql_handler(
+                    "graphql/graphiql",
+                    GraphiQLHandler,
+                    graphiql=True
+                ),
                 # subscription/websockets handler
                 self._create_handler("subscriptions",
                                      SubscriptionHandler,
--- a/cylc/uiserver/main.py
+++ b/cylc/uiserver/main.py
@@ -29,7 +29,7 @@ from typing import Any, Tuple, Type
 from tornado import web, ioloop
 
 from cylc.flow.network.graphql import (
-    GraphQLCoreBackend, IgnoreFieldMiddleware
+    CylcGraphQLBackend, IgnoreFieldMiddleware
 )
 from cylc.flow.network.schema import schema
 from graphene_tornado.tornado_graphql_handler import TornadoGraphQLHandler
@@ -42,9 +42,6 @@ from .websockets.tornado import TornadoSubscriptionServer
 from .workflows_mgr import WorkflowsManager
 
 logger = logging.getLogger(__name__)
-
-# Avoid re-instantiation
-MIDDLEWARE = [IgnoreFieldMiddleware()]
 
 
 class MyApplication(web.Application):
@@ -146,8 +143,8 @@ class CylcUIServer(object):
             clazz,
             schema=schema,
             resolvers=self.resolvers,
-            backend=GraphQLCoreBackend(),
-            middleware=MIDDLEWARE,
+            backend=CylcGraphQLBackend(),
+            middleware=[IgnoreFieldMiddleware],
             **kwargs
         )
 
@@ -161,8 +158,8 @@ class CylcUIServer(object):
         # subscription/websockets server
         subscription_server = TornadoSubscriptionServer(
             schema,
-            backend=GraphQLCoreBackend(),
-            middleware=MIDDLEWARE,
+            backend=CylcGraphQLBackend(),
+            middleware=[IgnoreFieldMiddleware],
         )
         return MyApplication(
             static_path=self._static,
--- a/cylc/uiserver/websockets/tornado.py
+++ b/cylc/uiserver/websockets/tornado.py
@@ -4,12 +4,13 @@
 # The file was copied from this revision:
 # https://github.com/graphql-python/graphql-ws/blob/cf560b9a5d18d4a3908dc2cfe2199766cc988fef/graphql_ws/tornado.py
 
-from inspect import isawaitable
+from inspect import isawaitable, isclass
 
 from asyncio import create_task, gather, wait, shield, sleep
 from asyncio.queues import QueueEmpty
 from tornado.websocket import WebSocketClosedError
 from graphql.execution.executors.asyncio import AsyncioExecutor
+from graphql.execution.middleware import MiddlewareManager
 from graphql_ws.base import ConnectionClosedException, BaseConnectionContext, BaseSubscriptionServer
 from graphql_ws.observable_aiter import setup_observable_extension
 from graphql_ws.constants import (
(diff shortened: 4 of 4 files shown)


dwsutherland:
Note: should fail until Cylc-Flow sibling PR is merged.

Also need to fix bug in GraphiQL loading the schema:
GraphiQL_Broken_Schema_Load

dwsutherland:
Ok fixed the GraphiQL problem.

kinow:
@dwsutherland any reason for having a separate section for what was removed, but not for what was added and updated?

From what I understand, it looks like added and updated are both under `workflows {}`. Which should work fine, but will require an extra step to check whether the task or job exists, then update if so, or add if not. More for my own curiosity :grimacing:

dwsutherland:
> From what I understand, it looks like added and updated are both under workflows {}. Which should work fine, but will require an extra step to check whether the task or job exists, then update if so, or add if not. More for my own curiosity 😬

@kinow  - Trying to figure out what part of the code you're referring too 😕   (sorry)

kinow:
> 
> 
> > From what I understand, it looks like added and updated are both under workflows {}. Which should work fine, but will require an extra step to check whether the task or job exists, then update if so, or add if not. More for my own curiosity 😬
> 
> @kinow - Trying to figure out what part of the code you're referring too 😕 (sorry)

Hi @dwsutherland 

Sorry, I am talking about the GraphQL output. The query response has `deltas { pruned: ..., workflows: ...}`. I was just wondering why we didn't have something like `deltas { pruned: ..., added: ..., modified: ... }`

Sorry for not being very clear.

hjoliver:
>  Trying to figure out what part of the code you're referring to

In gif above (in the Issue description) there's a "pruned" section that singles out what's been removed in the latest data push.  @kinow is wondering (I think, based on a conversation we had on Friday) if it is also possible to single out what's been added?  (then I guess the main body of the data structure would contain just what changed - but has not been added or removed .., .rather than what's now present, which includes what's new and what's changed).

hjoliver:
(Doh, I was typing at the same time as @kinow :grimacing: )

kinow:
> 
> 
> (Doh, I was typing at the same time as @kinow 😬 )

But your explanation was much better than mine. I'm working exactly on this, but still doing JS now, not querying the endpoint yet. So good timing for you to have a look at this @dwsutherland 😬

dwsutherland:
Ah, I see.. I haven't really made that distinction at the workflow level either, but intend to do so (for the sake of our event driven future).. I guess if something's been added and then modified (if that's even possible in one push), they could be in both `added` and `modified` (is that okay? the receiver would then need to order their operations)

kinow:
> 
> 
> Ah, I see.. I haven't really made that distinction at the workflow level either, but intend to do so (for the sake of our event driven future).. I guess if something's been added and then modified (if that's even possible in one push), they could be in both `added` and `modified` (is that okay? the receiver would then need to order their operations)

Ah, I guess it would still work. I would just go and:

1. first remove the pruned items
2. then add the new ones
3. finally update (including anything that was added)

We can do that later. I'll start using this code for the JS code and send a functional reviewer later (not looking much at the code changes, but more how it works, whether it does what it is supposed to do, etc) 👍 Thanks!

dwsutherland:
> Ah, I guess it would still work. I would just go and:
> 
> 1. first remove the pruned items
> 2. then add the new ones
> 3. finally update (including anything that was added)
> 

Maybe do the pruning last, although I would hope/think it wouldn't matter as something shouldn't be pruned and updated in the same push.

kinow:
@dwsutherland, the original query used by the tree view (current & infinite) has something like:

```graphql
query {
  workflows (ids: ["five"]) {
    ...
    taskProxies (sort: ...) {
      ...
      firstParent { }
      task { }
      jobs (sort: ...) { }
    }
  }
}
```

Using GraphiQL, I created the following subscription for deltas.

```graphql
subscription {
  deltas (id: "five") {
    pruned { familyProxies taskProxies jobs }
    workflow {
      id
      status
      taskProxies {
        id
        name
        state
        ...
        firstParent {}
        ... # same structure as in the query
      }
      familyProxies { }
    }
  }
}
```

The subscription is submitted and I am getting the results, but the `deltas.workflow.taskProxies` (which I am writing the code to add/update now) never includes `firstParent`, `task`, or `jobs`. Only a flat list of `key: value`.

Is that intentional? The JS code fails when I add a delta `TaskProxy` to the tree, because the code for the tree view is expecting a `TaskProxy` with `task`, `jobs`, and the `firstParent` (to create hierarchy this last one).

Thanks!

dwsutherland:
> The subscription is submitted and I am getting the results, but the `deltas.workflow.taskProxies` (which I am writing the code to add/update now) never includes `firstParent`, `task`, or `jobs`. Only a flat list of `key: value`.
> Q1: Is that intentional? The JS code fails when I add a delta `TaskProxy` to the tree, because the code for the tree view is expecting a `TaskProxy` with `task`, `jobs`, and the `firstParent` (to create hierarchy this last one).

I need to fix how this field is resolved.. I was in two minds to where the resolver would need to look, but this use case makes it clear that sub-field resolution should use the local data-store (not the other delta data) post delta application...

> Q2: just to confirm, the sorting (`sort( key: ... )`) is not supported in subscriptions?

Should be able to reinstate this for subscriptions
 
> Q3: in the query, I can send `ids: ["..."]`, which I think accepts a `workflow.id` or `workflow.name`. But in the delta subscription, I think it is accepting only `workflow.id`. The components/views are all using the `workflow.name` for now (they get it from the URL, `/#/tree/<workflow-name>`). Should we have a standard and have `id` and `ids` both to allow the same?

Yes, use `id` for the update application (as ID is set for every delta element), however `name` should be in the initial data query and the view can display that ... We only send what's changed (other than ID of a delta element).

I'm working on some changes in the backend to distinguish between `added` and `updated` deltas.. Will also work on nested field resolvers & sorting/filtering!

kinow:
Thanks for the quick reply @dwsutherland ! It's been super easy to work with the deltas so far (easier than I expected). I've left most of the logic in-place in my branch, so once you have the updates for the code I will check it out and test it while finishing the JS code.

dwsutherland:
@kinow - Also, I didn't implemented nested resolving because it's not consistently possible.. i.e.:
```
a {
  b {
    c
    d
  }
}
```
What if ``a` & `d` is updated but `b` isn't? Or if `a` is a delta but it's reference to `b` hasn't changed (so the delta doesn't contain that value)

It's technically possible, but would require an argument of each level of nest to specify whether to resolve from the delta info or the data-store.

dwsutherland:
In the case of `firstParent` & `jobs` we should be able to implement that because new `taskProxy` will have a reference to `firstParent` in the delta, and new jobs will put a reference in `taskProxy` .. However a change in Job state won't result in `taskProxies { jobs { . . . }}` showing up because the reference didn't change.

kinow:
Hmmm, tricky. I can handle the flat structure on the client side. I will have to think a little. We don't have a flat structure in the initial query.

It is not easy to fetch tasks or jobs in the original workflow data. These are under `workflow.taskProxies` or `workflow.taskProxies.jobs` respectively. There is no lookup table anywhere.

So I think the options I have now are iterate in JS (which could have performance issues) or try to use a flat structure for the tree component. So instead of using `workflow.taskProxies.jobs`, I would use `workflow.taskProxies` and `workflow.jobs` in the query, and then associate them in JS. That way I would be able to receive deltas `delta.workflow.jobs` and just add to the Vuex store. 

🤔

dwsutherland:
> Hmmm, tricky. I can handle the flat structure on the client side. I will have to think a little. We don't have a flat structure in the initial query.
> 
> It is not easy to fetch tasks or jobs in the original workflow data. These are under `workflow.taskProxies` or `workflow.taskProxies.jobs` respectively. There is no lookup table anywhere.
> 
> So I think the options I have now are iterate in JS (which could have performance issues) or try to use a flat structure for the tree component. So instead of using `workflow.taskProxies.jobs`, I would use `workflow.taskProxies` and `workflow.jobs` in the query, and then associate them in JS. That way I would be able to receive deltas `delta.workflow.jobs` and just add to the Vuex store.
> 
> 🤔

Perhaps we can get what we need from `firstParent` alone .. And construct the ordered infinite-tree object reference from that (instead of using flatten)... This will be expensive on initial query, but easy with the deltas (?)

kinow:
>This will be expensive on initial query, but easy with the deltas (?)

This is exactly what I had in mind. A bit expensive for the JS code to create it, but once we have the initial structure created, updating with deltas will be a breeze.

I'll start changing the code now and see how it looks like, if there are any other issues. But I think that should be possible - and I think the deltas will be simpler that way too.

dwsutherland:
(BTW - I think it possible to do nested deltas, but as mentioned, it would require using the data-store to supplement the node relations (hence a bit or work) Or always sending the relationship IDs in the delta elements)

dwsutherland:
> I'll start changing the code now and see how it looks like, if there are any other issues. But I think that should be possible - and I think the deltas will be simpler that way too.

No harm in trying 🍀

dwsutherland:
Delta split into `added`, `updated`, `pruned`:
uis_added_updated

dwsutherland:
Next thing to do is enable subscription nesting with optional data-store fetching (i.e. to supplement the node relations not sent with the delta)... Possibly achieved by:
- Structure/store the delta data in map form for resolver referencing (re-created every delta)
- Create deltas resolvers and/or modify to fetch data from either data-store or delta-store.

kinow:
> Delta split into `added`, `updated`, `pruned`:

Thanks @dwsutherland ! Synced my local branch and did a bit of manual smoke testing.

First tried a simple query with jobs.

image

No issues.

Then tried a simple subscription with pruned/added/updated jobs.

```graphql
subscription {
  deltas (ids: ["kinow|five"]) {
    id
    pruned {
      jobs
    }
    added {
      jobs {
        id
      }
    }
    updated {
      jobs {
        id
      }
    }
  }
}
```

image

Also no issues :tada: 

I noticed the filter for the deltas also changed from `id` to `ids` (got an error in the console for Cylc UI on my `deltas-1` branch), but it's nice as it matches the name used to filter the queries too :+1:

kinow:
> Next thing to do is enable subscription nesting with optional data-store fetching (i.e. to supplement the node relations not sent with the delta)... Possibly achieved by:
> 
>     * Structure/store the delta data in map form for resolver referencing (re-created every on every delta)
> 
>     * Create delta resolvers and/or modify to fetch data from either data-store or delta-store.

Brilliant, I think with this I should be able to use the `firstParent` in the UI, and finish building the hierarchy when the data is added/updated :rocket:

kinow:
@dwsutherland this morning I replied to a comment in an ApolloClient issue, and for that I had to update and test my branch. I found one strange change. On `master` this returns `familyProxies: []` for my workflow five (doesn't use families, only `root`).

```graphql
query {
  workflows (ids: ["kinow|five"]) {
    familyProxies (exids: ["root"]) {
      id
    }
  }
}
```

But using this branch I got a `familyProxies: null`. That breaks the Tree view. I could change the code to work around that if necessary, but it would be easier if we had the old behavior of returning an empty list instead :grimacing: 

Cheers
Bruno

p.s.: let me know if you need any help with deltas as we did with WebSockets (i.e. I would basically rely a lot on your help, but could try to write some Python code for the deltas)

dwsutherland:
The client now decides whether to strip empty fields by the presence of `stripNull` in the `variables` (`variable_values`) of the subscription:
stripNull_full_args

The query null stripping at the UIS isn't working (via GraphiQL) for what ever reason (will investigate), but the main use in subscriptions is working (as shown above).

@kinow - I fixed the null return issue above (It was setting empties to `null` but not stripping them):
image

dwsutherland:
Ok, this should be ready for review now.

dwsutherland:
Model 2 is in play here:
image

- A delta comes in from the scheduler, is applied to the UIS data-store and a delta-store is created from it, then the delta-store is added to the queues of the subscriptions that are subscribed to that workflow, the subscription resolvers use the deltaStore (or optionally the dataStore).
- This means the same delta-store object is shared amongst the potentially many subscriptions,
it is disposed of naturally once all subscriptions lose reference to it (i.e. move to the next delta).

dwsutherland:
> That one is OK, but this one has two unit tests failing I think. Could you have a look later?

@kinow  - Yes, to be expected, `cylc-flow` `master` doesn't have the sibling [PR](https://github.com/cylc/cylc-flow/pull/3500) in yet:
```
E   ModuleNotFoundError: No module named 'cylc.flow.network.graphql'
```

kinow:
@dwsutherland whenever there is a new `TaskProxy`, it appears that the array returned for `taskproxies` always contains a `null` element.

image

Is that expected? Did I forget to filter something, or maybe I have something wrong with my current query?

```graphql
subscription {
  deltas (ids: ["kinow|five"], stripNull: true) {
    id
    added (stripNull: true) {
      cyclePoints: familyProxies(ids: ["root"], stripNull: true) {
        cyclePoint
      }
			taskProxies(sort: { keys: ["cyclePoint"] }, stripNull: true) {
        id
        name
        state
        isHeld
        cyclePoint
        latestMessage
        firstParent {
          id
          name
          cyclePoint
          state
        }
        task {
          meanElapsedTime
          name
        }
        jobs(sort: { keys: ["submit_num"], reverse:true }) {
          id
          batchSysName
          batchSysJobId
          host
          startedTime
          submittedTime
          finishedTime
          state
          submitNum
        }
      }
      familyProxies (exids: ["root"], sort: { keys: ["firstParent"]}) {
        id
        name
        state
        cyclePoint
        firstParent {
          id
          name
          cyclePoint
          state
        }
      }
    }
    pruned {
      familyProxies
      taskProxies
      jobs
    }
  }
}
```

dwsutherland:
> Is that expected? Did I forget to filter something, or maybe I have something wrong with my current query?

At the moment it is, because the presence of `stripNull: true` as a variable is what triggers the stripping.. 

About to change this behaviour to only need it's presence as an argument (it's default is `stripNull: false` everywhere otherwise)

kinow:
One more from today's tests @dwsutherland 

The following is a simplified version of the current query/subscription from Cylc UI, where we exclude the `root` family (you gave this tip some time ago actually, as before we filtered it in the code):

```graphql
query {
  workflows (ids: ["kinow|five"]) {
    familyProxies (exids: ["root"]) {
      id
    }
  }
}
```

Now with deltas, it looks like it is not excluded.

```graphql
subscription {
  deltas (ids: ["kinow|five"]) {
    id
    added {
      familyProxies (exids: ["root"]) {
        id
      }
    }
  }
}
```

This always returns the root family (when it's added to the pool). I **guess** it might be the `exids` ID part. The family proxy ID of a root family proxy, is actually something like `"id": "kinow|five|20130921T1200+12|root"` (copied from GraphiQL). So perhaps for the query it supports excluding by name & id, but subscription is looking only at the ID? If so, would it be possible to support the name for subscriptions too? 

Thanks!

kinow:
Another thing I noticed today, I added more fields than can be updated, and will remove later if necessary.

I remembered that some time ago we added an option to move workflows from one computer to another. So I thought that the workflow `port` and `host` could change. I wasn't expecting that to change, but I got the port number as `0` in the updated.

image

That looks like a bug @dwsutherland ?

dwsutherland:
> So I thought that the workflow port and host could change. I wasn't expecting that to change, but I got the port number as 0 in the updated.

Ah you spotted it! Yes, not sure how to get around other than setting `deltaStore: false` for `workflows`  argument.. or filtering out `0`  as well (currently ignore `(0, 0.0, False)`

dwsutherland:
Update:  `stripNull` is no longer needed in the variables to trigger the stripping.

hjoliver:
"Model 2" above seems pretty sensible, on the face of it :+1:

dwsutherland:
Can get initial "burst" of data-store data when you first subscribe now:
image


@kinow - I fixed that filter issue:
image

dwsutherland:
The other root subscription entry points have been added i.e.:
```
subscription {
  workflows {
    id
    name
    status
    stateTotals
  }
}
```
The WUI `master` branch is now reactive:
reactive_root_subs


Actually smoother than I thought... A the moment WUI `master` is just subscribed to the `dataStore`, so is pushed all requested data on every delta.

oliver-sanders:
The WUI is a lot slicker now!

kinow:
Posting here too what I posted on Riot as it's easier to reference/search @dwsutherland 

image

>(what I remember of that day: I was testing several things in JS, so I didn't need to touch the backend; jupyterhub+uiserver were running since the morning (though maybe I restarted it once?), and my workflow five too; I restarted the UI several times, and I had only 1 tab on firefox; occasionally I would open a tab on chromium to use the vue dev tools; not sure if any of that helps; will test again tomorrow; 👋🛏︎)

kinow:
@dwsutherland I deleted my existing branches, then re-synced repos, and checked out the branches for uiserver and flow.

Then deleted my `venv` folder, and recreated everything from start, re-installing uiserver, then flow. My venv has uiserver from this branch, and flow from the 3500 PR (confirmed).

Started jupyterhub, and flow (five). The jupyterhub python process was using ~45GB, and five using its regular ~400MB of memory. Then started closing everything.

Finally, opened one browser window to browse localhost:8000 and launch/spawn the uiserver process. Then closed my browser. I had GIMP (with screenshots), terminals, and a mousepad with notes. Nothing else running.

Then took the first screenshot (included watch with time).

image

15:12, ~66MB resident memory.

After a little over 10 minutes, took another screenshot.

image

15:24, ~75MB resident memory.

And then I waited until it reached 80MB.

image

15:33, ~82 MB resident memory.

hjoliver:
That does look suspiciously leaky :thinking:

dwsutherland:
Tried to have a look for the UIS leak:

**GScan (Only):**
13:59
```
Server:   75.03 KB
Workflow Mgr:   75.03 KB
Data Store Mgr:   75.03 KB
Resolvers:   75.03 KB
App/Handlers:  541.51 KB
data-store:   30.48 KB|edges: 23,    6.55 KB|families: 7,    1.41 KB|family_proxies: 28,    6.86 KB|jobs: 22,    5.84 KB|tasks: 8,    1.55 KB|task_proxies: 26,    6.50 KB|workflow: 1,   80     B|delta_times: 7,  968     B
```
14:12
```
Server:   77.38 KB
Workflow Mgr:   77.38 KB
Data Store Mgr:   77.38 KB
Resolvers:   77.38 KB
App/Handlers:  543.85 KB
data-store:   32.35 KB|edges: 29,    7.67 KB|families: 7,    1.41 KB|family_proxies: 35,    8.02 KB|jobs: 20,    4.45 KB|tasks: 8,    1.55 KB|task_proxies: 32,    7.48 KB|workflow: 1,   80     B|delta_times: 7,  968     B
```
**GScan + Tree:**
14:24
```
Server:   77.82 KB
Workflow Mgr:   77.82 KB
Data Store Mgr:   77.82 KB
Resolvers:   77.82 KB
App/Handlers:  544.30 KB
data-store:   30.48 KB|edges: 23,    6.55 KB|families: 7,    1.41 KB|family_proxies: 28,    6.86 KB|jobs: 22,    5.84 KB|tasks: 8,    1.55 KB|task_proxies: 26,    6.50 KB|workflow: 1,   80     B|delta_times: 7,  968     B
```

**GScan + Tree + GraphiQL:**
14:30
```
Server:   90.42 KB
Workflow Mgr:   90.42 KB
Data Store Mgr:   90.42 KB
Resolvers:   90.42 KB
App/Handlers:  557.97 KB
data-store:   30.48 KB|edges: 23,    6.55 KB|families: 7,    1.41 KB|family_proxies: 28,    6.86 KB|jobs: 22,    5.84 KB|tasks: 8,    1.55 KB|task_proxies: 26,    6.50 KB|workflow: 1,   80     B|delta_times: 7,  968     B
```

**WUI shutdown:**
14:35
```
Server:   89.70 KB
Workflow Mgr:   89.70 KB
Data Store Mgr:   89.70 KB
Resolvers:   89.70 KB
App/Handlers:  557.24 KB
data-store:   30.48 KB|edges: 23,    6.55 KB|families: 7,    1.41 KB|family_proxies: 28,    6.86 KB|jobs: 22,    5.84 KB|tasks: 8,    1.55 KB|task_proxies: 26,    6.50 KB|workflow: 1,   80     B|delta_times: 7,  968     B
```

**Workflow Shutdown:**
```
Server:   28.54 KB
Workflow Mgr:   28.54 KB
Data Store Mgr:   28.54 KB
Resolvers:   28.54 KB
App/Handlers:  496.90 KB
data-store:  248     B
```

How ever throughout the workflow running the memory (`data` below) of the UIServer (not JupyterHub) grew from ~18Mb when the workflow wasn't running to 151Mb:
Screenshot from 2020-05-07 14-32-37


This footprint didn't drop when I stopped the workflow, however, it did stop growing..

So the hunt continues (it's not the data objects though).. Perhaps the HWMark of ZMQ?

kinow:
Doing a bit of testing in the UI with these branches, might spend a bit of time today before calling it a day to look at what else I can find about it too :+1: but looks like you are on the right path, getting closer to solving this issue :) :clap:

dwsutherland:
@kinow, @hjoliver - Testing this on `master`, and same thing.. It went from ~70Mb to ~80Mb in 10min:
Screenshot from 2020-05-07 15-11-32

So not this PRs problem.. 

Maybe we should open another issue? 
If this PR has made it worse, it may be that ZMQ High Water Mark (HWM), because there is more being published by the workflow.

kinow:
>@kinow, @hjoliver - Testing this on master, and same thing.. It went from ~70Mb to ~80Mb in 10min:

Ah! Excellent!

Let me just run a few more tests to see why I got that 10GB the other day. Now I am testing with subscriptions running in the UI. I will check my results against `master`, and if I get same results there, +1 to move it to the backlog and approve this one :+1:

kinow:
@dwsutherland while testing the UI today found a few issues in JS. Like `null` jobs. But I couldn't find any `null` in the GraphQL response. So I am assuming I am doing something wrong in the JS code.

But there was another issue that happened when I submitted a query that expected variables (e.g. the workflowId to use in `deltas(workflows: [$workflowId])`, but I forgot to add the variables in GraphiQL. There were also 5 other subscription, so maybe one of the other ones caused it?

```
2020-05-07 16:25:28,007 cylc.flow.network.resolvers WARNING  Traceback (most recent call last):
  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/network/resolvers.py", line 387, in subscribe_delta
    self.data_store_mgr.data[w_id]
KeyError: 'kinow|families2'
```

The value in `KeyError` is the id of one of my suites that I was running during the tests. Any idea if that's a possible new bug in the resolver for the deltas (because of the `subscribe_delta` in the traceback)?

dwsutherland:
> The value in KeyError is the id of one of my suites that I was running during the tests. Any idea if that's a possible new bug in the resolver for the deltas (because of the subscribe_delta in the traceback)?

@kinow  - Fixed (in `cylc-flow`) ... I check for the key in the data store before adding the sub queue, it happened because the data store wasn't available at the time (as the sync must have been requesting the initial load from the workflow to build it).

kinow:
> > The value in KeyError is the id of one of my suites that I was running during the tests. Any idea if that's a possible new bug in the resolver for the deltas (because of the subscribe_delta in the traceback)?
> 
> @kinow - Fixed (in `cylc-flow`) ... I check for the key in the data store before adding the sub queue, it happened because the data store wasn't available at the time (as the sync must have been requesting the initial load from the workflow to build it).

That was fast!! :rocket:

dwsutherland:
> That was fast!! 🚀

Not sure where that `null` is popping up, maybe can happen with a wrong combination of strip nulls I think ... but the current WUI `master` doesn't use it I think.

kinow:
@dwsutherland screen shots of my tests here: https://gist.github.com/kinow/7a80583b44dcbdaf7dfc622350ac2a6e

You have probably read it all in the Riot channel. The last image is from when I attached a profiler in the running `cylc-uiserver` to have a quick peek at it. Only thing that called my attention was the number of threads? Don't remember seeing that before (confirmed on `htop` after enabling the option to show user threads). Probably to handle workflows data flow? I remember somewhere in uiserver we have threads. Maybe not important, but just in case...

kinow:
(From Riot)

With workflow `five`, the following query is returning only the `foo` task.

```graphql
subscription ($workflowId: ID) {
  deltas (workflows: [$workflowId]) {
    id
    added {
      taskProxies(sort: { keys: ["cyclePoint"] }) {
        ...TaskProxyData
    }
  }
}


fragment TaskProxyData on TaskProxy {
  id
  name
  state
  isHeld
  cyclePoint
  latestMessage
  firstParent {
    id
    name
    cyclePoint
    state
  }
  task {
    meanElapsedTime
    name
  }
}
```

And variables

```js
{
  "workflowId": "kinow|five"
}
```

I'm getting only the `delta.added` for the `foo` task proxy. Here's the part with `foo` and `bar` in the `suite.rc` file.

```
    [[dependencies]]
        [[[R1]]]
            graph = "prep => foo"
        [[[PT12H]]]
            graph = "foo[-PT12H] => foo => bar"
```

dwsutherland:
> I'm getting only the delta.added for the foo task proxy. Here's the part with foo and bar in the suite.rc file.

Yeah... need to make `ghosts: true` .. Going to change the default.

cylc/uiserver/data_store_mgr.py
```diff
@@ -140,26 +144,33 @@ def update_workflow_data(self, topic, delta, w_id):
...
-            self.reconcile_update(topic, delta, w_id)
+        for field, sub_delta in delta.ListFields():
+            delta_time = getattr(sub_delta, 'time', 0.0)
+            # If the workflow has reloaded clear the data before
+            # delta application.
+            if sub_delta.reloaded:
+                if field.name == WORKFLOW:
+                    self.data[w_id][field.name].Clear()
+                else:
+                    self.data[w_id][field.name].clear()
```
oliver-sanders, line 158: What's going on here, is there a difference between `Clear` and `clear`?

dwsutherland: Ones a Protobuf element (i.e. `PbWorkflow`), so uses `Clear`, the other's a dictionary.

kinow:
From RIOT

Re-synced branches this morning after approving it some time ago. There were now `null` values in the GraphQL response of the UI subscription. This caused task proxies not to be removed (as a node would get a `firstParent { name: null }` I think).

image

Adding `stripNull: true`, in GraphiQL fixed the problem. But it failed with ApolloClient. See query used [in this gist](https://gist.github.com/kinow/772c59a938fa74ee6cd8fc72f6e18aed).

Looks like the `__typename` causes the initial data burst to be empty. That `__typename` is put in the query by ApolloClient, and is used in the UI tree component to decide how to render each node.

Having both `__typename` and `stripNull: true` results in empty initial data burst. If I remove `stripNull` then the query works.

kinow:
@dwsutherland another comment/question, not related to the issue above. I decided to have a quick look at the Graph component this morning, checking if the delta subscription would work.

I don't know much about the Graph component & JS cytoscape, but guessed it was still worth to use the momentum to try and learn and review this PR at the same time.

With this query

<details>
<summary>collapsed deltas query with edges</summary>
```
subscription ($workflowId: ID) {
  deltas (workflows: [$workflowId], stripNull: true) {
    added {
      edges {
        id
        source
        target
        suicide
        cond
      }
    }
    updated {
      edges {
        id
        source
        target
        suicide
        cond
      }
    }
    pruned {
      edges
    }
  }
}
```
</details>

I get the initial burst of data, and eventually added and pruned data (I guess for `five` we never get updated edges as the workflow cycles are always the same?).

However, every now and then I also get this empty response.

```js
{
  "deltas": {
    "added": {},
    "updated": {},
    "pruned": {}
  }
}
```

Is that correct? I was expecting to get data iff there was some data in the server to be sent to the client.

image

dwsutherland:
> Having both `__typename` and `stripNull: true` results in empty initial data burst. If I remove stripNull then the query works.

Fixed (in `cylc-flow` PR)

dwsutherland:
> Is that correct? I was expecting to get data iff there was some data in the server to be sent to the client.

I get the initial burst:
```
{
  "deltas": {
    "added": {
      "edges": [
        {
          "id": "sutherlander|five|foo.20130808T0000Z|bar.20130808T0000Z",
          "source": "sutherlander|five|20130808T0000Z|foo",
          "target": "sutherlander|five|20130808T0000Z|bar",
          "suicide": false,
          "cond": false
        },
        {
          "id": "sutherlander|five|prep.20130808T0000Z|foo.20130808T0000Z",
          "source": "sutherlander|five|20130808T0000Z|prep",
          "target": "sutherlander|five|20130808T0000Z|foo",
          "suicide": false,
          "cond": false
        }
      ]
    },
    "updated": {},
    "pruned": {}
  }
}
```
And then added and pruned thereafter...

I could probably strip `deltas` further (i.e. not send if sub-fields are empty), it happens because the whole delta-store is yielded with it's skeleton structure.

kinow:
@dwsutherland synced this branch to pick up latest fix for the `__typename` with `stripNull: true`.

That appears to have fixed the initial data burst. But the JS code failed to render the tree again. A bit of debugging the queries again, it looks like another interesting issue.

Here's what I am getting in `deltas.updated.jobs` **with `stripNull: false`**.

image

Now `deltas.updated.jobs` **with `stripNull: true`**.

image

@dwsutherland it looks like when `stripNull: true`, **the updated jobs in the deltas are always parent-less**. Without the `firstParent`, I can still find the node parent through its ID, but so far I managed to link nodes without having to parse the ID. Is that something simple to fix in the backend?

dwsutherland:
> @dwsutherland it looks like when stripNull: true, the updated jobs in the deltas are always parent-less. Without the firstParent, I can still find the node parent through its ID, but so far I managed to link nodes without having to parse the ID. Is that something simple to fix in the backend?

This works:
```
      jobs (stripNull: true) {
        id
        state
        firstParent: taskProxy(stripNull: false, deltaStore: false) {
          id
          __typename
        }
      }
```

It's because the jobs `taskProxy` field isn't update/in-the-delta .. This will lookup the data-store for the missing relationship

kinow:
Tested, and indeed works! Thanks @dwsutherland !

dwsutherland:
> Tested, and indeed works! Thanks @dwsutherland !

Ok original should work now:
```
      jobs (stripNull: true) {
        id
        state
        firstParent: taskProxy {
          id
          __typename
        }
      }
```
Ignored stripping of sub-field if it's return type is in `cylc.flow.network.schema.NODE_MAP`.

oliver-sanders:
> It's because the jobs taskProxy field isn't update/in-the-delta .. This will lookup the data-store for the missing relationship

Aaahhh, that's what the `deltaStore` argument is for.

dwsutherland:
> > It's because the jobs taskProxy field isn't update/in-the-delta .. This will lookup the data-store for the missing relationship
> 
> Aaahhh, that's what the `deltaStore` argument is for.

It looks up the relationship from the data-store without the arg now, but will still determine whether to try retrieve the node from the delta-store or data-store.

kinow:
@dwsutherland I have both branches synced (did that after our Riot meeting some days ago).

And have also finished the code and tests in the WUI. So today and tomorrow are days that I am planning to spend a lot of time looking at the screen while it runs `five` (tasks attached to ROOT/cyclepoint) and `families2` (a workflow I copied from @hjoliver that uses families, and has at least one family whose parent is another family, like ROOT -> FAM1 -> FAM2 -> tasks).

During development I spent a looooong time with `five`. So decided to focus more on `families2` today. And I noticed that after several minutes running (+10min, some times +50 minutes) I would get an error as the JS code tried to **add a job, but couldn't find its parent**.

After some adding some try/catch with a `debugger` statement, I managed to capture the moment the error occurs.

And it's strange that apparently I am receiving a Job from the GraphQL endpoint before the cyclepoint and tasks? See screenshot below.

image

The cycle point for the year 2772 is missing. The JS code does some business logic when adding nodes to the tree, like cycle points, tasks, families, etc. However, what you are seeing in the browser console is the output of the `lookup`, a helper `Map` that I use to access elements in the tree.

For cycle points, all that I do when adding one is check `if (cyclePoint && !this.lookup.has(cyclePoint.id)) { this.lookup.set(cyclePoint.id, cyclePoint) ... }`. So cycle points are always added, as long as they do not exist already.

Here's the GraphQL deltas that I received when the exception occurred.

<details>
<summary>GraphQL deltas response...</summary>
```js
{
  "id": "kinow|families2",
  "shutdown": false,
  "added": {
    "jobs": [
      {
        "id": "kinow|families2|27720611T1203+12|qux|1",
        "firstParent": {
          "id": "kinow|families2|27720611T1203+12|qux",
          "__typename": "TaskProxy"
        },
        "batchSysName": "background",
        "host": "localhost",
        "state": "ready",
        "submitNum": 1,
        "__typename": "Job"
      }
    ],
    "__typename": "Added"
  },
  "updated": {
    "taskProxies": [
      {
        "id": "kinow|families2|27700611T1203+12|qux",
        "state": "succeeded",
        "isHeld": false,
        "latestMessage": "succeeded",
        "firstParent": {
          "id": "kinow|families2|27700611T1203+12|FAM6",
          "name": "FAM6",
          "cyclePoint": "27700611T1203+12",
          "state": "succeeded",
          "__typename": "FamilyProxy"
        },
        "task": {
          "meanElapsedTime": 20,
          "name": "qux",
          "__typename": "Task"
        },
        "__typename": "TaskProxy"
      },
      {
        "id": "kinow|families2|27720611T1203+12|qux",
        "state": "ready",
        "isHeld": false,
        "latestMessage": "",
        "firstParent": {
          "id": "kinow|families2|27720611T1203+12|FAM6",
          "name": "FAM6",
          "cyclePoint": "27720611T1203+12",
          "state": "ready",
          "__typename": "FamilyProxy"
        },
        "task": {
          "meanElapsedTime": 20,
          "name": "qux",
          "__typename": "Task"
        },
        "__typename": "TaskProxy"
      }
    ],
    "jobs": [
      {
        "id": "kinow|families2|27700611T1203+12|qux|1",
        "firstParent": {
          "id": "kinow|families2|27700611T1203+12|qux",
          "__typename": "TaskProxy"
        },
        "finishedTime": "2020-06-11T15:31:36+12:00",
        "state": "succeeded",
        "__typename": "Job"
      }
    ],
    "familyProxies": [
      {
        "id": "kinow|families2|27720611T1203+12|FAM5",
        "state": "ready",
        "firstParent": {
          "id": "kinow|families2|27720611T1203+12|root",
          "name": "root",
          "cyclePoint": "27720611T1203+12",
          "state": "ready",
          "__typename": "FamilyProxy"
        },
        "__typename": "FamilyProxy"
      },
      {
        "id": "kinow|families2|27720611T1203+12|FAM6",
        "state": "ready",
        "firstParent": {
          "id": "kinow|families2|27720611T1203+12|FAM5",
          "name": "FAM5",
          "cyclePoint": "27720611T1203+12",
          "state": "ready",
          "__typename": "FamilyProxy"
        },
        "__typename": "FamilyProxy"
      },
      {
        "id": "kinow|families2|27700611T1203+12|FAM6",
        "state": "succeeded",
        "firstParent": {
          "id": "kinow|families2|27700611T1203+12|FAM5",
          "name": "FAM5",
          "cyclePoint": "27700611T1203+12",
          "state": "succeeded",
          "__typename": "FamilyProxy"
        },
        "__typename": "FamilyProxy"
      },
      {
        "id": "kinow|families2|27700611T1203+12|FAM5",
        "state": "succeeded",
        "firstParent": {
          "id": "kinow|families2|27700611T1203+12|root",
          "name": "root",
          "cyclePoint": "27700611T1203+12",
          "state": "succeeded",
          "__typename": "FamilyProxy"
        },
        "__typename": "FamilyProxy"
      }
    ],
    "__typename": "Updated"
  },
  "pruned": {
    "__typename": "Pruned"
  },
  "__typename": "Deltas"
}
```
</details>

Pruned is empty, so we ignore that.

Added has one entry, the job.

Now the interesting part. **The updated deltas include the families and task proxies**.

I will start writing some debugging code to capture 100 GraphQL messages, and store so that next time this exception occurs I can take a look at previous messages and confirm I didn't get a `added.cyclePoints`. But from looking at the code, and at the data received, **it looks to me like somehow the backend skipped sending me the `deltas.added.cyclePoint`, and sent the family proxies in the updated key**.

Does it make sense? Do you think this scenario is possible by some rare chance?

Thanks!

p.s.1: deltas.added.cyclePoint is for 

```
    cyclePoints: familyProxies(ids: ["root"], ghosts: true) {
      cyclePoint
    }
```

p.s.2: once families2 is passing with no runtime issues/errors, I still need to test the deltas in the WUI with complex :fearful:

kinow:
Funny, couldn't reproduce it at home, but I noticed my `families2` workflow is different, as it's using integer cycle points.

Let me try to reproduce it again next Monday (@dwsutherland don't bother looking into this, as I know you are busy with migration, and it could be something in my environment/branch).

## Notes to self

1. workflow five at home

```
[scheduling]
    initial cycle point = 20130808T00
#    final cycle point = 20130808T12
    [[dependencies]]
        [[[R1]]]
            graph = "prep => foo"
        [[[PT12H]]]
            graph = "foo[-PT12H] => foo => bar"

[runtime]
  [[root]]
      script="sleep 12"

[visualization]
    initial cycle point = 20130808T00
    final cycle point = 20130808T12
    [[node attributes]]
        foo = "color=red"
        bar = "color=blue"
```

2. workflow families2 at home

```
[scheduling]
  cycling mode = integer
  initial cycle point = 1
  [[dependencies]]
     [[[R/^/P1]]]  # 1, 2, 3, 4, 5, ...
        graph = """foo[-P1] => foo
                 foo => f1 & f2 => bar"""
     [[[R/^/P2]]]  # 1, 3, 5, ...
        graph = "foo => f3 & f4 => bar"

[runtime]
  [[root]]
      script = "sleep $((5 + RANDOM % 10))"
  [[FAM]]
  [[f1, f2, f3, f4]]
     inherit = FAM
```

Executed the complex workflow today too, and after 10 minutes got no errors in the browser console (only warnings like "[Violation] 'message' handler took 197ms'", which indicates there are ApolloClient functions taking too long to execute; I believe these are functions `apply...` which apply the deltas).

Also executed five and families2, but found no error. When I started the tests, I had cylc-uiserver with latest commit `411da1ae685f472ad0983f2c09123a549185ee50`. Syncing repository it had no changes. cylc-flow was on `a3218bd2e43c8bab4eb5ff27157e94004134612c`, and syncing I got `04d5cb3210f7c934d5528d144458f41333ee8669` as last commit (this is from this week I think).

I had also prepared a diff that might be helpful when inspecting deltas:

```diff
diff --git a/src/components/cylc/tree/deltas.js b/src/components/cylc/tree/deltas.js
index 1176836..f8d90b8 100644
--- a/src/components/cylc/tree/deltas.js
+++ b/src/components/cylc/tree/deltas.js
@@ -195,6 +195,8 @@ function handleDeltas (deltas, tree) {
   }
 }
 
+const TEMP_DELTAS = []
+
 /**
  * @param {null|{
  *   id: string,
@@ -228,6 +230,7 @@ export function applyDeltas (deltas, tree) {
       tree.clear()
       return
     }
+    TEMP_DELTAS.push(deltas)
     if (tree.isEmpty()) {
       // When the tree is null, we have two possible scenarios:
       //   1. This means that we will receive our initial data burst in deltas.added.workflow
@@ -246,6 +249,12 @@ export function applyDeltas (deltas, tree) {
       } catch (error) {
         // eslint-disable-next-line no-console
         console.error('Error applying initial data burst for deltas', error)
+        // eslint-disable-next-line no-console
+        console.log('Printing latest 10 deltas...')
+        TEMP_DELTAS
+          .slice(-10)
+          // eslint-disable-next-line no-console
+          .map(delta => console.log(delta))
         throw error
       }
     } else {
@@ -258,6 +267,12 @@ export function applyDeltas (deltas, tree) {
       } catch (error) {
         // eslint-disable-next-line no-console
         console.error('Error applying deltas', error)
+        // eslint-disable-next-line no-console
+        console.log('Printing latest 10 deltas...')
+        TEMP_DELTAS
+          .slice(-10)
+          // eslint-disable-next-line no-console
+          .map(delta => console.log(delta))
         throw error
       }
     }
```

That simply prints the latest 10 deltas (I think, haven't tested the code yet). On Monday I will use this code again from my NIWA laptop, and will report back what were the latest deltas, what was the difference in my `families2` workflow at NIWA. That should clarify whether we could have a bug in the UIS, or if it's something in JS or in my environment.

kinow:
My branches at work are up to date too. The workflow `families2` I have here is different (though no idea how that happened).

```
[scheduling]
initial cycle point = now

[[dependencies]]
[[[P1Y]]]
graph = "FAM3:succeed-all => FAM6"

[runtime]
    [[root]]
        init-script = echo "I'm first"
        env-script = echo "Hi first, I'm second"
        script = sleep 20; echo "RubyRubyRubyRuby"
        exit-script = echo 'Yay!'
        err-script = echo 'Boo!'
    [[FAM]]
    [[FAM2]]
        inherit = FAM
    [[FAM3]]
        inherit = FAM2
    [[foo]]
        inherit = FAM3
    [[bar]]
        inherit = FAM3
    [[FAM4]]
    [[FAM5]]
    [[FAM6]]
        inherit = FAM5
    [[qux]]
        inherit = FAM6
    [[qaz]]
        inherit = FAM4, FAM
    [[qar]]
        inherit = FAM, FAM4, FAM5
```

However, I haven't been able to reproduce this error today after running the workflow in two browsers for ~30 minutes.

kinow:
Spoke too soon, just had that again! Let me apply this patch, and try to capture a sequence of deltas to see if that proves helpful.

kinow:
Error occurred after another 5 minutes running the workflow `families2` above here at NIWA.

image

The 10 deltas in order to the last one, which caused the error: https://gist.github.com/kinow/374dee8fd5f08620c193e43fe3ab5938

Investigating now, as it could be something in the JS code :+1:

kinow:
I repeated this test, now printing also everything that was pruned. I had a theory that maybe the parent wasn't there because I pruned it.

For the job without parent, for its cycle point, nothing was pruned. Meaning it wasn't added, and instead the WUI received the data only in the `updated` key. Will have to wait for an assessment of @dwsutherland here to see if it could be something in the backend :+1:

kinow:
Query used: https://gist.github.com/kinow/a1f99b40d0deb3766e13b0584a84773c

Variables has only the `$workflowId` set to `kinow|families2`.

dwsutherland:
> I repeated this test, now printing also everything that was pruned. I had a theory that maybe the parent wasn't there because I pruned it.
> 
> For the job without parent, for its cycle point, nothing was pruned. Meaning it wasn't added, and instead the WUI received the data only in the `updated` key. Will have to wait for an assessment of @dwsutherland here to see if it could be something in the backend 👍

I think it might be something to do with job deltas being event driven and pool updates collected at the end of a main-loop iteration(?) hmmm... not sure.. How does a job start and/or message back without the taskProxies being created in the data-store.. doesn't seem possible ..

Probably a follow on PR would be fine though

PR #120, opened 2020-02-09
Fixed reload data-store update bug

### LINKED ISSUE #119: `cylc reload` screws up the UIS?
**Describe the bug**

- Run a workflow `foo`, observe it in the web UI.
- run `cylc reload foo`
- the UI still recognizes that `foo` is running, but gets no data to display
- stopping and restarting the UIS fixes the problem (so, it appears to be the UIS not the scheduler that is the problem)

**Release version(s) and/or repository branch(es) affected?**

Current master all components

**Pull requests welcome!**

dwsutherland:
These changes close #119 

**Requirements check-list**

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -127,15 +127,14 @@ class DataStoreMgr:
 
         """
         # wait until data-store is populated for this workflow
-        loop_cnt = 0
-        while loop_cnt < self.INIT_DATA_WAIT_TIME:
-            if w_id not in self.data:
+        if w_id not in self.data:
+            loop_cnt = 0
+            while loop_cnt < self.INIT_DATA_WAIT_TIME:
+                if w_id in self.data:
+                    break
                 sleep(self.INIT_DATA_RETRY_DELAY)
                 loop_cnt += 1
                 continue
-            break
-        if w_id not in self.data:
-            return
         if topic == 'shutdown':
             self.workflows_mgr.stopping.add(w_id)
             self.w_subs[w_id].stop()
@@ -145,7 +144,13 @@ class DataStoreMgr:
         # If the workflow has reloaded recreate the data
         # otherwise apply the delta if it's newer than the previously applied.
         if delta.reloaded:
-            self.data[w_id][topic] = {ele.id: ele for ele in delta.deltas}
+            if topic == WORKFLOW:
+                self.data[w_id][topic].CopyFrom(delta)
+            else:
+                self.data[w_id][topic] = {
+                    ele.id: ele
+                    for ele in delta.deltas
+                }
             self.data[w_id]['delta_times'][topic] = delta_time
         elif delta_time >= self.data[w_id]['delta_times'][topic]:
             apply_delta(topic, delta, self.data[w_id])


cylc/uiserver/data_store_mgr.py
```diff
@@ -144,13 +143,23 @@ def update_workflow_data(self, topic, delta, w_id):
...
-            self.data[w_id][topic] = {ele.id: ele for ele in delta.deltas}
-            self.data[w_id]['delta_times'][topic] = delta_time
-        elif delta_time >= self.data[w_id]['delta_times'][topic]:
-            apply_delta(topic, delta, self.data[w_id])
-            self.data[w_id]['delta_times'][topic] = delta_time
-            self.reconcile_update(topic, delta, w_id)
+        try:
+            if delta.reloaded:
+                if hasattr(delta, 'last_updated'):
+                    self.data[w_id][topic].CopyFrom(delta)
```
kinow: So if `delta.last_updated`, then we just use the protobuf [`CopyFrom`](https://developers.google.com/protocol-buffers/docs/pythontutorial#standard-message-methods).

Do we need to update the comment block above? :point_up:

cylc/uiserver/data_store_mgr.py
```diff
@@ -144,13 +143,23 @@ def update_workflow_data(self, topic, delta, w_id):
...
+                    self.data[w_id][topic] = {
+                        ele.id: ele
+                        for ele in delta.deltas
+                    }
+                self.data[w_id]['delta_times'][topic] = delta_time
+            elif delta_time >= self.data[w_id]['delta_times'][topic]:
+                apply_delta(topic, delta, self.data[w_id])
+                self.data[w_id]['delta_times'][topic] = delta_time
+                self.reconcile_update(topic, delta, w_id)
+        except Exception:
```
kinow: Can/should we catch `KeyError` or some other more specific exception?

dwsutherland: I guess a lot of things could potentially go wrong in the application and reconciliation of the UIS data-store sync (attribute errors should be the most rare, and only happen with API changes).

cylc/uiserver/data_store_mgr.py
dwsutherland: `last_updated` is a attribute of `PbWorkflow`, and the CopyFrom clears the existing data (from the respective item in the store), and merges in the new values..

@hjoliver - "Hold the door" on merging, there may be a more explicit way of finding out what delta it is (don't want to use `topic`, as that is about to change)..

cylc/uiserver/data_store_mgr.py
kinow: Thanks for confirming! Marking as resolved then.

cylc/uiserver/data_store_mgr.py
dwsutherland: Ok good to go now.

cylc/uiserver/data_store_mgr.py
hjoliver: > I guess a lot of things could potentially go wrong in the application and reconciliation of deltas in the UIS data-store sync

The usual practice is to handle specific known exceptions, and crash spectacularly if something new and unknown goes wrong (which is how you identify what else can go wrong that you haven't thought of, rather than hiding the problem).  Is there some reason to do it differently here?

dwsutherland: > > I guess a lot of things could potentially go wrong in the application and reconciliation of deltas in the UIS data-store sync
> 
> The usual practice is to handle specific known exceptions, and crash spectacularly if something new and unknown goes wrong (which is how you identify what else can go wrong that you haven't thought of, rather than hiding the problem). Is there some reason to do it differently here?

The issue is it won't crash the program because it runs in a thread pool (we don't even see the error)..

dwsutherland: @hjoliver - ok removed now.

PR #125, opened 2020-02-26
Use Tornado default WebSocket check_origin function

### LINKED ISSUE #109: Implement same origin check for WebSockets

When I implemented the Tornado handler for WebSockets, I implemented a method of the `WebSocketHandler` to check same origin that returns simply `True` for every request.

This was useful to quickly validate that the client & server interaction worked. We just solved #106 and now I think the WebSocket won't change much.

It could be a good time to investigate how to implement the same origin check. This is implemented normally for JS in browsers, but not WebSockets, which is considered a security risk

>Web browsers allow any site to open a websocket connection to any other, instead of using the same-origin policy that governs other network access from javascript. This can be surprising and is a potential security hole, so since Tornado 4.0 WebSocketHandler requires applications that wish to receive cross-origin websockets to opt in by overriding the check_origin method (see that method’s docs for details). Failure to do so is the most likely cause of 403 errors when making a websocket connection.

The Tornado docs explain why that method was added: https://www.tornadoweb.org/en/stable/websocket.html#tornado.websocket.WebSocketHandler.check_origin

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #109 

When I implemented the WebSocket handler, I used an existing example from Tornado docs, which had the `check_origin` function. As I didn't know how to use it well, I've disabled this check so that it would cause less issues for development. Time to revisit it and refine/review.

The function can be removed from the `SubscriptionHandler`. This will keep the parent `WebSocketHandler`'s [`check_origin`](https://github.com/tornadoweb/tornado/blob/f7d94d0e8a57f51ffc0aac5d55ed7bf9f9936b4d/tornado/websocket.py#L489) default behavior:

- If the client connecting doesn't provide the `Origin` HTTP header, then this function is ignored
- But if the client does provide it (e.g. most web browsers do) then this function is used
- `check_origin` will return `True` meaning that the request passed the test, and `False` otherwise
- The test is simple, it parses the URL from the `Origin` HTTP header, removes protocol, fragments, path, etc, leaving only the domain part. Then compares against the value of the HTTP header `Host`. If the same, then the function will return `True`. Otherwise, it means the connection is coming from another host (e.g. someone could be trying to connect to the WebSocket from mydashboard.somesite.com).

The risk without this, is that CORS/XORS attacks would be possible. Where someone could set up a site somewhere else and try to send malicious messages down the stream to the server. We have other authentication mechanisms in place to prevent that, but it's still recommended to enable this (both by OWASP and by the Tornado docs).

**Requirements check-list**

- [x] I have opened a documentation PR at https://github.com/cylc/cylc-doc/pull/116


kinow:
Tested locally, and it worked with no issues. Can't think of a deployment scenario right now where we would have issues, but if that happens, we should be able to find a workaround. Better than to leave it with no security protection to only add after something bad happens.

Set to 0.3, will mark as ready for review after the 0.2 release :+1:

kinow:
Added docs to the existing PR to `cylc-doc` for WebSocket security, replacing a TODO marker I had left there previously.

Issue #123, opened 2020-02-19
[question] Make Cylc UI Server a Traitlet application?


kinow:
**Describe exactly what you would like to see in an upcoming release**

I had it in my notes from a long time ago, when I first started chopping parts of the JupyterHub single-user application, to get our UI Server working.

One thing that I completely ignored at that time was the whole [Traitlets](https://github.com/ipython/traitlets) code.

Not sure if we need that. In theory the application spawned by the JupyterHub spawners doesn't need to use traitlets, but I never looked if there are any advantages of using it (hence the "question" label).

The only advantages of using it, I **think**, would be a) being able to use Python in the `jupyterhub_config.py` to define variables that could be shared/used in different sections of the file, e.g.

```py
LOG_DATEFMT = '%Y-%m-%dT%H:%M:%S'
c.JupyterHub.log_datefmt = LOG_DATEFMT
c.CylcUIServer.log_datefmt = LOG_DATEFMT
```

and b) using the same tooling as the default application launched by the Hub spawners (i.e. notebook).

But there are probably cons too, like another dependency to the project, we wouldn't have our own configuration file, maybe we want to have more control on how configuration is done, etc.

No strong opinion on this, just remembered about this because yesterday I was reading the JupyterHub code while writing some cookie documentation. 

**Additional context**

Here's one example from Jupyter Core. This is the base class of the `NotebookApp`, which is used by JupyterHub by default with the spawners, and we have replaced with our UI Server.

![image](https://user-images.githubusercontent.com/304786/74869719-5d738e80-53bd-11ea-9374-3134775102f8.png)

It provides the application configuration. The `jupyterhub_config.py` leverages of traitlets to load properties. That's how we have `c.JupyterHub.properties`, `c.ConfigProxy.properties`, and `c.BatchSpawner.properties`, etc.

Traitlets supports things such as defining an object property that will be initialized from the configuration file, with a default value, but that can be overridden by an environment variable.

![image](https://user-images.githubusercontent.com/304786/74870017-dd99f400-53bd-11ea-843f-f75a861f7231.png)

**Pull requests welcome!**

oliver-sanders:
I think this is the right way to go for configuring the hub.

A good trial might be to make the hard-coded scan interval configurable as something like:

```
c.CylcHub.scan_interval = 5  # scan every 5 seconds
```

[...]

PR #127, opened 2020-03-17
Imitate JupyterHub's cookie_secret feature, re-creating a new random secret on every spawn

### LINKED ISSUE #69: Define for how long cookies are persistent
Before I forget, I am looking at the Cylc UI Server now (because I spent some of the most frustrating hours recently troubleshooting a Webpack issue), and found a comment I left while writing the old `cylc-singleuser` prototype.

https://github.com/cylc/cylc-uiserver/blob/ca7e65ed2aab0391e5bd63a2c4409d1c4fc00e72/cylc/uiserver/main.py#L113-L115

From memory, what I recall is that this value could be fixed or randomly generated. If a static value is used, then the cookies survive even after a server restart.

If dynamic (e.g. some random value generated on startup), then the cookie exist for as long as the application is running. It is a parameter that changes a bit security in the UI Server, but I am not sure how important it is.

But just in case, better have this logged in an issue so we don't forget before releasing to production.

Bruno

kinow:
These changes close #69 

Better err on the side of being too cautious I think. Instead of using a static value, this PR sets a random value upon each spawn. JupyterHub's default singleuser app does the same, but also
supports loading it from an external file (using a Traitlets feature, that we can either use too if #123 is done, or re-implement) which can be done later.

**Requirements check-list**

### DIFF
--- a/cylc/uiserver/main.py
+++ b/cylc/uiserver/main.py
@@ -185,9 +185,10 @@ class CylcUIServer(object):
                     {"path": self._static}
                 )
             ],
-            # FIXME: decide (and document) whether cookies will be permanent
-            # after server restart.
-            cookie_secret="cylc-secret-cookie"
+            # always generate a new cookie secret on launch
+            # ensures that each spawn clears any cookies from
+            # previous session, triggering OAuth again
+            cookie_secret=os.urandom(32)
         )
 
     def start(self, debug: bool):


kinow:
Tested locally, and couldn't notice any difference. Restarted the backend and refreshing the tab it showed me the page to launch my UI Server again, and everything worked as expected.

PR #129, opened 2020-03-24
Add mypy to Travis CI and use PEP-484 type hints

kinow:

Not sure if the right time, as there is ongoing work in Cylc UI Server for the deltas. This PR can be closed or postponed for later too.

I think adding MyPy to Cylc Flow and Cylc UI Server would be great. It reduces errors (not adding links here, but PEP-484, Dropbox, Guido himself, and others have links and blogs about it) and also helps if you use IDE or other tools like linters that support types.

For Cylc Flow it would be a major work due to the size of the code base. We can use it later in specific packages there. But for Cylc UI Server, I think we should be able to adopt it sooner as its code base is more recent and smaller.

It uses PEP 484 type hints such as `def fn(param: int, param2: str)`, even though we also have the docstring comments. The reason being that that was the only way I could fix MyPy issues. Looks to me like it doesn't know how to parse the types from docstrings (or purposefully chose to use only type hints?). 

Any thoughts? Tests are passing, did some manual testing with Cylc UI and 2 workflows, and it looks OK. The command in `.travis.yml` shows how to run MyPy. I will comment in a few places MyPy found something interesting too :+1: 

**Requirements check-list**

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -66,7 +66,7 @@ class DataStoreMgr:
         self.w_subs: Dict[Any, Any] = {}
         self.topics = {topic.encode('utf-8') for topic in DELTAS_MAP}
         self.topics.add(b'shutdown')
-        self.loop: asyncio.AbstractEventLoop = asyncio.get_running_loop()
+        self.loop: Any = None
         # Might be options other than threads to achieve
         # non-blocking subscriptions, but this works.
         self.executor = ThreadPoolExecutor()
@@ -83,6 +83,8 @@ class DataStoreMgr:
         blocking the main loop.
 
         """
+        if self.loop is None:
+            self.loop = asyncio.get_running_loop()
         if w_id in self.w_subs:
             return
         self.executor.submit(
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -48,6 +48,8 @@ from cylc.flow.data_store_mgr import (
 )
 from .workflows_mgr import workflow_request
 
+from typing import Any, Dict, List, Optional
+
 logger = logging.getLogger(__name__)
 
 
@@ -58,18 +60,22 @@ class DataStoreMgr:
     INIT_DATA_RETRY_DELAY = 0.5  # seconds
     RECONCILE_TIMEOUT = 5.  # seconds
 
-    def __init__(self, workflows_mgr):
+    def __init__(self, workflows_mgr: Any) -> None:
         self.workflows_mgr = workflows_mgr
-        self.data = {}
-        self.w_subs = {}
+        self.data: Dict[Any, Any] = {}
+        self.w_subs: Dict[Any, Any] = {}
         self.topics = {topic.encode('utf-8') for topic in DELTAS_MAP}
         self.topics.add(b'shutdown')
-        self.loop = None
+        self.loop: asyncio.AbstractEventLoop = asyncio.get_running_loop()
         # Might be options other than threads to achieve
         # non-blocking subscriptions, but this works.
         self.executor = ThreadPoolExecutor()
 
-    async def sync_workflow(self, w_id, *args, **kwargs):
+    async def sync_workflow(
+            self,
+            w_id: str,
+            *args: List[Any],
+            **kwargs: Dict[Any, Any]) -> None:
         """Run data store sync with workflow services.
 
         Subscriptions and sync management is instantiated and run in
@@ -77,8 +83,6 @@ class DataStoreMgr:
         blocking the main loop.
 
         """
-        if self.loop is None:
-            self.loop = asyncio.get_running_loop()
         if w_id in self.w_subs:
             return
         self.executor.submit(
@@ -86,7 +90,7 @@ class DataStoreMgr:
         )
         await self.entire_workflow_update(ids=[w_id])
 
-    def purge_workflow(self, w_id):
+    def purge_workflow(self, w_id: str) -> None:
         """Purge the manager of a workflow's subscription and data."""
         if w_id in self.w_subs:
             self.w_subs[w_id].stop()
@@ -94,7 +98,12 @@ class DataStoreMgr:
         if w_id in self.data:
             del self.data[w_id]
 
-    def start_subscription(self, w_id, reg, host, port):
+    def start_subscription(
+            self,
+            w_id: str,
+            reg: str,
+            host: str,
+            port: int) -> None:
         """Instantiate and run subscriber data-store sync.
 
         Args:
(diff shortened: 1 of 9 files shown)


cylc/uiserver/data_store_mgr.py
```diff
@@ -229,10 +250,10 @@ def reconcile_update(self, topic, delta, w_id):
              'command': req_method,
              'req_context': w_id}
             for w_id, info in self.workflows_mgr.workflows.items())
-        gathers = ()
+        gathers: List[Any] = []
         for kwargs in req_kwargs:
             if not ids or kwargs['req_context'] in ids:
-                gathers += (workflow_request(**kwargs),)
+                gathers.append(workflow_request(**kwargs))
```
kinow: I used a `List` type here as with `Tuple`s you have to specify each possible type within a tuple (i.e. you can have `List[Any]` meaning a list of any type, but `Tuple[Any]` actually means a tuple with one element of any type. If you have multiple values in the tuple, you must define `Tuple[str, int, bool, Any, ...]`, which is harder for these cases where we create it dynamically).

It appears to work the same way?

cylc/uiserver/handlers.py
```diff
@@ -45,45 +49,54 @@ def set_default_headers(self) -> None:
         self.set_header("Content-Type", 'application/json')
 
 
-class MainHandler(HubOAuthenticated, BaseHandler):
+class MainHandler(HubOAuthenticated, BaseHandler):  # type: ignore
```
kinow: The only places where I used the `type: ignore` flag to disable mypy were places where subclassing a type without type hints was necessary, like here where we subclass types from JupyterHub that doesn't have type stubs yet.

cylc/uiserver/main.py
```diff
@@ -42,13 +42,13 @@
...
-    is_closing = False
+    is_closing: bool = False
 
-    def signal_handler(self, signum, frame):
+    def signal_handler(self, signum: Any, frame: Any) -> None:
         logger.info('exiting...')
         self.is_closing = True
 
-    def try_exit(self, uis):
+    def try_exit(self, uis: "CylcUIServer") -> None:
```
kinow: Need to use a string to define the type because `CylcUIServer` is defined further down in this file, and `CylcUIServer` also uses this type (cyclic) so one would need to use this trick of string type.

cylc/uiserver/tests/test_workflows_mgr.py
```diff
@@ -88,7 +88,7 @@
     def side_effect(*_, **__):
         raise socket.error
     mocked_get_host_ip_by_name.side_effect = side_effect
-    reg, host, port, pub_port, client = await est_workflow(
+    reg, host, port, pub_port, client, result = await est_workflow(
```
kinow: The `est_workflow` function could return a tuple with 5 elements, or a tuple with 6 elements. I've changed it to always return a 6-elements tuple, with the last one being `None` when there were only 5...

cylc/uiserver/workflows_mgr.py
```diff
@@ -80,7 +91,7 @@
             if flags.debug:
                 raise
             logger.error("ERROR: %s: %s\n", exc, host)
-            return (reg, host, port, pub_port, None)
+            return (reg, host, port, pub_port, None, None)
```
kinow: Further down, this function returns a tuple with 6 elements instead of 5. So callers of this API need to either check the elements of the tuple, then unpack or iterate it correctly. Now returning always the 6 elements users can ignore that element counting step.

mypy.ini
```diff
@@ -0,0 +1,40 @@
...
+ignore_missing_imports = True
+[mypy-graphql_ws.*]
+ignore_missing_imports = True
+[mypy-graphene.*]
+ignore_missing_imports = True
+[mypy-zmq.*]
+ignore_missing_imports = True
+
+[mypy-cylc.uiserver.tests.*]
+ignore_errors = True
```
kinow: Ignoring tests for now to make this change simpler.

mypy.ini
```diff
@@ -0,0 +1,40 @@
...
+no_implicit_optional = True
+disallow_any_generics = True
+disallow_any_unimported = False
+warn_redundant_casts = True
+warn_unused_ignores = True
+warn_unused_configs = True
+show_traceback = True
+show_error_codes = True
+pretty = True
+python_version = 3.7
```
kinow: Above are the global settings. Then we have a INI group setting style where we can have `[mypy-${MODULE-NAME}]` to apply settings per modules.

Below we ignore the imported modules that apparently do not have types yet.

kinow:
Ah, interesting! Running the mypy command from `.travis.yml` worked fine locally, but failed on Travis.

Found some issues in mypy about differences among versions of mypy, so pinned it down, no success, but still a good idea.

Also found issues about VirtualEnv, venv, and OS specific issues. So deleted my VirtualEnv venv, and created a fresh one with `python -m venv venv`. And now successfully reproduced the failures from Travis.

Quite interesting, looks like there are differences in virtual environments created with VirtualEnv and venv, TIL!

kinow:
Turns out having the Python modules compiled was the real solution. So moving `mypy` to after `pip install .` fixed the issue :+1: ready for review now.

kinow:
>Does this make PEP484 mandatory for all functions/methods?

Yes, it does. And I'm happy to leave this open and discuss/test how it works with the code and documentation.

As we don't have—at the moment—users consuming our code API, the typing would be only for ourselves and to try to reduce a bit possible bugs.

oliver-sanders:
Of all of the Cylc libraries the ui-server or to quote David "code-junction" (I like that analogy) is definitely the best candidate for this.

kinow:
Moved to 0.3

dwsutherland:
Yes, I'd be a proponent of finding some sort of convention to declaring types and organising doc-strings.. I'm always a little lost to know what would be useful, so would need to do some reading to comment further.. If something passes the pros and cons test, then I'm happy to follow the consensus (I just need to learn more!)..

hjoliver:
(This branch is now conflicted)

I'm not sufficiently up with the pros and cons here to decide, might be something to discuss next meeting.

kinow:
> (This branch is now conflicted)

Started fixing conflicts, but unfortunately it's a more-than-30mins job, so leaving it for later.

> I'm not sufficiently up with the pros and cons here to decide, might be something to discuss next meeting.

Low priority, so no hurry on this. For cylc-uiserver, we wouldn't gain much. But cylc-uiserver is one of our simplest python projects, so easier to experiment with. The real gain would be adding it to cylc-flow, but only if other devs are interested in maintaining types.

Adding mypy to cylc-flow won't be easy I think :grimacing: (unless we ignore lots of errors with things like `# type: ignore` or `mypy.ini` exclusions).

kinow:
Might be easier to start from scratch, preferably after discussion with other devs. Adding `mypy` requires every dev adjusting their IDE/editors, and understanding how typing works in the code and in the docs (e.g. pydocs). There might be some cylc-doc work involved to show types in the API using the annotation-mode supported by mypy (I think it needs to be PEP-484 style, but I could be wrong :+1: )

PR #131, opened 2020-04-01
Pass base URL to the JS rendered template

kinow:
These changes partially address https://github.com/cylc/cylc-ui/issues/343

Backend PR, requirement for https://github.com/cylc/cylc-ui/pull/438 (frontend)

**Requirements check-list**
- [x] Does not need tests (why? just passing a new value to the template, if broken the existing tests should fail I think).

### DIFF
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -60,7 +60,9 @@ class MainHandler(HubOAuthenticated, BaseHandler):
         """Render the UI prototype."""
         index = os.path.join(self._static, "index.html")
         user = self.get_current_user()
-        base_url = user['server']
+        protocol = self.request.protocol
+        host = self.request.host
+        base_url = f'{protocol}://{host}/{user["server"]}'
         self.render(index, python_base_url=base_url)
 
 
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -59,7 +59,9 @@ class MainHandler(HubOAuthenticated, BaseHandler):
     def get(self):
         """Render the UI prototype."""
         index = os.path.join(self._static, "index.html")
-        self.render(index)
+        user = self.get_current_user()
+        base_url = user['server']
+        self.render(index, python_base_url=base_url)
 
 
 class UserProfileHandler(HubOAuthenticated, APIHandler):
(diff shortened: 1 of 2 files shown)


cylc/uiserver/handlers.py
```diff
@@ -59,7 +59,11 @@ def initialize(self, path):
     def get(self):
         """Render the UI prototype."""
         index = os.path.join(self._static, "index.html")
-        self.write(open(index).read())
+        user = self.get_current_user()
+        protocol = self.request.protocol
+        host = self.request.host
```
kinow, line 64: In Tornado, `host` includes hostname and port, e.g. `localhost:8000`. With the `protocol`, and the user's server path, we have everything the JS client needs :+1:

kinow:
Summary of change: this PR builds URL like `http://localhost:8000/user/kinow` and pass it over to the template engine. The template engine loads `index.html` replacing the value `python_base_url` by this URL. Then in the linked PR Vue.JS will load this value and use throughout the app, instead of parsing the URL in JS :+1:

cylc/uiserver/tests/test_handlers.py
```diff
@@ -43,7 +43,10 @@ def get_app(self) -> Application:
 
     def test_jupyterhub_version_returned(self):
         with patch.object(MainHandler, 'get_current_user') as mocked:
-            mocked.return_value = {'name': 'yossarian'}
+            mocked.return_value = {
+                'name': 'yossarian',
```
oliver-sanders, line 47: Hey Yoyo.

cylc/uiserver/handlers.py
```diff
@@ -59,7 +59,11 @@ def initialize(self, path):
...
+        base_url = f'{protocol}://{host}/{user["server"]}'
+        self.render(index, python_base_url=base_url)
```
oliver-sanders, line 66: So this is in-effect the redirect to the UI?

What's the functional difference between `self.write` and `self.render`?

(I don't understand enough about Tornado)

kinow: With `.write` you have to replace variables beforehand. I used it because we didn't need to replace anything in the code, no need to "render" anything.

Now with the variable, calling `.render` is simpler, as it does variable replacement for us.

oliver-sanders: Thanks for explaining.

PR #134, opened 2020-05-08
Use recv_nowait for asyncio Queue, to avoid blocking it, and allow the coroutine to exit if the connection was closed

kinow:

The coroutine has a `while True`, the exists once the connection is closed. However, the `queue.get()` call is blocking, and it appears it won't leave the coroutine get to the line of code where it checks the connection status.

This PR uses the `recv_nowait`, combined with `asyncio.sleep(0)`. This seems to work fine for me, tested with the following diff:

```diff
diff --git a/cylc/uiserver/workflows_mgr.py b/cylc/uiserver/workflows_mgr.py
index 48f327f..20d3914 100644
--- a/cylc/uiserver/workflows_mgr.py
+++ b/cylc/uiserver/workflows_mgr.py
@@ -111,6 +111,7 @@ class WorkflowsManager:
 
     async def gather_workflows(self):
         """Scan, establish, and discard workflows."""
+        print(f"We have {len(asyncio.all_tasks())} tasks")
         scanflows = {}
         cre_owner, cre_name = re_compile_filters(None, ['.*'])
         scan_args = (
```

And also looking at `htop` resident memory.

The memory stayed at ~63MB for me, and with no clients, it displayed 1 task. An extra client created 5 new tasks, leaving it at 6 tasks. Another client, and then 11 tasks. Closing one client, back to 6. Closing another tab then 1 task.

Memory stable the whole time. Closing the workflow didn't seem to have affected the memory, but I cannot tell whether it needs a GC, or if it will simply keep some of the old structure (Protobuf, schema, resolvers, etc). But at least this seems to be a fix in the right direction to solve the problem of resident memory?

**Requirements check-list**
- [x] Does not need tests (why? not easy to write a test for this I think).

### DIFF
--- a/cylc/uiserver/websockets/tornado.py
+++ b/cylc/uiserver/websockets/tornado.py
@@ -22,6 +22,8 @@ from typing import Union, Awaitable, Any, List, Tuple, Dict
 
 setup_observable_extension()
 
+NO_MSG_DELAY = 1.0
+
 
 class TornadoConnectionContext(BaseConnectionContext):
     async def receive(self):
@@ -76,7 +78,8 @@ class TornadoSubscriptionServer(BaseSubscriptionServer):
                 task = ensure_future(
                     self.on_message(connection_context, message), loop=self.loop)
                 pending.add(task)
-            await sleep(1)
+            else:
+                await sleep(NO_MSG_DELAY)
 
         self.on_close(connection_context)
         for task in pending:
--- a/cylc/uiserver/websockets/tornado.py
+++ b/cylc/uiserver/websockets/tornado.py
@@ -6,7 +6,8 @@
 
 from inspect import isawaitable
 
-from asyncio import ensure_future, gather, wait, shield
+from asyncio import ensure_future, gather, wait, shield, sleep
+from asyncio.queues import QueueEmpty
 from tornado.websocket import WebSocketClosedError
 from graphql.execution.executors.asyncio import AsyncioExecutor
 from graphql_ws.base import ConnectionClosedException, BaseConnectionContext, BaseSubscriptionServer
@@ -25,7 +26,7 @@ setup_observable_extension()
 class TornadoConnectionContext(BaseConnectionContext):
     async def receive(self):
         try:
-            msg = await self.ws.recv()
+            msg = self.ws.recv_nowait()
             return msg
         except WebSocketClosedError:
             raise ConnectionClosedException()
@@ -58,19 +59,24 @@ class TornadoSubscriptionServer(BaseSubscriptionServer):
         await self.on_open(connection_context)
         pending = set()
         while True:
+            message = None
             try:
                 if connection_context.closed:
                     raise ConnectionClosedException()
                 message = await connection_context.receive()
             except ConnectionClosedException:
                 break
+            except QueueEmpty:
+                pass
             finally:
                 if pending:
                     (_, pending) = await wait(pending, timeout=0, loop=self.loop)
 
-            task = ensure_future(
-                self.on_message(connection_context, message), loop=self.loop)
-            pending.add(task)
+            if message:
+                task = ensure_future(
+                    self.on_message(connection_context, message), loop=self.loop)
+                pending.add(task)
+            await sleep(1)
 
         self.on_close(connection_context)
         for task in pending:
--- a/cylc/uiserver/websockets/tornado.py
+++ b/cylc/uiserver/websockets/tornado.py
@@ -6,7 +6,7 @@
 
 from inspect import isawaitable
 
-from asyncio import ensure_future, gather, wait, shield, sleep
+from asyncio import create_task, gather, wait, shield, sleep
 from asyncio.queues import QueueEmpty
 from tornado.websocket import WebSocketClosedError
 from graphql.execution.executors.asyncio import AsyncioExecutor
@@ -74,8 +74,8 @@ class TornadoSubscriptionServer(BaseSubscriptionServer):
                     (_, pending) = await wait(pending, timeout=0, loop=self.loop)
 
             if message:
-                task = ensure_future(
-                    self.on_message(connection_context, message), loop=self.loop)
+                task = create_task(
+                    self.on_message(connection_context, message))
                 pending.add(task)
             else:
                 await sleep(NO_MSG_DELAY)
--- a/cylc/uiserver/websockets/tornado.py
+++ b/cylc/uiserver/websockets/tornado.py
@@ -26,10 +26,9 @@ NO_MSG_DELAY = 1.0
 
 
 class TornadoConnectionContext(BaseConnectionContext):
-    async def receive(self):
+    def receive(self):
         try:
-            msg = self.ws.recv_nowait()
-            return msg
+            return self.ws.recv_nowait()
         except WebSocketClosedError:
             raise ConnectionClosedException()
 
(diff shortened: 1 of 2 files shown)


cylc/uiserver/websockets/tornado.py
```diff
@@ -58,19 +59,24 @@ async def _handle(self, ws, request_context):
...
                     (_, pending) = await wait(pending, timeout=0, loop=self.loop)
 
-            task = ensure_future(
-                self.on_message(connection_context, message), loop=self.loop)
-            pending.add(task)
+            if message:
+                task = ensure_future(
+                    self.on_message(connection_context, message), loop=self.loop)
+                pending.add(task)
+            await sleep(1)
```
dwsutherland, line 79: Could probably only wait if no message was received:
```
[I 2020-05-10T09:38:48.469 JupyterHub log:174] 200 GET /hub/error/503?url=%2Fuser%2Fsutherlander%2Ffavicon.png (@127.0.0.1) 3.47ms
We have 6 tasks
We have 6 tasks
We have 6 tasks
09:39:15.051 [ConfigProxy] info: 200 GET /api/routes 
[I 2020-05-10T09:39:15.073 JupyterHub proxy:320] Checking routes
We have 6 tasks
We have 6 tasks
We have 8 tasks
We have 6 tasks
We have 6 tasks
We have 6 tasks
gen exit
gen exit
We have 8 tasks
We have 8 tasks
We have 8 tasks
We have 10 tasks
We have 8 tasks
We have 8 tasks
We have 8 tasks
We have 8 tasks
We have 8 tasks
We have 8 tasks
gen exit
We have 3 tasks
We have 1 tasks
We have 1 tasks
We have 1 tasks
We have 1 tasks
We have 1 tasks
We have 1 tasks
We have 1 tasks
We have 1 tasks
^C[C 2020-05-10T09:43:24.323 JupyterHub app:2684] Received signal SIGINT, initiating shutdown...
[I 2020-05-10T09:43:24.324 JupyterHub app:2360] Cleaning up single-user servers...
[I 2020-05-10T09:43:24.325 JupyterHub proxy:706] Cleaning up proxy[14716]...
09:43:24.325 [ConfigProxy] warn: Terminated
2020-05-10 09:43:24,326 cylc.uiserver.main INFO     exiting...
2020-05-10 09:43:24,335 cylc.uiserver.main INFO     exit success
[I 2020-05-10T09:43:26.171 JupyterHub app:2392] ...done
(proto-uiserver) sutherlander@cortex-vbox:cylc-uiserver$ jupyterhub
```

kinow: >Could probably only wait if no message was received: 

Great idea! Diff applied (thanks!)

>Is 1.0sec delay Ok? shorter? longer?

Good question. My first try had `sleep(0)`. I think one blog post I was looking at, with examples for non-blocking asyncio queue, had the sleep(1) so I went with that.

Just had a look at CPython, but looks like it simply [waits on a future](https://github.com/python/cpython/blob/2c3d508c5fabe40dac848fb9ae558069f0576879/Lib/asyncio/queues.py#L166), using CPython's interval default interval (couldn't find the source for that :).

Any thoughts on `0`, `1`, or greater anybody? Happy to update it.

kinow:
>Approve in general (solves accumulating coroutine issue) tada

Hooray!

>However, the accumulating memory issue appears to remain (perhaps the coroutine return information is not being deleted?)

Hmm. Let me check that (will use @hjoliver's new documentation on profiling Cylc 8 to plot the memory tomorrow morning)

cylc/uiserver/websockets/tornado.py
dwsutherland: This fine I think, because the looping over our async gen (iterator) happens in a different loop;
https://github.com/cylc/cylc-uiserver/blob/14e9570aa0b7823ce89c93f9a822a8925a70cb01/cylc/uiserver/websockets/tornado.py#L191-L207
so 1sec (average of 0.5sec) to setup a new message on that connection.

dwsutherland:
> > Approve in general (solves accumulating coroutine issue) tada
> 
> Hooray!
> 
> > However, the accumulating memory issue appears to remain (perhaps the coroutine return information is not being deleted?)
> 
> Hmm. Let me check that (will use @hjoliver's new documentation on profiling Cylc 8 to plot the memory tomorrow morning)
> 
> Thanks for the great feedback @dwsutherland !


The memory profile is flat without subscriptions/WUI connected...
So we are in the right place, but need to figure out where this data isn't being disposed.. Maybe the ConnectionContext, or AsyncExecutor in graphql (or both), or elsewhere.

kinow:
Here's my profile output:

memory-profile

cylc/uiserver/websockets/tornado.py
oliver-sanders: (I tested with a 0 second sleep and didn't see any CPU consequences)

cylc/uiserver/websockets/tornado.py
```diff
@@ -21,11 +22,13 @@
...
+
 
 class TornadoConnectionContext(BaseConnectionContext):
     async def receive(self):
         try:
-            msg = await self.ws.recv()
+            msg = self.ws.recv_nowait()
             return msg
         except WebSocketClosedError:
             raise ConnectionClosedException()
```
hjoliver, line 34: Shouldn't this not be a co-routine anymore? (It is now 'await'-less)

hjoliver: (I guess it doesn't matter, is kind of pointless?)

cylc/uiserver/websockets/tornado.py
```diff
@@ -58,19 +61,25 @@ async def _handle(self, ws, request_context):
         await self.on_open(connection_context)
         pending = set()
         while True:
+            message = None
             try:
                 if connection_context.closed:
                     raise ConnectionClosedException()
                 message = await connection_context.receive()
```
hjoliver, line 68: (So the `await` here is pointless?)

cylc/uiserver/websockets/tornado.py
kinow: Oh, good point! It looks like it doesn't need to be `async` any more (or at least I don't see any reason). The function in the parent class is `def receive`, and implementations may use `async` if necessary I think.

cylc/uiserver/websockets/tornado.py
kinow: :+1: removed `await`

cylc/uiserver/websockets/tornado.py
kinow: Done! :+1:  thanks!

kinow:
(Did a bit of testing attaching the debugger and looking at `asyncio.all_tasks()`, everything looks OK after latest change to remove `async` from the function.)

cylc/uiserver/websockets/tornado.py
```diff
@@ -58,19 +60,25 @@ async def _handle(self, ws, request_context):
...
             finally:
                 if pending:
                     (_, pending) = await wait(pending, timeout=0, loop=self.loop)
 
-            task = ensure_future(
-                self.on_message(connection_context, message), loop=self.loop)
-            pending.add(task)
+            if message:
+                task = ensure_future(
+                    self.on_message(connection_context, message), loop=self.loop)
```
hjoliver, line 78: Should this be changed to use `create_task`? (see Python docs: https://docs.python.org/3/library/asyncio-task.html#creating-tasks)

hjoliver: (kind of off-topic, sorry, but I was trying to understand what this does :grin: )

kinow: >Important
>
>See also the create_task() function which is the preferred way for creating new Tasks. 

Doesn't hurt to change to the latest recommended way. Let me change it and quickly test it. :+1:

kinow: Done!

PR #138, opened 2020-05-14
Remove Tornado from server header

kinow:

Harmless little change. Was looking at requests/responses in Firefox today while playing with deltas, and noticed the Server signature in the response headers.

So to prevent [server fingerprinting](https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/01-Information_Gathering/02-Fingerprint_Web_Server) we just omit the value.

I think in real production environments there may be other tools like reverse proxies that could hide the header too. But doesn't hurt to remove it.

While attackers could still guess we are using Tornado (if targetting a host with Cylc, and aware that we use Tornado), at least it won't give away the version installed in the server (the current header says Tornado/$version).

**Requirements check-list**

### DIFF
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -36,6 +36,7 @@ class BaseHandler(web.RequestHandler):
         self.set_header("Access-Control-Allow-Origin", "*")
         self.set_header("Access-Control-Allow-Headers", "x-requested-with")
         self.set_header('Access-Control-Allow-Methods', 'POST, GET, OPTIONS')
+        self.set_header('Server', '')
 
 
 class APIHandler(BaseHandler):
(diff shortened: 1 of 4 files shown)


dwsutherland:
@kinow - Is it not meant to show `server: TornadoServer/6.0.4`? Still does for me:
image

kinow:
Uh, funny. Could it be browser cache? What happens if you open in a different browser @dwsutherland ?

kinow:
From Riot, @dwsutherland found that not all routes were protected.

- [x] BaseHandler sets the header
- [ ] Some handlers are not extending the BaseHandler, so they keep sending the server header
- [ ] Tornado's default static handler appears to also leak the server info
- [ ] JupyterHub as well :point_up:

kinow:
Only the `oauth_callback` call from JH appears to insist on using the Tornado default header. Other handlers should be returning an empty string, including JH handlers.

kinow:
Hm. Tests passing locally for me. Error in CI doesn't seem related to this change?

PR #145, opened 2020-07-15
Next development version 0.3.dev

kinow:

For the next release. I used 0.3.dev, but happy to change it to another value if others prefer. For me that's useful as I keep switching branches and venvs, and doing `pip list` when using a version like this, it's easier to confirm you are using the version you expect to, not one accidentally installed from PYPI (though there's also the link location in the output, example below)

image

Furthermore, if a user - by any chance - is using the version from the `master` branch, the version could tell us that s/he is doing so, and we could then ask to install from tag/PYPI/Conda/etc, or work around any issues in the latest version - I think :o) 

**Requirements check-list**

### DIFF
--- a/cylc/uiserver/__init__.py
+++ b/cylc/uiserver/__init__.py
@@ -13,4 +13,4 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-__version__ = "0.2"
+__version__ = "0.3.dev"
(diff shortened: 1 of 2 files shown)


kinow:
Rebased

hjoliver:
Seems sensible; we should probably do this for cylc-flow too...

kinow:
> Seems sensible; we should probably do this for cylc-flow too...

And Cylc UI as well.

kinow:
Rebased

kinow:
The new 0.3 changelog entry has been added in another PR already :+1:

hjoliver:
(one review will do)

PR #156, opened 2020-09-14
workflow stopped delta fix

dwsutherland:

The final/contact delta was not using the correct object, so the `stopped` message was not getting sent to the WUI.

This PR fixes this issue:
```
{
  "data": {
    "deltas": {
      "id": "sutherlander|nox",
      "shutdown": false,
      "added": {
        "workflow": {
          "id": "sutherlander|nox"
        }
      },
      "updated": {
        "workflow": {
          "id": "sutherlander|nox",
          "status": "stopped"
        }
      }
    }
  }
}
```

One review should do..

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -72,9 +72,9 @@ class DataStoreMgr:
         self.delta_queues = {}
 
     def update_contact(self, w_id, contact_data=None):
-        delta = DELTAS_MAP[WORKFLOW]()
-        delta.time = time.time()
-        flow = delta.updated
+        delta = DELTAS_MAP[ALL_DELTAS]()
+        delta.workflow.time = time.time()
+        flow = delta.workflow.updated
         flow.id = w_id
         if contact_data:
             # update with contact file data
@@ -94,10 +94,9 @@ class DataStoreMgr:
             flow.status = SuiteStatus.STOPPED.value
 
         # Apply to existing workflow data
-        apply_delta(WORKFLOW, delta, self.data[w_id])
         if 'delta_times' not in self.data[w_id]:
-            self.data[w_id]['delta_times'] = {}
-        self.data[w_id]['delta_times'][WORKFLOW] = delta.time
+            self.data[w_id]['delta_times'] = {WORKFLOW: 0.0}
+        self.apply_all_delta(w_id, delta)
         # Queue delta for subscription push
         self.delta_store_to_queues(w_id, ALL_DELTAS, delta)
 
@@ -209,6 +208,11 @@ class DataStoreMgr:
             self.delta_store_to_queues(w_id, topic, delta)
             self.workflows_mgr.stopping.add(w_id)
             return
+        self.apply_all_delta(w_id, delta)
+        self.delta_store_to_queues(w_id, topic, delta)
+
+    def apply_all_delta(self, w_id, delta):
+        """Apply the AllDeltas delta."""
         for field, sub_delta in delta.ListFields():
             delta_time = getattr(sub_delta, 'time', 0.0)
             # If the workflow has reloaded clear the data before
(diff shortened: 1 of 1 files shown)


kinow (approved):
Worked fine @dwsutherland 

Are we able to use the new integration or functional test code to write a test for this? If it's simple, it'd be good, otherwise I'd be fine with the change as-is :+1: 

Note: we didn't have the workflow data in the updated delta @dwsutherland :sweat_smile: I think you had tested it with GraphiQL or the command line?

After adding the workflow data locally to `deltas.updated` in the Cylc UI, got the final message with the `stopped` status.

image

dwsutherland:
> Note: we didn't have the workflow data in the updated delta @dwsutherland 😅 I think you had tested it with GraphiQL or the command line?

Not sure I understand; we can put the `stopped` status anywhere we like, I chose the updated delta, and tested it with subscriptions in GraphiQL.

> Are we able to use the new integration or functional test code to write a test for this? If it's simple, it'd be good, otherwise I'd be fine with the change as-is

Yes, we need to test this aspect of a shutdown suite picked up by the scan.. But it would take some work, probably should have been done when the feature was added ... But that was my bad, as the changes are mostly my suggestions.

kinow:
>Not sure I understand; we can put the stopped status anywhere we like, I chose the updated delta, and tested it with subscriptions in GraphiQL.

Oh, no, that's not what I meant.

In Cylc UI, the `subscription` query had a `deltas.added.workflows`, but did not have a `deltas.updated.workflows`. I'm assuming that even if we had, it wouldn't be working as expected without this fix.

cylc/uiserver/data_store_mgr.py
```diff
@@ -72,9 +72,9 @@ def __init__(self, workflows_mgr):
         self.delta_queues = {}
 
     def update_contact(self, w_id, contact_data=None):
-        delta = DELTAS_MAP[WORKFLOW]()
-        delta.time = time.time()
-        flow = delta.updated
+        delta = DELTAS_MAP[ALL_DELTAS]()
+        delta.workflow.time = time.time()
+        flow = delta.workflow.updated
```
oliver-sanders, line 77: Trying to create a delta for a workflow but we can't use `DELTAS_MAP[WORKFLOW]`?

cylc/uiserver/data_store_mgr.py
```diff
@@ -94,10 +94,12 @@ def update_contact(self, w_id, contact_data=None):
...
 
         # Apply to existing workflow data
-        apply_delta(WORKFLOW, delta, self.data[w_id])
         if 'delta_times' not in self.data[w_id]:
-            self.data[w_id]['delta_times'] = {}
-        self.data[w_id]['delta_times'][WORKFLOW] = delta.time
+            self.data[w_id]['delta_times'] = {
+                key: 0.0
+                for key in DATA_TEMPLATE
+            }
```
oliver-sanders, line 101: Set all delta times to `0.0`, unless there is a delta time in `self.data` in which case do nothing?

oliver-sanders (commented):
Errr, ok, I really don't understand what's going on with these changes, could you help me understand, I need to get my head around this.

cylc/uiserver/data_store_mgr.py
dwsutherland: Yeah might only need WORKFLOW key created.

cylc/uiserver/data_store_mgr.py
dwsutherland: `DELTAS_MAP[WORKFLOW]` is the individual workflow delta, but we create the `delta_store` out of the `AllDeltas` message using .. It doesn't really matter, because it will only contain the workflow data we set anyway..

https://github.com/cylc/cylc-flow/blob/eab1b1c8fea70b982b19bde6cce5570165bd74dc/cylc/flow/data_store_mgr.py#L199-L242

cylc/uiserver/data_store_mgr.py
dwsutherland: Done.

oliver-sanders (approved):

PR #160, opened 2020-10-12
New tests for the data store manager

kinow:
This is a small change with no associated Issue.

Just adding a few more tests for Cylc UI Server :+1: 

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Appropriate tests are included (unit and/or functional).
- [x] No change log entry required (why? e.g. invisible to users).
- [x] No documentation update required.
- [x] No dependency changes.

PR #161, opened 2020-10-15
Add more tests for workflow_mgr

kinow:

Similar to #160 , but adding more tests for `workflow_mgr`. Still trying to understand how its methods work, quite interesting. Found already one pydoc error. So if I cannot understand how to write tests, at least this PR will fix that pydoc :sweat_smile: 

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/tests/conftest.py
+++ b/cylc/uiserver/tests/conftest.py
@@ -45,6 +45,9 @@ class AsyncClientFixture(ZMQSocketBase):
             raise self.returns
         return self.returns
 
+    def stop(self, *args, **kwargs):
+        pass
+
 
 @pytest.fixture
 def async_client():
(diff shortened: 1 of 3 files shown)


kinow:
I think this was more of an exercise to understand the state changes in the UI Server. When writing tests, I normally focus on the `public` API, which might have played against me here (still influenced by Java :grimacing: ).

To test the private functions, I tried to trigger the right state changes in `.update`, which will call `_workflow_state_changes` to compute the next states based on the active, inactive, and what came out of the "pipes".

I tried using fixtures, and avoiding duplication as much as possible, though there might be some better way of writing these tests and reduce even more the amount of code written for tests.

Happy to take any feedback and improve it. Should improve a bit our coverage prior to the next release :+1:

oliver-sanders (approved):
Nice.

dwsutherland (approved):
Looks good, tests passing 👍

PR #167, opened 2020-12-13
Use JupyterHub 1.3.* and Tornado 6.1.*. Auto spawn UI Server (new JHub feature)

### LINKED ISSUE #66: Investigate if it is possible to control the redirect URL in JupyterHub
In JupyterHub, when you request a URL like `http://localhost:8000/user/kinow/#/workflows/five`, but the hub notices that your UI Server is down, it will redirect you to a URL with the button to "Launch Server".

Screenshot_2019-08-19_15-14-02

This button will initiate the process to start the Cylc UI Server, and upon success, will redirect it to a URL. It is possible to see that the URL to be redirected is passed as a URL query parameter, e.g. `http://localhost:8000/hub/spawn/kinow?next=%2Fhub%2Fuser%2Fkinow%2F`

Which decodes to `next=/hub/user/kinow/`, and then gets redirected later to `/user/kinow/` internally by proxy+hub.

It can be annoying to users - or at least it is annoying to devs working on views in Vue.js - to be on a page, re-launch the hub, just to be taken back to the initial page... having to go through all the work to find which workflow you were running, and find your URL again (the browser history for my case doesn't always contain the previous URL... maybe because of [our Vue.js app's route configured to use hash mode](https://router.vuejs.org/guide/essentials/history-mode.html)).

It would be nice if instead it redirected to your previous URL.

image

Cheers
Bruno

### LINKED ISSUE #163: Check if we need/want/can update to Tornado 6.1

If the next release of JupyterHub does that, we will automatically need to follow and update our dependency version, and check if our handlers or custom WebSocket code needs any adjustments.

We also need to consider the version range in JupyterHub (I think they are very general in their requirements, so probably not an issue). If it's a 15 minutes task, I think it should be OK to update, and if there's no risk of incompatibilities with JupyterHub.

Otherwise we can just leave it for Cylc 8.0 final release, or 8.1, etc.

**Additional context**

https://www.tornadoweb.org/en/stable/releases/v6.1.0.html

And notes I posted on Element chat:

```

    This is the last release of Tornado to support Python 3.5. Future versions will require Python 3.6 or newer. (i.e. if we update Cylc UIS for 6.1, then Cylc 8 will require Python 3.6 or newer, which is fine as I believe we are aiming at 3.7 for now)
    Binary wheels are now provided for Windows, MacOS, and Linux (amd64 and arm64). (nice! it should improve build time, as it won't need to run pip install and re-build Tornado from sources [can't remember if the same applies for Conda env])
    ssl.CertificateError during the SSL handshake is now handled correctly.
    Reads that are resolved while the stream is closing are now handled correctly.
    The ping_interval timer is now stopped when the connection is closed. (tornado.websocket; quite sure it won't affect, as we are not using the ping_interval, but I used it for testing/development; need to double check it won't affect us)
    websocket_connect now raises an error when it encounters a redirect instead of hanging. (no idea if we ever had this issue)

One issue I wanted to see soon was HTTP/2 support, but it's still experimental, in a forked repo of the maintainer of tornado, far away of being usable (and HTTP/3 is now a draft at the IETF as of oct/2020, and last year there was something about websockets + http/2 but I never got a chance to experiment with that to see what advantages it has over http/1+websockets).

```

**Pull requests welcome!**

kinow:
These changes close #66 
These changes close #163 

Our fix for #66 was included in 1.1 but I missed that. Via Element (hurrah to Gitter's integration) I saw today's announcement in the JupyterHub room about 1.3.0.

This PR upgrades JupyterHub to 1.3.0. Manual testing found nothing wrong. ~And we might also be able to automatic spawn the Cylc UI Server now. I think this feature was either in 1.1 or 1.2 of JupyterHub. So we shouldn't need to click that button to launch the UI Server any longer after restarting JupyterHub and accessing it for the first time (can't recall if we need to enable a flag in JupyterHub configuration file).~ Changed the JupyterHub configuration file that we provide as demo to include the auto spawn of 1 second (effectively enabling that setting, which is set to `0` by default, forcing us to click to launch the UI Server).

Also upgraded to Tornado 6.1, fixing #163. 

**Requirements check-list**
- [x] Created an issue at [cylc-uiserver conda-forge repository](https://github.com/conda-forge/cylc-uiserver-feedstock) with version changes (if you changed dependencies in `setup.py`, see `recipe/meta.yaml`). - https://github.com/conda-forge/cylc-uiserver-feedstock/pull/11

### DIFF
--- a/jupyterhub_config.py
+++ b/jupyterhub_config.py
@@ -41,6 +41,7 @@ c.Spawner.cmd = ['cylc-uiserver']
 #  Should be a subclass of Spawner.
 c.JupyterHub.spawner_class = 'jupyterhub.spawner.LocalProcessSpawner'
 
+c.JupyterHub.implicit_spawn_seconds = 0.01
 
 # --- Cylc-ise Jupyterhub
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,10 @@ None or N/A.
 
 ### Enhancements
 
+[#167](https://github.com/cylc/cylc-uiserver/pull/167) - Upgrade
+JupyterHub to 1.3.x, and Tornado to 6.1.x. Set auto spawn timeout
+to 1 second (effectively enabling it) in our demo configuration.
+
 [#125](https://github.com/cylc/cylc-uiserver/pull/125) - Use Tornado
 default WebSocket check_origin function.
 [#124](https://github.com/cylc/cylc-uiserver/pull/124) - Add decorator
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ def find_version(*file_paths):
 install_requires = [
     ('cylc-flow @ https://github.com/cylc/cylc-flow'
      '/tarball/master#egg=cylc-8.0a3.dev'),
-    'jupyterhub==1.1.*',
+    'jupyterhub==1.3.*',
     'tornado==6.0.*',
     'graphene-tornado==2.6.*',
     'graphql-ws>=0.3.1,<0.4'
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ install_requires = [
     ('cylc-flow @ https://github.com/cylc/cylc-flow'
      '/tarball/master#egg=cylc-8.0a3.dev'),
     'jupyterhub==1.3.*',
-    'tornado==6.0.*',
+    'tornado==6.1.*',
     'graphene-tornado==2.6.*',
     'graphql-ws>=0.3.1,<0.4'
 ]


kinow:
Looking if JupyterHub 1.3 works with the latest Tornado, for #163

kinow:
The setting to spawn the UI Server automatically is `# c.JupyterHub.implicit_spawn_seconds = 0`. You can read more about it at https://jupyterhub.readthedocs.io/en/stable/reference/config-reference.html

```
## Trigger implicit spawns after this many seconds.
#  
#  When a user visits a URL for a server that's not running, they are shown a
#  page indicating that the requested server is not running with a button to
#  spawn the server.
#  
#  Setting this to a positive value will redirect the user after this many
#  seconds, effectively clicking this button automatically for the users,
#  automatically beginning the spawn process.
#  
#  Warning: this can result in errors and surprising behavior when sharing access
#  URLs to actual servers, since the wrong server is likely to be started.
#  Default: 0
# c.JupyterHub.implicit_spawn_seconds = 0
```

The default value, `0`, appears to keep it disabled. So I've changed it to `1`. That's useful, at least for development I think?

kinow:
And no issues with Tornado 6.1. It shouldn't cause any issues for JupyterHub as they haven't changed their range of `tornado >5.1`.

oliver-sanders:
Looking at the JupyterHub docs for `implicit_spawn_seconds`:

```
#  Warning: this can result in errors and surprising behavior when sharing access
#  URLs to actual servers, since the wrong server is likely to be started.
```

Not quite sure what this means.

jupyterhub_config.py
```diff
@@ -41,6 +41,7 @@
 #  Should be a subclass of Spawner.
 c.JupyterHub.spawner_class = 'jupyterhub.spawner.LocalProcessSpawner'
 
+c.JupyterHub.implicit_spawn_seconds = 1
```
oliver-sanders, line 44: The docs say any positive value so I guess we could go with `0.01` or something `<1` to reduce lag.

oliver-sanders (approved):
Tested automatic spawning, worked first time no problems, very nice.

jupyterhub_config.py
kinow: Good idea. Done. And tested.

hjoliver (approved):
Agreed, nice :+1:

Issue #111, opened 2020-01-22
Cylc Hub configuration


oliver-sanders:
At the moment we have the jupyterhub configuration defined in a single python file within the repo.

We intend to have all Cylc configuration files contained within the same tree for ease of management (https://github.com/cylc/cylc-flow/issues/3170) something along the lines of this:

* `<configuration_root_dir>/`
    * `flow/`
       * `8.0.0/`
           * `flow.rc`
       * `8.0.1/`
           * `flow.rc`
    * `hub/` (or `ui-server`?)
       * `1.0.0/`
           * `ui_server.py`

How do we want to move forward with the configuration options:

* [ ] Do nothing:
  * Give users a template jupyterhub config file.
  * leave it up to them to put this file somewhere sensible.
  * Leave it up to them to specify the path to their own jupyterhub config file.
* [x] Create a `cylc hub` command which:
  * Is a thin wrapper to jupyterhub which sets the config path:
      * Adds the `jupytefhub_config.py` file from this repo (which sets the logo etc).
      * Adds user/site config from the Cylc configuration tree.

**Pull requests welcome!**

oliver-sanders:
Suggest creating a quick simple `cylc hub` command that wraps Jupyterhub:
* Easy to do with the `cylc.command` entry point.
* Should ensure jupyterhub runs with the Cylc configuration.
* Puts `jupyterhub` behind a `cylc` command (will appear to `ps` as such) which allows command whitelisting.
* Should also load the user config `~/.cylc/ui-server/[<version>/]ui_server.py` (after the base Cylc config).
* Should still work when launched manually (providing you specify the base config file or are in the `cylc-uiserver` directory), you just loose the user config file stuff.
* Add a support for a site config later.

oliver-sanders:
Added a `cylc hub` command in #173 along with a user config.

We now have the ability to define the configuration hierarchy how we want it, leaving this open as a question until we discuss & agree how we would like this to work.

oliver-sanders:
superseded

Issue #80, opened 2019-09-19
Revisit jupyterhub_config location for Cylc UI


kinow:
Developers may have different directory structures for their projects. Some checking out the code in a common directory like `~/code/`, others using different parent directories (e.g. `~/javascript/`, `~/python/`), or even using multiple git working trees.

Cylc UI Server has a `jupyterhub_config.py` file used for development. That file needs a location of the static files, produced by Cylc UI. And at the moment it is expecting the developer to have both projects sharing a root directory.

We need to find another solution that accommodates these possible developer environment configurations.

kinow:
@oliver-sanders can we close this one after #173?

oliver-sanders:
Oh yes.

PR #173, opened 2021-02-18
use cylc CLI and config dir

### LINKED ISSUE #123: [question] Make Cylc UI Server a Traitlet application?

oliver-sanders:
closes #123 

* put commands behind the `cylc` console script
* source the base config by default
* source the user config from the .cylc dir when present

The `cylc hub` script just calls Jupyterhub sourcing our configuration module which loads the default and user configuration files. This command can be called from anywhere on the filesystem and appears in the `cylc help` output:

```console
$ cylc help all
broadcast ............. Override [runtime] configuation items in a running suite.
cat-log ............... View Cylc suite and job log files.
check-versions ........ Check that Cylc versions match on different platforms.
...
hub ................... Launch the Cylc hub for running the Cylc Web GUI.
...
```

I've added a preliminary user config location for now, will need to add a site configuration too, will raise the config locations (#111) at the next meetup.

closes #161 
addresses #111 (full solution requires resolution of outstanding questions)
addresses #80 (the dependence on having cylc-ui and cylc-uiserver in the same dir remains for now)

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,29 +19,33 @@ Workflow Services (WFS).
 
 ## Installation
 
-To install the production version, run:
+For production:
 
-- `pip install cylc-uiserver`
+```console
+$ pip install cylc-uiserver
+```
 
-And for the development version, run the following from a clone of the project
-git repository.
+For development run the following from a clone of the project git repository:
 
-- `pip install -e .`
+```console
+$ pip install -e .
+```
 
-Once that is done, the Hub can be started with `jupyterhub`.
+## Running
+
+```console
+$ cylc hub
+```
+
 The default URL is [http://localhost:8000](http://localhost:8000).
 
-### Starting multiple instances of jupyterhub
+# Configuring
 
-If you need to run multiple instances of `jupyterhub`, you will have to follow
-the steps described below:
+The default "base" configuration is defined in
+`cylc.uiserver.config_defaults.py`, these values can be overridden by the
+user config in `~/.cylc/hub/config.py`.
 
-- copy `jupyterhub_config.py` to a different location
-- change the following settings (change the ports to your environment):
-    * `c.JupyterHub.bind_url = 'http://:7000'`
-    * `c.JupyterHub.hub_bind_url = 'http://127.0.0.1:7878'`
-    * `c.JupyterHub.proxy_api_port = 9001`
-- finally start `jupyterhub` using that configuration file
+See the Jupyterhub documentation for details on configuration options.
 
 ## Copyright and Terms of Use
 
--- a/cylc/uiserver/config_defaults.py
+++ b/cylc/uiserver/config_defaults.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Copyright (C) NIWA & British Crown (Met Office) & Contributors.
 #
 # This program is free software: you can redistribute it and/or modify
@@ -19,8 +18,10 @@
 from pathlib import Path
 import pkg_resources
 
-HERE = Path(__file__).resolve().parent
-DIST = HERE.joinpath(Path('../cylc-ui/dist'))
+from cylc.uiserver import __file__ as uis_pkg
+
+
+DIST = Path(uis_pkg).parents[3] / 'cylc-ui/dist'
 
 
 # --- Extra arguments to be passed to the single-user server.
@@ -33,7 +34,7 @@ c.Spawner.args = ['-s', str(DIST)]
 #  Some spawners allow shell-style expansion here, allowing you to use
 #  environment variables. Most, including the default, do not. Consult the
 #  documentation for your spawner to verify!
-c.Spawner.cmd = ['cylc-uiserver']
+c.Spawner.cmd = ['cylc', 'uiserver']
 
 
 # --- The class to use for spawning single-user servers.
--- a/cylc/uiserver/config_defaults.py
+++ b/cylc/uiserver/config_defaults.py
@@ -21,48 +21,22 @@ import pkg_resources
 from cylc.uiserver import __file__ as uis_pkg
 
 
-DIST = Path(uis_pkg).parents[3] / 'cylc-ui/dist'
-
-
-# --- Extra arguments to be passed to the single-user server.
-
-#  Some spawners allow shell-style expansion here, allowing you to use
-#  environment variables here. Most, including the default, do not. Consult the
-#  documentation for your spawner to verify!
-c.Spawner.args = ['-s', str(DIST)]
-
-#  Some spawners allow shell-style expansion here, allowing you to use
-#  environment variables. Most, including the default, do not. Consult the
-#  documentation for your spawner to verify!
+# the command the hub should spawn (i.e. the cylc uiserver itself)
 c.Spawner.cmd = ['cylc', 'uiserver']
 
-
-# --- The class to use for spawning single-user servers.
-
-#  Should be a subclass of Spawner.
+# the spawner to invoke this command
 c.JupyterHub.spawner_class = 'jupyterhub.spawner.LocalProcessSpawner'
 
+# this auto-spawns uiservers without user interaction
 c.JupyterHub.implicit_spawn_seconds = 0.01
 
-# --- Cylc-ise Jupyterhub
-
-# TODO: move logo to a shared location
-# https://github.com/cylc/cylc-admin/issues/69
-c.JupyterHub.logo_file = str(DIST.joinpath(Path('img/logo.svg')))
-# use ISO8601 (expanded) date format for logging
-c.JupyterHub.log_datefmt = '%Y-%m-%dT%H:%M:%S'
-# specify custom HTML templates
+# apply cylc styling to jupyterhub
+c.JupyterHub.logo_file = str(Path(uis_pkg).parent / 'logo.svg')
+c.JupyterHub.log_datefmt = '%Y-%m-%dT%H:%M:%S'  # ISO8601 (expanded)
 c.JupyterHub.template_paths = [
+    # custom HTML templates
     pkg_resources.resource_filename(
         'cylc.uiserver',
         'templates'
     )
 ]
-c.JupyterHub.tornado_settings = {
-    'headers': {
-      # 'Access-Control-Allow-Origin': '*',
-      # 'Access-Control-Allow-Headers': '*',
-      # 'Access-Control-Allow-Methods': '*',
-      'Server': ''
-   },
-}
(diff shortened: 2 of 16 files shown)


setup.py
```diff
@@ -43,8 +42,7 @@ def find_version(*file_paths):
 #     don't pin versions, we will get whatever cylc-flow needs, and not
 #     the bleeding-edge version.
 install_requires = [
-    ('cylc-flow @ https://github.com/cylc/cylc-flow'
-     '/tarball/master#egg=cylc-8.0a3.dev'),
+    'cylc-flow>=8.0a3.dev',
```
oliver-sanders, line 45: Will need to pip install from the git repo in GH actions when implemented.

oliver-sanders:
Note: travis dead, will need to test locally until convterted to GH Actions

README.md
```diff
@@ -19,29 +19,49 @@ Workflow Services (WFS).
...
 
-Once that is done, the Hub can be started with `jupyterhub`.
 The default URL is [http://localhost:8000](http://localhost:8000).
 
-### Starting multiple instances of jupyterhub
+# Configuring
 
-If you need to run multiple instances of `jupyterhub`, you will have to follow
-the steps described below:
+The default "base" configuraion is defined in
```
kinow, line 60: s/configuraion/configuration

kinow (approved):

cylc/uiserver/config_defaults.py
```diff
@@ -19,8 +18,10 @@
 from pathlib import Path
 import pkg_resources
 
-HERE = Path(__file__).resolve().parent
-DIST = HERE.joinpath(Path('../cylc-ui/dist'))
+from cylc.uiserver import __file__ as uis_pkg
+
+
+DIST = Path(uis_pkg).parents[3] / 'cylc-ui/dist'
```
dwsutherland, line 24: Had to setup a link here:
```
(uiserver) sutherlander@cortex-vbox:python3.7$ pwd
/home/sutherlander/.envs/uiserver/lib/python3.7
(uiserver) sutherlander@cortex-vbox:python3.7$ ls -l
total 12
lrwxrwxrwx   1 sutherlander sutherlander    31 Feb 24 01:09 cylc-ui -> /home/sutherlander/cylc/cylc-ui
drwxr-xr-x 142 sutherlander sutherlander 12288 Feb 24 00:56 site-packages
```

I suppose there's a better way of doing this..

dwsutherland (approved):
Works fine for me 👍  .. Happy to merge.

cylc/uiserver/config_defaults.py
oliver-sanders: Ah, ok, I'm guessing you've not installed the UIS in editable mode so `cylc.uiserver.__file__` is showing up as being within the python lib rather than the UIS working copy.

I'm not sure how we are going to install the UI into the UIS yet, I think `DIST` should be configurable somehow.

oliver-sanders:
I've now made the path to the UI files a configurable traitlet. The UIS now loads in the same config file as the hub but has it's own `UIServer` namespace which so far contains just one config, `ui_path`.

* For editable installations (`pip install -e .`) you do not need to set anything, it loads `../cylc-ui`.
* For non-editable installations (`pip install .`) you will need to point the UIS at the UI files from your `~/.cylc/hub/config.py` file using the `c.UIServer.ui_path` configuration.

I'm not sure I'm working with traitlets correctly, let me know if I'm going wrong. If it's more involved, happy to drop the last commit for now.

kinow (commented):
~I think it didn't pick my config?~ (ah, now

```
[I 2021-02-25 09:16:02.207 JupyterHub spawner:1451] Spawning jupyterhub-singleuser --port=57197
Failed to set groups [Errno 1] Operation not permitted
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/bin/jupyterhub-singleuser", line 6, in <module>
    from jupyterhub.singleuser import main
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/jupyterhub/singleuser/__init__.py", line 5, in <module>
    from .app import main
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/jupyterhub/singleuser/app.py", line 16, in <module>
    App = import_item(JUPYTERHUB_SINGLEUSER_APP)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/traitlets/utils/importstring.py", line 30, in import_item
    module = __import__(package, fromlist=[obj])
ModuleNotFoundError: No module named 'notebook'
```
Does it mean users must create the config elsewhere first? I assume that means users that have a `jupyterhub_config.py` already in use, will have to link/copy it where `cylc hub` looks for, or deploy it to a folder and start `jupyterhub`?

oliver-sanders:
If you already have a `jupyterhub_config` for running regular jupyterhub then your options are:

1. Specify it on the CLI (which you would need to do anyway) i.e `cylc hub -f /path/to/config/file`.
2. Symlink it to `~/.cylc/hub/config.py` (will be automatically loaded by `cylc hub`)
3. Import it from `~/.cylc/hub/config.py` (will be automatically loaded by `cylc hub`)
4. Run plain `jupyterhub` for full control.


Not sure how you got that traceback, is that from running `jupyterhub`?

cylc/uiserver/config_defaults.py
oliver-sanders: You should now be able to set the static path for your UI installation using the user config like so:

```python
#  ~/.cylc/hub/config.py
c.Spawner.cmd = ['cylc', 'uiserver', '-s', '<path>']
```

kinow:
>Not sure how you got that traceback, is that from running jupyterhub?

Yes, running `jupyterhub` raises this error. So far the only differences I could tell after this PR were having to install `cylc-flow` first, and not being able to use `jupyterhub` directly.

Running `cylc hub` works. I think users who want/prefer to use `jupyterhub` may find it confusing, and have to search for a way to extract the configuration used by Cylc for Jupyterhub (e.g. if they are deploying to k8s in a HA env following the jupyterhub guide). But we probably won't have users with that set up initially, so not a blocker IMO.

If I install with `pip install .` (not editable), then I think I get the same error as @dwsutherland after I launch the UIS.

```
2021-03-02 14:17:00,839 tornado.application ERROR    Uncaught exception GET /user/kinow/?redirects=1 (127.0.0.1)
HTTPServerRequest(protocol='http', host='localhost:8000', method='GET', uri='/user/kinow/?redirects=1', version='HTTP/1.1', remote_ip='127.0.0.1')
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/tornado/web.py", line 1702, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/tornado/web.py", line 1905, in wrapper
    return method(self, *args, **kwargs)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/tornado/web.py", line 3173, in wrapper
    return method(self, *args, **kwargs)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/cylc/uiserver/handlers.py", line 71, in get
    self.render(index, python_base_url=base_url)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/tornado/web.py", line 863, in render
    html = self.render_string(template_name, **kwargs)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/tornado/web.py", line 1009, in render_string
    t = loader.load(template_name)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/tornado/template.py", line 446, in load
    self.templates[name] = self._create_template(name)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/tornado/template.py", line 477, in _create_template
    with open(path, "rb") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/cylc-ui/dist/index.html'
2021-03-02 14:17:00,841 tornado.access ERROR    500 GET /user/kinow/?redirects=1 (127.0.0.1) 2.72ms
```

oliver-sanders:
> and not being able to use jupyterhub directly.

If you want to bypass `cylc hub` just specify the path to the config file manually:

```console
$ jupyterhub -f cylc/uiserver/config.py
```

The `jupyterhub_config.py` file we have on master would not have been included with the pypi package so running `jupyterhub` directly like this was only ever a dev option for in-place installations.

> If I install with pip install . (not editable), then I think I get the same error as @dwsutherland after I launch the UIS.

For a non-editable install you will need to specify the ui path in your user config:

```
#  ~/.cylc/hub/config.py
c.Spawner.cmd = ['cylc', 'uiserver', '-s', '<path>']
```

kinow (approved):

kinow:
>If you want to bypass cylc hub just specify the path to the config file manually:
>
>For a non-editable install you will need to specify the ui path in your user config:

Thanks!

PR #175, opened 2021-02-19
play service

### LINKED ISSUE #166: mutation: play
Add a mutation and resolver for invoking `cylc play`.

If possible extend the GraphQL schema at the UIS (since this is not a service that a running Scheduler can/should perform).

It will be necessary to go via the command line in order to invoke wrapper script as users should be able to configure the `CYLC_VERSION` in the mutation.

The mutation should support the main CLI options to `cylc play`, namely:

* `--initial-cycle-point`
* `--final-cycle-point`
* `--start-cycle-point`
* `--stop-cycle-point`
* `--hold`
* `--hold-point`
* `-m`
* `--host`
* `--debug`
* `-s`

(note some of the other opts and the `START_POINT` arg will be disappearing soon)

To reduce the lag between starting the workflow and it appearing in the data-store we could use the `--format=json` option which returns the contact file data in JSON format when `cylc play` exits.

Note that other similar mutations are to follow e.g. `cylc install` & `cylc reinstall`.

**Pull requests welcome!**

oliver-sanders:
closes: #166

* Add a `play` mutation for starting workflow runs, the first of the "services" in the original Cylc8 plan.
* Add mypy config file and fix complaints.
* Fix flake8 complaints.

**Requirements check-list**
- [ ] Does not contain off-topic changes (includes unrelated flake8 and mypy fixes).
- [ ] Appropriate tests are included (system tests required, infrastructure not yet in place)
- [x] Only test dependency changes

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -15,8 +15,179 @@
 
 """GraphQL resolvers for use in data accessing and mutation of workflows."""
 
+import getpass
+import logging
+import os
+from subprocess import Popen, PIPE, DEVNULL
+
 from cylc.flow.network.resolvers import BaseResolvers
 from cylc.flow.data_store_mgr import WORKFLOW
+
+
+# show traceback from cylc commands
+DEBUG = True
+
+logger = logging.getLogger(__name__)
+
+
+def snake_to_kebab(snake):
+    """Convert snake_case text to --kebab-case text.
+
+    Examples:
+        >>> snake_to_kebab('foo_bar_baz')
+        '--foo-bar-baz'
+        >>> snake_to_kebab('')
+        ''
+        >>> snake_to_kebab(None)
+        TypeError: None type
+
+    """
+    if isinstance(snake, str) and snake:
+        return f'--{snake.replace("_", "-")}'
+    else:
+        raise TypeError(type(snake))
+    return ''
+
+
+def check_cylc_version(version):
+    """Check the provided Cylc version is available on the CLI.
+
+    Sets CYLC_VERSION=version and tests the result of cylc --version
+    to make sure the requested version is installed and selectable via
+    the CYLC_VERSION environment variable.
+    """
+    proc = Popen(
+        ['cylc', '--version'],
+        env={**os.environ, 'CYLC_VERSION': version},
+        stdin=DEVNULL,
+        stdout=PIPE,
+        stderr=PIPE,
+        text=True
+    )
+    ret = proc.wait(timeout=5)
+    out, err = proc.communicate()
+    return ret or out.strip() == version
+
+
+class Services:
+    """Cylc services provided by the UI Server."""
+
+    @staticmethod
+    async def play(workflows, args, workflows_mgr):
+        """Calls `cylc play`."""
+        response = []
+
+        # get ready to run the command
+        try:
+            # check that the request cylc version is available
+            cylc_version = None
+            if 'cylc_version' in args:
+                cylc_version = args['cylc_version']
+                if not check_cylc_version(cylc_version):
+                    return {'error': {'message': [
+                        f'cylc version not available: {cylc_version}'
+                    ]}}
+                args = dict(args)
+                args.pop('cylc_version')
+
+            # build the command
+            cmd = ['cylc', 'play', '--color=never']
+            for key, value in args.items():
+                if value is False:
+                    # don't add binary flags
+                    continue
+                key = snake_to_kebab(key)
+                if not isinstance(value, list):
+                    value = [value]
+                for item in value:
+                    cmd.append(key)
+                    if item is not True:
+                        # don't provide values for binary flags
+                        cmd.append(item)
+
+        except Exception as exc:
+            # oh noes, something went wrong, send back confirmation
+            return response.append(
+                [
+                    {
+                        'error': {
+                            'message': str(exc)
+                        }
+                    }
+                ]
+            )
+
+        # start each requested flow
+        me = getpass.getuser()
+        for user, flow, _ in workflows:
+            cmd_key = f'{user}|{flow}'
+            try:
+                if user != me:
+                    # TODO: multi-user support
+                    # forward this request to the other users UIS
+                    raise ValueError(f'Cannot start flow for user: {user}')
+
+                # add the workflow to the command
+                cmd = [*cmd, flow]
+
+                # get a representation of the command being run
+                cmd_repr = ' '.join(cmd)
+                if cylc_version:
+                    cmd_repr = f'CYLC_VERSION={cylc_version} {cmd_repr}'
+                logger.info(f'$ {cmd_repr}')
+
+                # run cylc run
+                proc = Popen(
+                    cmd,
+                    stdin=DEVNULL,
+                    stdout=PIPE,
+                    stderr=PIPE,
+                    text=True
+                )
+                ret = proc.wait(timeout=20)
+
+                if ret:
+                    # command failed
+                    _, err = proc.communicate()
+                    raise Exception(
+                        f'Could not start {flow} - {cmd_repr}'
+                        # suppress traceback unless in debug mode
+                        + (f' - {err}' if DEBUG else '')
+                    )
+
+            except Exception as exc:
+                # oh noes, something went wrong, send back confirmation
+                response.append(
+                    (
+                        cmd_key,
+                        [
+                            {
+                                'error': {
+                                    'message': str(exc)
+                                }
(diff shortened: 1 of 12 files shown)


cylc/uiserver/resolvers.py
```diff
@@ -15,10 +15,189 @@
...
+            )
+
+        # start each requested flow
+        me = getpass.getuser()
+        for user, flow, _ in workflows:
+            cmd_key = f'{user}|{flow}'
+            try:
+                if user != me:
+                    # TODO: multi-user support
+                    # foreward this request to the other users UIS
```
kinow, line 134: s/foreward/forward

cylc/uiserver/schema.py
```diff
@@ -0,0 +1,224 @@
...
+
+                Required if not defined in flow.cylc.
+            ''')
+        )
+        start_cycle_point = CyclePoint(
+            description=sstrip('''
+                Set the start cycle point, which may be after the initial cycle
+                point.
+
+                If the specified start point is not in the sequeunce, the next
```
kinow, line 120: s/sequeunce/sequence

cylc/uiserver/resolvers.py
```diff
@@ -15,10 +15,189 @@
...
+        ['cylc', '--version', '--color=never'],
+        env={**os.environ, 'CYLC_VERSION': version},
+        stdin=DEVNULL,
+        stdout=PIPE,
+        stderr=PIPE,
+        text=True
+    )
+    ret = proc.wait(timeout=5)
+    out, err = proc.communicate()
+    return ret or out.strip() == version
```
kinow, line 69: I tested the function above in the IDE (checked out the branch, "Python Console" auto-imports the venv and I can play with the new code) and it always returned `1` (not a bool, but no biggie I think).

image

I tried the command above and it failed too.

```py
(venv) kinow@ranma:~/Development/python/workspace/cylc-uiserver$ cylc --version --color=never
Error: cylc --color=never: unknown utility. Abort.
Type "cylc help all" for a list of utilities.
```

I think we need to replace `--version` with `version`, i.e. `cylc version --color=never`?

Screenshot from 2021-03-05 10-17-50

cylc/uiserver/resolvers.py
```diff
@@ -15,10 +15,189 @@
...
+import logging
+import os
+from shlex import quote
+from subprocess import Popen, PIPE, DEVNULL
+
 from cylc.flow.network.resolvers import BaseResolvers
 from cylc.flow.data_store_mgr import WORKFLOW
 
 
+DEBUG = True
```
kinow, line 28: Shouldn't we use the `debug` argument from the `Play` mutation (from the schema)?

cylc/uiserver/resolvers.py
```diff
@@ -15,10 +15,189 @@
...
+            # oh noes, something went wrong, send back confirmation
+            response.append(
+                [
+                    {
+                        'error': {
+                            'message': str(exc)
+                        }
+                    }
+                ]
+            )
```
kinow, line 125: Should we return the error response here instead of continuing the execution flow?

cylc/uiserver/resolvers.py
```diff
@@ -15,10 +15,189 @@
...
+                    _, err = proc.communicate()
+                    raise Exception(
+                        f'Could not start {flow} - {cmd_repr}'
+                        # suppress traceback unless in debug mode
+                        + (f' - {err}' if DEBUG else '')
+                    )
+
+            except Exception as exc:
+                # oh noes, something went wrong, send back confirmation
+                breakpoint()
```
kinow, line 167: Debug left-over

cylc/uiserver/resolvers.py
```diff
@@ -15,10 +15,189 @@
...
+        stderr=PIPE,
+        text=True
+    )
+    ret = proc.wait(timeout=5)
+    out, err = proc.communicate()
+    return ret or out.strip() == version
+
+
+class Services:
+    """Services provided by the UI Server."""
```
kinow, line 73: When I first read it I thought these were [JupyterHub services](https://jupyterhub.readthedocs.io/en/stable/reference/services.html) :rofl:

cylc/uiserver/resolvers.py
```diff
@@ -15,10 +15,189 @@
...
+                    if item is not True:
+                        # don't provide values for binary flags
+                        if key == '--set':
+                            # TODO
+                            # don't quote this value as it messes with things
+                            # I think subprocess quotes everything anyway
+                            # double check
+                            cmd.append(item)
+                        else:
+                            cmd.append(quote(item))
```
kinow, line 113: Should we fail somewhere if the `mode` is not provided? I got an error in GraphiQL when I didn't provide it.

Screenshot from 2021-03-05 10-48-56

But providing one of the valid enum values fixed it.

Screenshot from 2021-03-05 10-48-40

kinow (changes requested):
Few comments, but looking good! And the best part was being able to call the new Play mutation from the Cylc UI :+1: :rocket:

cylc/uiserver/resolvers.py
oliver-sanders: Good testing!

I added the `--color=never` for safety as an afterthought, clearly not as safe as I had thought!

cylc/uiserver/resolvers.py
oliver-sanders: Whoops, I had used `RunMode.Live.name` rather than `.value` in the schema!

cylc/uiserver/resolvers.py
oliver-sanders: Ah, sorry I didn't add any comments here to explain what this variable is for.

This is a response to the "no traceback" secure mode, I thought I would add a flag here to make it easier to implement this later.

There is a cylc-flow issue, which is slightly unhelpful as I think it's the UIS that this really applies to - https://github.com/cylc/cylc-flow/issues/3834

cylc/uiserver/resolvers.py
oliver-sanders: Good point, will clarify.

We called these "services" back at the beginning of the Cylc8 program which is why I named it this here, however, they aren't really services any more, just methods. Perhaps we need a better name.

cylc/uiserver/resolvers.py
oliver-sanders: Yeah, that would make sense!

oliver-sanders:
Sorry, sloppier code than usual, I think I forgot why I had marked this as a draft, thanks for the great reviewing.

kinow (approved):

kinow:
@oliver-sanders it has conflicts now. Rebasing should also trigger GH actions after @MetRonnie fixed it :tada:

dwsutherland (approved):
A good way of forking the schema to add content irrelevant to `cylc-flow` (was bound to happen eventually)..
Works in my environment.. Good to merge once de-conflicted (and tests pass? GH Actions?)

cylc/uiserver/resolvers.py
```diff
@@ -15,10 +15,190 @@
...
+                cmd = [*cmd, quote(flow)]
+
+                # get a representation of the command being run
+                cmd_repr = ' '.join(cmd)
+                if cylc_version:
+                    cmd_repr = f'CYLC_VERSION={cylc_version} {cmd_repr}'
+                logger.info(f'$ {cmd_repr}')
+
+                # run cylc run
+                proc = Popen(
```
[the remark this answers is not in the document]
oliver-sanders: This is to prevent tethering the UIS version to the Flow version allowing users to start flows with whatever Cylc versions are made available via the Cylc wrapper script.

Hopefully we will be able to push for greater uniformity in Cylc version by making it easier to upgrade UI/UIS/Flow versions on-the-fly (e.g. could add an "upgrade" service to the UIS for restarting flows under different Cylc versions), however, there is still a large use case for running with diff versions e.g. at the MO:

* Our OS/PS trailing is performed with the same Cylc version as the flows will eventually be put into production with.
* We deploy the latest Cylc release but do not make it default until we have had people try it out for a few weeks.

Hopefully the supported version range will be fairly minimal though trailing could cause a drift of say 6 months.

oliver-sanders:
The coverage on actions was only reporting the `cylc/uiserver/__init__.py` file, I think this is because of the way `pytest --cov` was being run which was causing it to skip the `.coveragerc` file.

However, I'm only seeing 28.27% on actions vs 69.28% locally...

oliver-sanders:
Rebased, deconflicted, added mypy to actions, tests pass, merging with two approvals.

PR #181, opened 2021-03-09
Upload coverage only for 3.8, and use editable mode to install the module

### LINKED ISSUE #179: coverage: investigate results
Coverage results don't match local runs.

See comment https://github.com/cylc/cylc-uiserver/pull/175#issuecomment-793725736

kinow:
These changes close #179 (I think :sweat_smile: )

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,9 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
 
+      # NOTE: We install in both editable, and non-editable mode. The former is
+      # necessary to produce coverage reports with the correct numbers.
+      # https://github.com/cylc/cylc-uiserver/issues/179
       - name: Install cylc-uiserver & dependencies
         run: |
           pip install git+https://github.com/cylc/cylc-flow@master
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,8 @@ jobs:
   test:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 15
+    env:
+      PYTHON_COVERAGE_ENABLED: '3.8'
     strategy:
       matrix:
         os: ['ubuntu-latest']
@@ -44,6 +46,7 @@ jobs:
         run: pytest --cov
 
       - name: Coverage
+        if: ${{ success() &&  matrix.python == env.PYTHON_COVERAGE_ENABLED }}
         run: |
           coverage xml --ignore-errors
           bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -44,6 +54,7 @@ jobs:
         run: pytest --cov
 
       - name: Coverage
+        if: ${{ success() &&  matrix.python == env.PYTHON_COVERAGE_ENABLED }}
```
kinow, line 57: `success()` is optional, but we probably don't care if the build failed. If it goes ahead, we could have a single point in the coverage graph where coverage was reported as `0` or lower than the expected.

Also, unless we limit when we upload the coverage, the build in Codecov will have multiple XML's uploaded, which is not necessary, and may cause troubles if the coverage module is not working as expected on win/mac/linux. I think for CI we can pick linux + python stable or the python version most of us are using? :man_shrugging:

.github/workflows/test.yml
```diff
@@ -29,10 +31,18 @@ jobs:
...
+      # https://github.com/cylc/cylc-uiserver/issues/179
       - name: Install cylc-uiserver & dependencies
         run: |
           pip install git+https://github.com/cylc/cylc-flow@master
-          pip install .[all]
+          if [ "${{ matrix.python }}" == "${{ env.PYTHON_COVERAGE_ENABLED }}" ]; then
+            pip install -e .[all]
+          else
+            pip install .[all]
+          fi
```
kinow, line 44: Installing in editable mode only when we want to run coverage, and in non-editable mode otherwise. This way we end up testing both? I think we had an issue some time ago where something worked in editable mode, and not in the default mode. Can't remember if it was relative imports, or some CI stuff…

.github/workflows/test.yml
```diff
@@ -10,6 +10,8 @@ jobs:
   test:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 15
+    env:
+      PYTHON_COVERAGE_ENABLED: '3.8'
```
kinow, line 14: Happy to change (or if anyone commits to this branch changing this value)

kinow:
To review it, scroll down to the build checks. Expand it, open the `ubuntu-latest, 3.8`. Locate the `Test` job, expand it, confirm the output of the tests coverage match what we have when running the tests with coverage locally.

Codecov reports appear to have an issue comparing pull requests, otherwise its UI would show the coverage for this pull request (or at least it's not working for me). Once merged, I believe the newly reported coverage metric should be close to what we expect (NB: branch coverage calculation in Python coverage differs [or used to] to how Codecov calculates it).

.github/workflows/test.yml
oliver-sanders: Seems like a sensible value to me.

PR #182, opened 2021-03-13
Add dependabot to GH repository

kinow:

Maybe worth a try here before considering adding it to cylc-flow? It's a free service, used by other projects. It runs daily, and reports any dependency in our dependency range that can be updated.

I normally go through the main dependencies in our `setup.py` and look them up at PYPI. But dependabot will handle that.

Furthermore, it is probably already enabled in GitHub if our repository settings has not disabled the dependabot security checks (click settings / security I think, then look for dependabot security).

I enabled it for one of my repositories with Python. It executed almost immediately.

image

Each PR created by dependabot will run the CI pipeline too, confirming whether there was any build/test failures due to the new version. Developers can also check out the branch and try the code.

image

So I think it could be helpful to use this, instead of relying on us remembering to update the dependencies before a release?

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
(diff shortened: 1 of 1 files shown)


kinow:
GH actions successfully run against this repository, coverage updated correctly too. Everything looks OK, merging.

PR #185, opened 2021-03-19
play: fix play service

oliver-sanders:
Fix an issue with a default value in the schema which was causing issues in the UI, may come back to this later, for now easier to remove it (note the end result is the same either way around).

Also fix the response to allow the UI to detect command failure.

**Requirements check-list**
- [ ] Tests (needs some kind of full-system test, don't have time to create a new framework pre-beta)
- [x] No change log entry required (why? un-released)
- [x ] No dependency changes.

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -74,7 +74,23 @@ class Services:
     """Cylc services provided by the UI Server."""
 
     @staticmethod
-    async def play(workflows, args, workflows_mgr):
+    def _error(message):
+        """Format error case response."""
+        return [
+            False,
+            str(message)
+        ]
+
+    @staticmethod
+    def _return(message):
+        """Format success case response."""
+        return [
+            True,
+            message
+        ]
+
+    @classmethod
+    async def play(cls, workflows, args, workflows_mgr):
         """Calls `cylc play`."""
         response = []
 
@@ -85,9 +101,9 @@ class Services:
             if 'cylc_version' in args:
                 cylc_version = args['cylc_version']
                 if not check_cylc_version(cylc_version):
-                    return {'error': {'message': [
+                    return cls._error(
                         f'cylc version not available: {cylc_version}'
-                    ]}}
+                    )
                 args = dict(args)
                 args.pop('cylc_version')
 
@@ -108,22 +124,13 @@ class Services:
 
         except Exception as exc:
             # oh noes, something went wrong, send back confirmation
-            return response.append(
-                [
-                    {
-                        'error': {
-                            'message': str(exc)
-                        }
-                    }
-                ]
-            )
+            return cls._error(exc)
 
         # start each requested flow
         me = getpass.getuser()
         for user, flow, _ in workflows:
-            cmd_key = f'{user}|{flow}'
             try:
-                if user != me:
+                if user and user != me:
                     # TODO: multi-user support
                     # forward this request to the other users UIS
                     raise ValueError(f'Cannot start flow for user: {user}')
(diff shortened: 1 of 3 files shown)


cylc/uiserver/schema.py
```diff
@@ -150,9 +150,7 @@ class Arguments:
                 this cycle point.
             ''')
         )
-        mode = RunMode(
-            default_value=RunMode.Live.value  # type: ignore
-        )
+        mode = RunMode()
```
oliver-sanders, line 153: This default value was causing client-server errors, not sure why, removing for the beta.

cylc/uiserver/resolvers.py
```diff
@@ -108,22 +124,13 @@ async def play(workflows, args, workflows_mgr):
...
         except Exception as exc:
             # oh noes, something went wrong, send back confirmation
-            return response.append(
-                [
-                    {
-                        'error': {
-                            'message': str(exc)
-                        }
-                    }
-                ]
```
oliver-sanders, line 118: Graphene does this templating for us.

mypy.ini
```diff
@@ -3,6 +3,6 @@ python_version = 3.7
 ignore_missing_imports = True
 files = cylc/uiserver
 # don't run mypy on these files directly
-exclude = cylc/uiserver/(websockets/tornado.py|config_defaults.py)
+exclude = cylc/uiserver/(websockets/tornado.py|config_defaults.py|tests/*)
```
oliver-sanders, line 6: MyPy really doesn't like the compiled protobuf file imported in the tests. Inline ignore comments didn't seem to be enough to suppress the error. Tried copying the cylc-flow mypy config but that didn't fix it so just ignoring the tests for now, can re-instate later.

kinow: Did you add the `# type: ignore` in the ... wrong line? I noticed from @MetRonnie that he had to add the comment in the initial line when you had a multi-line import.

If I remove `test/*` from the exclusion in your branch, the command fails. Then this diff fixes it.

```diff
diff --git a/cylc/uiserver/tests/conftest.py b/cylc/uiserver/tests/conftest.py
index 167ba6c..a8a838e 100644
--- a/cylc/uiserver/tests/conftest.py
+++ b/cylc/uiserver/tests/conftest.py
@@ -19,7 +19,7 @@ import inspect
 
 import pytest
 import zmq
-from cylc.flow.data_messages_pb2 import (
+from cylc.flow.data_messages_pb2 import (  # type: ignore
     PbEntireWorkflow, PbWorkflow, PbFamilyProxy)
 from cylc.flow.network import ZMQSocketBase
 
diff --git a/mypy.ini b/mypy.ini
index 20eca2f..b0b4d93 100644
--- a/mypy.ini
+++ b/mypy.ini
@@ -3,6 +3,6 @@ python_version = 3.7
 ignore_missing_imports = True
 files = cylc/uiserver
 # don't run mypy on these files directly
-exclude = cylc/uiserver/(websockets/tornado.py|config_defaults.py|tests/*)
+exclude = cylc/uiserver/(websockets/tornado.py|config_defaults.py)
 # don't run mypy on those files when imported from other modules
 follow_imports = silent

```

kinow (approved):
Looks good. Just one comment about `mypy`

mypy.ini
MetRonnie: Yes, it only worked for me when I used `# type: ignore` on the first line of the multiline import

oliver-sanders: Tried `type: ignore` but there were problems besides that, managed to get them fixed here https://github.com/cylc/cylc-uiserver/pull/188/commits/0029244360b68328861211bb8ff24ab4a5ab5f8f

PR #198, opened 2021-03-30
Update deps.

hjoliver:
Add two missing dependencies identified by the conda feedstock bot. 

They were being pulled in as dependencies of other packages, so no need to re-do the 0.3.0 release for this. (I will add them to the conda recipe for the current release though, since that isn't done yet).

**Requirements check-list**
- [x] Updated conda feedstock PR https://github.com/conda-forge/cylc-uiserver-feedstock/pull/12

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,14 +42,17 @@ def find_version(*file_paths):
 #     dependencies first. This way, if other dependencies (e.g. jupyterhub)
 #     don't pin versions, we will get whatever cylc-flow needs, and not
 #     the bleeding-edge version.
+# NB: no graphene version specified; we only make light use of it in our own
+#     code, so graphene-tornado's transitive version should do.
+#     Same for traitlets via jupyterhub transitive deps.
 install_requires = [
     'cylc-flow>=8.0b0',
     'jupyterhub==1.3.*',
     'tornado==6.1.*',
-    'traitlets>=5.0.5',
-    'graphene>=2.1.8',
     'graphene-tornado==2.6.*',
     'graphql-ws>=0.3.1,<0.4'
+    'traitlets',
+    'graphene',
 ]
 
 # Only include pytest-runner in setup_requires if we're invoking tests
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,8 @@ install_requires = [
     'cylc-flow>=8.0b0',
     'jupyterhub==1.3.*',
     'tornado==6.1.*',
+    'traitlets>=5.0.5',
+    'graphene>=2.1.8',
     'graphene-tornado==2.6.*',
     'graphql-ws>=0.3.1,<0.4'
 ]


kinow (approved):

kinow (changes requested):
Maybe we can pin on `x.x.*`? For hotfixes, etc?

kinow:
> > Maybe we can pin on `x.x.*`? For hotfixes, etc?
> 
> Probably sensible, but does that mean we should have tested with x.x.0...n where n is the current version we're using? (In case those earlier versions don;'t work)

Ah, these are transitive dependencies from `jupyterhub` and `graphene-tornado`. Hmmm. Might be better to pin to the exact version as you did, to match what we have in the upstream dependencies.

Or should we just leave as transitive dependencies, removing from `setup.py`? This way we will always get the same version used by `jupyterhub` and by `graphene-tornado`?

hjoliver:
> Ah, these are transitive dependencies from jupyterhub and graphene-tornado .... Or should we just leave as transitive dependencies, removing from setup.py? This way we will always get the same version used by jupyterhub and by graphene-tornado?

We import from graphene directly, in the schema:
```
$ git grep import | grep graphene 
cylc/uiserver/schema.py:from graphene import (
cylc/uiserver/schema.py:from graphene.types.generic import GenericScalar
```

And @oliver-sanders recent config changes uses traitlets:
```
$ git grep traitlets
cylc/uiserver/main.py:from traitlets import (
cylc/uiserver/main.py:from traitlets.config import (
cylc/uiserver/main.py:        # set traitlets values from the config
```
So I figure we should be treating them as explicit dependencies now, no?

kinow:
In Java and Javascript that's what I'd do. In Java normally version ranges are not used, only exact versions

In JS, however, I'd match the dependency of the upstream dependency. 

So for traitlets that'd be traitlets>=4.3.2, and graphene graphene>=3.0b1 I think?

This way when a new version of JupyterHub is released, for example, we only need to check if they changed their dependency range for traitlets.

If not, we change only the version of jupyterhub. And if traitlets is released, we know we are matching what's declared in jupyterhub. WDYT?

setup.py
```diff
@@ -46,6 +46,8 @@ def find_version(*file_paths):
     'cylc-flow>=8.0b0',
     'jupyterhub==1.3.*',
     'tornado==6.1.*',
+    'traitlets>=5.0.5',
+    'graphene>=2.1.8',
```
kinow, line 50: I believe setuptools also installs dependencies in order. So this graphene gets installed, then I **think** if the version doesn't match with the version used in `graphene-tornado` below, setuptools may decide to use the version from `graphene-tornado`.

hjoliver: Good point. Damn, this dependencies stuff is more involved than I thought :grimacing:

hjoliver:
> In Java and Javascript that's what I'd do. In Java normally version ranges are not used, only exact versions
> 
> In JS, however, I'd match the dependency of the upstream dependency.
> 
> So for traitlets that'd be [`traitlets>=4.3.2`](https://github.com/jupyterhub/jupyterhub/blob/a91197635ab1a9d8c9b3d2e83b4f213d0d535d8d/requirements.txt#L15), and graphene `graphene>=2.1,<3` I think? (NOTE: no tags for graphene-tornado, but you can confirm the version from their [sources](https://pypi.org/project/graphene-tornado/#files))
> 
> This way when a new version of JupyterHub is released, for example, we only need to check if they changed their dependency range for traitlets.
> 
> If not, we change only the version of jupyterhub in our setup.py. If only traitlets is released, then at least we know we are matching what's declared in jupyterhub. WDYT?

Just to be clear, are you (a) agreeing that we should have these as explicit dependencies now, because we import them directly in our own code; but (b) suggesting that we match versions with what our other dependencies require for the same packages?

(b) sounds somewhat sensible at first glance, but (at second glance!) I don't think we should need to be aware of the package versions required by our dependencies (and their dependencies, and so on...) ??

kinow:
Yeah, even in Java and JS some projects I worked we enforced never to use a transitive dependency without declaring it (there are plug-ins to enforce that during build). But I've seen people ignoring that too, and just trusting that the dependency in the final resolution of the dependency manager system was golden :man_shrugging: 

I think either a) or b) would be fine. Ignore my other option (c?) that was to pin like `1.1.*`, leaving bug fixes with a wild-card, as that could break our build. Sorry for the confusion and for not having a straight answer.

setup.py
oliver-sanders: As these are transitive dependencies it may be safer to specify them with no version constraint and bump them to the end of the list (which is effectively the same as not specifying them in the `setup.py` file).

kinow: Ah, good to know. I thought specifying after the dependency would trigger setuptools to install the latest. So +1 to moving them to the end of the file with no version range 👍

hjoliver: @kinow and @oliver-sanders - am I misunderstanding what a "transitive dependency" is, or have you missed my comment above?  https://github.com/cylc/cylc-uiserver/pull/198#issuecomment-809834016  

(I'm saying they are explicit dependencies, because we import them directly, not transitive!)

oliver-sanders: Perhaps more a misuse of "transitive" on my part.

They are deps which are brought in by other deps, we don't care much what version we get so long as those deps are happy. At the moment not specifying them as deps at all works fine but is also yuck since we do make some light use of them ourselves.

hjoliver: >  we do make some light use of them ourselves 

Well technically that means they are not transitive, so we should specify them as explicit cylc-uiserver deps, right?  Otherwise our code can break implicitly if JupyterHub (say) changes its deps under the hood.

oliver-sanders: Yes.

hjoliver: >  So this graphene gets installed, then I *think* if the version doesn't match with the version used in graphene-tornado below, setuptools may decide to use the version from graphene-tornado.

Actually, that would be an unsolvable conflict wouldn't it?  (Two packages request mutually incompatible versions of graphene).

hjoliver: Oh, you probably mean not conflicting, i.e. if the version ranges overlap which one will get priority (within the ranges).  But technically that shouldn't matter, because either package should be happy with any version within their specified range.

hjoliver: OK I've change the order, to put the most specific requirement first:
```
    'graphene-tornado==2.6.*',
    'graphene>=2.1.8',
```

setup.py
```diff
@@ -46,7 +46,9 @@ def find_version(*file_paths):
     'cylc-flow>=8.0b0',
     'jupyterhub==1.3.*',
     'tornado==6.1.*',
+    'traitlets>=5.0.5',
     'graphene-tornado==2.6.*',
+    'graphene>=2.1.8',
```
kinow, line 51: I think we can both move the order, and leave without any version range. Just `traitlets` and `graphene`? That way the version installed is always the version specified in the upstream dependency.

setup.py
hjoliver: (reverting - I forgot actually look at what version of graphene graphene-tornado wanted!)

setup.py
hjoliver: OK I finally understand.  I thought @oliver-sanders comment depended on these being transitive dependencies, which they aren't, but I guess our own light use of them is unlikely to be version-dependent.

setup.py
kinow, line 50: What I understood was to have

```suggestion
    'traitlets',
    'graphene',
```

But the `graphene` after `graphene-tornado` I think?

setup.py
hjoliver: Done.

setup.py
hjoliver: That's what I've done, isn't it?

hjoliver: Maybe you didn't refresh your browser tab?

hjoliver: (Oh, you probably reviewed just before I reverted the previous change then pushed the new one, sorry)

kinow (approved):

hjoliver:
I'll take this as approved by @oliver-sanders since I essentially (/eventually!) just applied his suggested change.

Issue #171, opened 2021-02-05
Cross-user operation - umbrella issue


hjoliver:
(And related topics, such as how UI Servers behave when not looked at, etc.). See also:
- #10 (authorization)
- #27 ("list other UI Servers running")

This is to record an @oliver-sanders "brain dump" on the topic, in response to questions from @jarich, from Element chat, as that seems to be the current best record of earlier discussions (primarily the Feb 2020 workshop).  To be refined and replaced by more focused Issues as we go...

> Brain dump time (apologies in advance), contains some extra details and contents of discussions not currently documented...
>
>>    If 2 or more users want to view Bob's workflows, will there be only one UI server running for Bob?
>>
>>    If I attempt to spawn a UI server from the command line for Bob, and there is already a UI server running for Bob, is that a no-op?
>
>    - There will only ever be one UI server (UIS) per "owner" running under one hub.
>    - When you visit the URL for that server Jupyterhub will spawn the UIS if it is not running then redirect, otherwise I think it is a simple redirect at the reverse-poxy (very efficient).
>    - The UIS knows about open connections so is able to tell when it is not in use. We anticipate building in a configurable shutdown timeout.
>    - An idle UIS should have a low footprint and starting up a UIS should be cheap.
>    - We currently anticipate that startup/shutdown for the UIS will be efficient and that aggressively shutting-down unused UIServers is likely a good idea.
>    - This will help to keep the UIServer flock on the latest version.
>    - UIServers can be manually shutdown through the web interface.
>
> Additionally:
>
>   - The websocket connections between the UIS and UI have a "heartbeat" which gives us a robust way of telling when connections are inactive from the UIS end.
>   - We have a plan for the UI to detect inactivity and suspend connections, freezing the UI but maintaining its state until the user interacts with it again.
>   - When this happens the UI would probably go grey with a message in the middle saying "paused, click to resume" or something like that.
>   - The time it takes to sync a workflow through to the UI is pretty good, even for larger flows so it should be fast to resume.
>
>> If the UI Server is not being run as a user with a password, then it will need to persist even with no connections. Or at least some part of it will have to. This could potentially be done with a flag on start up?
>
> We will want to make the "shutdown on inactivity" feature configurable anyway so should be good there, however, if the UIS went down for some other reason it would have to be brought back up manually.
>
>>    what is the pathway the user makes?
>
>   - The URL for the UI contains the username, these URLs can be shared and should work for other users.
[...]

Issue #10, opened 2019-03-10
Define how the authorization in the UI server will work with other components


kinow:
JupyterHub has authentication, and also a thin layer of authorization in its reverse proxy. But the Cylc UI server will need to redefine parts of its authentication.

This issue is to record discussions on authorization for the UI server, which will interface with JupyterHub and probably be used by Cylc Web as a bridge to communicate with cylc and JupyterHub.

See related discussion in cylc/cylc: https://github.com/cylc/cylc/issues/1901

See related discussion in cylc/web: https://github.com/cylc/cylc-web/issues/60

kinow:
Discussion on Discourse with good user requirements: https://cylc.discourse.group/t/the-role-of-the-ui-server-in-production-from-a-user-and-permissions-point-of-view-cylc8/55

kinow:
While testing the LDAP Authenticator today (#51), I started looking for the groups. I tried to understand the concept of groups in JupyterHub, related to Authenticators.

An authenticator and a spawner may use the system groups to authenticate the user, or when forking a spawner process to use the user/group IDs.

But after that, the REST API endpoint `/groups` and the SQLite database table `groups`, both seem to be more related to JupyterHub, and not to the system groups.

So even though my Linux user has over 5 groups (`kinow`, `cdrom`, `adm`, `docker`, etc), after my first log in, the groups table is empty.

I can use the REST API to add groups to my existing user, but these groups won't be really related to my OS users.

This may be useful in the future in case users want to have authorization based on their existing groups in Active Directory, FreeIPA, etc. I couldn't find an existing way to sync groups between OS and JupyterHub; though it should be doable either by extending the authenticator, or by using a [service](https://jupyterhub.readthedocs.io/en/stable/reference/services.html).

kinow:
Kubernetes has a good model for RBAC: https://kubernetes.io/docs/reference/access-authn-authz/rbac/

Jenkins too, and it's configurable (using multiple strategies, like their matrix authorization strategy: https://plugins.jenkins.io/matrix-auth/)

I think JupyterHub RBAC might look similar to the Kubernetes model. In which case, we could use their authorization too.

(commenting here because the issue I found for Cylc Flow appears to be more for the CLI https://github.com/cylc/cylc-flow/issues/2729, and whatever is defined here would drive the UI too I think)

hjoliver:
Comments on authorization here:

- https://github.com/cylc/cylc-uiserver/issues/27#issuecomment-668478159
[...]

PR #202, opened 2021-04-09
uis: authorise endpoint for multi-user use

### LINKED ISSUE #203: tests: test authentication failure for all handlers
Two tests which should assert authentication failure works as intended currently fail due to internal server errors which ocour before the authentication check takes place.

See https://github.com/cylc/cylc-uiserver/pull/202#discussion_r612289330

oliver-sanders:
This adds a trivial authorisation scheme to the UIS which rejects anyone but the UIS user.

addresses: #10, #171
opens and closes #203 

This will be followed by a configurable authorisation scheme in due course.

For testing, change the value of `ME` or try with a second user account. For now manual testing only, can implement automated next release.

**Requirements check-list**
- [x] Appropriate tests are included (see above)
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -36,15 +36,32 @@ logger = logging.getLogger(__name__)
 ME = getpass.getuser()
 
 
+def _authorised(req):
+    """Authorise a request.
+
+    Requires the request to be authenticated.
+
+    Currently returns True if the authenticated user is the same as the
+    user under which this UI Server is running.
+
+    Returns:
+        bool - True if the request passes authorisation, False if it fails.
+
+    """
+    user = req.get_current_user()
+    username = user.get('name', '?')
+    if username != ME:
+        logger.warn(f'Authorisation failed for {username}')
+        return False
+    return True
+
+
 def authorised(fun):
     """Provides Cylc authorisation for multi-user setups."""
 
     def _inner(self, *args, **kwargs):
         nonlocal fun
-        user = self.get_current_user()
-        username = user.get('name', '?')
-        if username != ME:
-            logger.warn(f'Authorisation failed for {username}')
+        if not _authorised(self):
             raise web.HTTPError(403)
         return fun(self, *args, **kwargs)
     return _inner
@@ -55,12 +72,8 @@ def async_authorised(fun):
 
     async def _inner(self, *args, **kwargs):
         nonlocal fun
-        user = self.get_current_user()
-        username = user.get('name', '?')
-        if username != ME:
-            logger.warn(f'Authorisation failed for {username}')
+        if not _authorised(self):
             raise web.HTTPError(403)
-            return
         return await fun(self, *args, **kwargs)
     return _inner
 
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -51,7 +51,7 @@ def _authorised(req):
     user = req.get_current_user()
     username = user.get('name', '?')
     if username != ME:
-        logger.warn(f'Authorisation failed for {username}')
+        logger.warning(f'Authorisation failed for {username}')
         return False
     return True
 
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -65,7 +65,7 @@ def async_authorised(fun):
     return _inner
 
 
-class BaseHandler(web.RequestHandler):
+class BaseHandler(HubOAuthenticated, web.RequestHandler):
 
     def set_default_headers(self) -> None:
         self.set_header("X-JupyterHub-Version", jupyterhub_version)
@@ -80,7 +80,7 @@ class StaticHandler(BaseHandler, web.StaticFileHandler):
     """A static handler that extends BaseHandler (for headers)."""
 
 
-class MainHandler(HubOAuthenticated, BaseHandler):
+class MainHandler(BaseHandler):
 
     # hub_users = ["kinow"]
     # hub_groups = []
@@ -102,7 +102,7 @@ class MainHandler(HubOAuthenticated, BaseHandler):
         self.render(index, python_base_url=base_url)
 
 
-class UserProfileHandler(HubOAuthenticated, BaseHandler):
+class UserProfileHandler(BaseHandler):
 
     def set_default_headers(self) -> None:
         super().set_default_headers()
@@ -117,7 +117,7 @@ class UserProfileHandler(HubOAuthenticated, BaseHandler):
 # This is needed in order to pass the server context in addition to existing.
 # It's possible to just overwrite TornadoGraphQLHandler.context but we would
 # somehow need to pass the request info (headers, username ...etc) in also
-class UIServerGraphQLHandler(HubOAuthenticated, TornadoGraphQLHandler):
+class UIServerGraphQLHandler(BaseHandler, TornadoGraphQLHandler):
 
     # Declare extra attributes
     resolvers = None
@@ -176,8 +176,7 @@ class UIServerGraphQLHandler(HubOAuthenticated, TornadoGraphQLHandler):
         await TornadoGraphQLHandler.run(self, *args, **kwargs)
 
 
-class SubscriptionHandler(BaseHandler, HubOAuthenticated,
-                          websocket.WebSocketHandler):
+class SubscriptionHandler(BaseHandler, websocket.WebSocketHandler):
 
     def initialize(self, sub_server, resolvers):
         self.queue = Queue(100)
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -62,7 +62,7 @@ def authorised(fun):
     def _inner(self, *args, **kwargs):
         nonlocal fun
         if not _authorised(self):
-            raise web.HTTPError(403)
+            raise web.HTTPError(403, reason='authorisation insufficient')
         return fun(self, *args, **kwargs)
     return _inner
 
@@ -73,7 +73,7 @@ def async_authorised(fun):
     async def _inner(self, *args, **kwargs):
         nonlocal fun
         if not _authorised(self):
-            raise web.HTTPError(403)
+            raise web.HTTPError(403, reason='authorisation insufficient')
         return await fun(self, *args, **kwargs)
     return _inner
 
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -16,6 +16,8 @@
 import json
 import os
 from asyncio import Queue
+import logging
+import getpass
 
 from graphene_tornado.tornado_graphql_handler import TornadoGraphQLHandler
 from graphql import get_default_backend
@@ -26,6 +28,41 @@ from tornado import web, websocket
 from tornado.ioloop import IOLoop
 
 from .websockets import authenticated as websockets_authenticated
+
+
+logger = logging.getLogger(__name__)
+
+
+ME = getpass.getuser()
+
+
+def authorised(fun):
+    """Provides Cylc authorisation for multi-user setups."""
+
+    def _inner(self, *args, **kwargs):
+        nonlocal fun
+        user = self.get_current_user()
+        username = user.get('name', '?')
+        if username != ME:
+            logger.warn(f'Authorisation failed for {username}')
+            raise web.HTTPError(403)
+        return fun(self, *args, **kwargs)
+    return _inner
+
+
+def async_authorised(fun):
+    """Provides Cylc authorisation for multi-user setups."""
+
+    async def _inner(self, *args, **kwargs):
+        nonlocal fun
+        user = self.get_current_user()
+        username = user.get('name', '?')
+        if username != ME:
+            logger.warn(f'Authorisation failed for {username}')
+            raise web.HTTPError(403)
+            return
+        return await fun(self, *args, **kwargs)
+    return _inner
 
 
 class BaseHandler(web.RequestHandler):
@@ -43,13 +80,6 @@ class StaticHandler(BaseHandler, web.StaticFileHandler):
     """A static handler that extends BaseHandler (for headers)."""
 
 
-class APIHandler(BaseHandler):
-
-    def set_default_headers(self) -> None:
-        super().set_default_headers()
-        self.set_header("Content-Type", 'application/json')
-
-
 class MainHandler(HubOAuthenticated, BaseHandler):
 
     # hub_users = ["kinow"]
@@ -61,6 +91,7 @@ class MainHandler(HubOAuthenticated, BaseHandler):
 
     @web.addslash
     @web.authenticated
+    @authorised
     def get(self):
         """Render the UI prototype."""
         index = os.path.join(self._static, "index.html")
@@ -71,13 +102,14 @@ class MainHandler(HubOAuthenticated, BaseHandler):
         self.render(index, python_base_url=base_url)
 
 
-class UserProfileHandler(HubOAuthenticated, APIHandler):
+class UserProfileHandler(HubOAuthenticated, BaseHandler):
 
     def set_default_headers(self) -> None:
         super().set_default_headers()
         self.set_header("Content-Type", 'application/json')
 
     @web.authenticated
+    @authorised
     def get(self):
         self.write(json.dumps(self.get_current_user()))
 
@@ -122,9 +154,12 @@ class UIServerGraphQLHandler(HubOAuthenticated, TornadoGraphQLHandler):
         return wider_context
 
     @web.authenticated
+    @authorised
     def prepare(self):
         super().prepare()
 
+    @web.authenticated
+    @async_authorised
     async def execute(self, *args, **kwargs):
         # Use own backend, and TornadoGraphQLHandler already does validation.
         return await self.schema.execute(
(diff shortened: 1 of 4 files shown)


oliver-sanders:
Four tests are unhappy due to unsatisfied auth.

cylc/uiserver/handlers.py
```diff
@@ -28,7 +30,55 @@
 from .websockets import authenticated as websockets_authenticated
 
 
-class BaseHandler(web.RequestHandler):
+logger = logging.getLogger(__name__)
+
+
+ME = getpass.getuser()
```
kinow, line 36: Not sure if this would work for every JupyterHub authenticator? Maybe for some users the user database is stored in a database, external system, etc. So maybe `getpass.getuser()` would lock us to linux & pam only?

oliver-sanders: In order for the UIS to run the user has to exist on the system so it should be good for most things but in more esoteric systems where the authenticated user does not match any system user it would break yes (although I'm not sure how the hub could spawn UI Servers in this case).

We will add the ability to configure authorisation (i.e. grant access to different accounts) that should help us to handle this sort of thing.

kinow: Maybe we could use Tornado's [`current_user`](https://www.tornadoweb.org/en/stable/guide/security.html#user-authentication) then?

When the hub is configured to use PAM, it will match what user you have logged in. Then we can compare that against the request path.

So if I logged in with a database backed authenticator and have the user `kinow`, but I'm accessing the URL `/#/workflows/userx|one`, then our code to authorise would just match `current_user` (`kinow` from the database in this case, but would work with PAM, etc) and fail if I don't have authorization I think?

Also, if we run Cylc UIS with `root` or another superuser, what would `getpass.getuser()` return? Wouldn't it return `root` instead of the logged-in user?

cylc/uiserver/tests/test_handlers.py
```diff
@@ -44,18 +47,35 @@ def get_app(self) -> Application:
...
+        response = self.fetch('/')
+        assert response.code == 500
+
+    # TODO: can't get this test to run due to handler setup?
+    # @pytest.mark.usefixtures("mock_authentication_none")
+    # def test_unauthenticated(self) -> None:
+    #     """Test 403 HTTP response (unauthenticated)."""
+    #     response = self.fetch('/')
+    #     assert response.code == 403
+    #     # assert response.reason == 'Forbidden'
```
oliver-sanders, line 71: I get 500 errors for these tests as the UIS hits an internal error before authentication has the change to fail.

Should find a way to configure things so these tests run in the future, in the mean time can test manually (behaviour unchanged).

oliver-sanders: https://github.com/cylc/cylc-uiserver/issues/new?assignees=&labels=&template=feature-request.md&title=

cylc/uiserver/handlers.py
oliver-sanders: > Also, if we run Cylc UIS with root or another superuser

The UIS should never be run as root. It is only run under a regular system user.

oliver-sanders: I think `current_user` is the result of `get_current_user` which returns the authenticated user who performed the request.

We need to know the user that the UIS is running under.

kinow: >The UIS should never be run as root. It is only run under a regular system user.

I thought BoM was considering using a system user with elevated privileges for the UIS.

oliver-sanders: I think that's for the Hub?

It should still work for an account with elevated privileges, but they would need to authenticate with the Hub as the elevated user, the hub would then spawn the UIS as that user.

kinow: Ah yeah, you are correct, the UIS should be with my own user, my bad!

datamel (approved):
Does the job as expected with regards to the need to secure the UI Server now so I'm happy to get this merged in for the beta 1. I have manually attempted to break into another user uiserver and can see this branch fixes the issue.
 
Have a couple of queries which don't affect approval...
- Are you sure the `nonlocal` is needed with wrapped function arguments?
- Probably don't need to override `run` and may also be more secure to override `post` instead? Both of these comments are tested on my POC PR: https://github.com/cylc/cylc-uiserver/pull/204.

Thanks for getting this fixed @oliver-sanders!

MetRonnie (approved):
<del>Haven't managed to use the UI (see Teams)</del> but tests pass, LGTM

Update: tested with `cylc.uiserver.handlers.ME = 'darmok'` and I get the "403: authorisation insufficient" message  when launching the cylc hub, as expected

oliver-sanders:
> Are you sure the nonlocal is needed with wrapped function arguments?

Needed 100% of the time, no.

I use `nonlocal` whenever I'm getting a variable from outside of the defined scope. Sometimes it is stictly necessary, sometimes it is not. It is less ambiguous to use `nonlocal` and serves as a sanity check as it will fail if the value *is* defined locally. It also saves Python the bother of poking around the local scope looking for something which we know won't be there there, and, as an added bonus It is an indication to others that the variable is stored in a higher level scope, i.e. "don't mess with this value, it may be reused".

> Probably don't need to override run and may also be more secure to override post instead?

Probably right, I followed the "better safe than sorry" on that one. I did try overriding `post` but had issues with 500 errors although that might have been before I changed the inheritance order.

cylc/uiserver/handlers.py
```diff
@@ -28,7 +30,55 @@
...
+    user under which this UI Server is running.
+
+    Returns:
+        bool - True if the request passes authorisation, False if it fails.
+
+    """
+    user = req.get_current_user()
+    username = user.get('name', '?')
+    if username != ME:
+        logger.warn(f'Authorisation failed for {username}')
```
kinow, line 54: `.warn` is deprecated in logging.

```suggest
       logger.warning(f'Authorisation failed for {username}')
```

kinow:
Tested with the `LocalProcessSpawner` and it worked fine (single user, his own hub & UIS).

Tested with `SudoSpawner` and at the moment getting several

```
2021-04-14 13:27:26,265 cylc.uiserver.websockets DEBUG    Unauthenticated WebSocket request!
2021-04-14 13:27:26,267 tornado.access WARNING  403 GET /user/kinow/subscriptions (127.0.0.1) 4.38ms
2021-04-14 13:28:26,317 cylc.uiserver.websockets DEBUG    Unauthenticated WebSocket request!
2021-04-14 13:28:26,320 tornado.access WARNING  403 GET /user/kinow/subscriptions (127.0.0.1) 10.12ms
```

One tab has `kinow` logged-in, and I can see the workflows stopped fine in the UI.

The other tab has `cylc1` logged-in (still have this user from the first time I tested Sudo Spawner in 2019? :laughing: ), but his tab keeps loading and it appears websockets are not happy. I might have something missing in the `cylc1` environment.

Hub is running as `root` (too lazy to set up the environment the way they [recommend](https://github.com/jupyter/jupyterhub/wiki/Using-sudo-to-run-JupyterHub-without-root-privileges)).

kinow:
Tested on Firefox, got the same behavior. Then, since I had two tabs, one with `kinow` working fine, and one with `cylc1` kinda broken, I tried accessing `http://localhost:8000/user/kinow/#/workflows/complex` from the tab logged-in as `cylc1`. And I got the following screen.

kinow:
BTW, stopping the Hub with 2 UIS's running with the Sudo Spawner, it tidily stopped the processes and logged everything (cc @jarich).

image

kinow (commented):
On `master` I also cannot use the second user... or at least it doesn't show anything in the UI. So for a separate issue.

~Only thing that's happening in this PR is that when I open `cylc1`, there are no warnings in the console. But when I open `kinow` UIS, even without trying to access `cylc1`'s UIS, I get:~

```
2021-04-14 14:00:23,157 cylc.uiserver.websockets DEBUG    Unauthenticated WebSocket request!
```

~Every few seconds a new line like this. I only have one tab open, with the user logged in.~

~Maybe I need to install something for `cylc1`? Maybe it needs to have at least 1 workflow? I tried creating `~/cylc-run`, but it still failed when I opened the UI :confused: ~

Solved! In the virtual machine, I had a browser minimized with a tab with my `kinow` user, but logged out. Then websocket connection was still open, causing the warnings! 👍

kinow (approved):

cylc/uiserver/handlers.py
hjoliver: https://docs.python.org/3/library/logging.html#logging.warning

hjoliver: (Checked we haven't used this anywhere else :+1: )

cylc/uiserver/handlers.py
hjoliver: >  in more esoteric systems where the authenticated user does not match any system user it would break yes (although I'm not sure how the hub could spawn UI Servers in this case).

Not sure that's true.  The authenticator doesn't have to know about users on the target system where the UIS and Schedulers run.  So long as our authorization system doesn't require the authenticated user to be valid in its domain, we should be able to authorize "alien" users to access the workflows, no?

hjoliver: (We do of course have to trust that the authentication system has been configured safely).

cylc/uiserver/handlers.py
kinow: And deprecated in 3.3: https://bugs.python.org/issue13235

oliver-sanders:
Good to know, I had spotted that warning at some point in the past.

cylc/uiserver/handlers.py
oliver-sanders: > we should be able to authorize "alien" users to access the workflows, no?

Yes, however, those users won't be able to start UI Servers.

(This thread was about the use of `getpass.getuser()` in the UIS for determining the name of UIS owner.)

cylc/uiserver/handlers.py
```diff
@@ -135,9 +183,13 @@ async def execute(self, *args, **kwargs):
             **kwargs,
         )
 
+    @web.authenticated
+    @async_authorised
```
kinow, line 187: I think we copied this code from a Tornado pull request. And back then, we didn't add the `web.authenticated` as `prepare` had these already. Now I suspect `web.authenticated` and `async_authorised` are executed 3 times?

`execute` is called by `graphene-tornado`'s `TornadoGraphQLHandler` in `execute_graphql_request`.

Probably we can leave `web.authenticated` and `async_authorised` in `prepare` only?

cylc/uiserver/handlers.py
```diff
@@ -148,6 +200,13 @@ def select_subprotocol(self, subprotocols):
...
 
     @websockets_authenticated
+    @authorised
+    def get(self, *args, **kwargs):
+        # forward this call so we can authenticate/authorise it
+        return websocket.WebSocketHandler.get(self, *args, **kwargs)
+
+    @websockets_authenticated
+    @authorised
     def open(self, *args, **kwargs):
```
kinow, line 210: I think this is never called without passing through `get` first. We will start an HTTP GET request with the [upgrade header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism). Then once that's accepted and authenticated and authorized, then the connection opens.

kinow (commented):
I think we are checking authorization multiple times for some requests, but would have to debug it to confirm.

@oliver-sanders for the `TODO` tests, there's a few things that we can change so that they pass

1. Add a cookie secret for the application (otherwise authentication is not even handled properly (see the [application](https://www.tornadoweb.org/en/stable/guide/security.html#user-authentication) in the code snippet)
2. Add a handler for the login URL (unauthenticated requests are not 403'd, instead they are redirected to the login URL I think? at least that's the way UIS works at the moment)
3. Confirm that we were redirected (or maybe force Tornado to throw 403, even though that's not how UIS works) 

Here's how I got both TODO's removed, and the tests passing locally

```diff
diff --git a/cylc/uiserver/tests/test_handlers.py b/cylc/uiserver/tests/test_handlers.py
index 0cc911f..a0266ce 100644
--- a/cylc/uiserver/tests/test_handlers.py
+++ b/cylc/uiserver/tests/test_handlers.py
@@ -24,7 +24,7 @@ from graphql_ws.constants import GRAPHQL_WS
 from tornado.httpclient import HTTPResponse
 from tornado.httputil import HTTPServerRequest
 from tornado.testing import AsyncHTTPTestCase, get_async_test_timeout
-from tornado.web import Application, HTTPError
+from tornado.web import Application, RequestHandler
 
 from cylc.uiserver.main import (
     MainHandler,
@@ -36,6 +36,12 @@ from cylc.uiserver.main import (
 import pytest
 
 
+class NoOpHandler(RequestHandler):
+
+    def get(self):
+        pass
+
+
 class MainHandlerTest(AsyncHTTPTestCase):
     """Test for the Main handler"""
 
@@ -43,8 +49,10 @@ class MainHandlerTest(AsyncHTTPTestCase):
         self.tempdir = tempfile.mkdtemp(suffix='mainhandlertest')
         return MyApplication(
             handlers=[
-                ('/', MainHandler, {"path": self.tempdir})
-            ]
+                ('/', MainHandler, {"path": self.tempdir}),
+                ('/hub/api/oauth2/authorize.*', NoOpHandler)
+            ],
+            cookie_secret='MainHandlerTest'
         )
 
     @pytest.mark.usefixtures("mock_authentication")
@@ -62,13 +70,12 @@ class MainHandlerTest(AsyncHTTPTestCase):
         response = self.fetch('/')
         assert response.code == 500
 
-    # TODO: can't get this test to run due to handler setup?
-    # @pytest.mark.usefixtures("mock_authentication_none")
-    # def test_unauthenticated(self) -> None:
-    #     """Test 403 HTTP response (unauthenticated)."""
-    #     response = self.fetch('/')
-    #     assert response.code == 403
-    #     # assert response.reason == 'Forbidden'
+    @pytest.mark.usefixtures("mock_authentication_none")
+    def test_unauthenticated(self) -> None:
+        """Test login URL redirect response (unauthenticated)."""
+        response = self.fetch('/')
+        assert response.code == 200
+        assert response.effective_url.index('/hub/api/oauth2/authorize') > 0
 
     @pytest.mark.usefixtures("mock_authentication_yossarian")
     def test_unauthorised(self):
@@ -88,8 +95,10 @@ class UserProfileHandlerTest(AsyncHTTPTestCase):
     def get_app(self) -> Application:
         return MyApplication(
             handlers=[
-                ('/userprofile', UserProfileHandler)
-            ]
+                ('/userprofile', UserProfileHandler),
+                ('/hub/api/oauth2/authorize.*', NoOpHandler)
+            ],
+            cookie_secret='MainHandlerTest'
         )
 
     @pytest.mark.usefixtures("mock_authentication")
@@ -107,13 +116,12 @@ class UserProfileHandlerTest(AsyncHTTPTestCase):
         response = self.fetch('/userprofile')
         assert response.code == 200
 
-    # TODO: can't get this test to run due to handler setup?
-    # @pytest.mark.usefixtures("mock_authentication_none")
-    # def test_unauthenticated(self) -> None:
-    #     """Test 403 HTTP response (unauthenticated)."""
-    #     response = self.fetch('/userprofile')
-    #     assert response.code == 403
-    #     # assert response.reason == 'Forbidden'
+    @pytest.mark.usefixtures("mock_authentication_none")
+    def test_unauthenticated(self) -> None:
+        """Test login URL redirect (unauthenticated)."""
+        response = self.fetch('/userprofile')
+        assert response.code == 200
+        assert response.effective_url.index('/hub/api/oauth2/authorize') > 0
 
     @pytest.mark.usefixtures("mock_authentication_yossarian")
     def test_unauthorised(self):

```

cylc/uiserver/handlers.py
kinow: We can look into these multiple calls to authenticate/authorize later :+1:

kinow:
(We can change the test in a follow-up, only posted here because I had it in GitHub notifications and decided to look why these tests were failing :sweat_smile: )

cylc/uiserver/handlers.py
hjoliver: OK, you're talking about the current PR, not where we're heading (any authenticated user can spawn a UIS on other users' accounts).

hjoliver (approved):
Tested as working :+1:

hjoliver:
@oliver-sanders - leaving it to you to merge now or deal with @kinow's comments first.

cylc/uiserver/handlers.py
oliver-sanders: > Now I suspect web.authenticated and async_authorised are executed 3 times?

Yep, I expect you are right, I've been over cautious here to be safe.

I suggest we open an issue to come back to this for a later beta release as I hope there is a better way of going about it. There must be a way of making auth happen by default?

#### context

I couldn't easily tell what path requests would take through the code. When you are subclassing a handler, without digging through the code you have no idea what methods that handler implements, (get, post, delete, etc). My fear is that a subclassed handler we use will later implement a method we aren't expecting and open up an unauthorised route into the code.

So I just slapped these everywhere as auth is computationally cheep to perform but an unauthorised endpoint could be very very expensive in other ways.

I would really like a way to change the default so we authorise/authenticate all methods of all endpoints unless stated otherwise. Some other web frameworks use `@expose`-like decorators for similar reasons.

kinow: Let's move it to a separate issue then. Thanks!

oliver-sanders: Ok bumped - https://github.com/cylc/cylc-uiserver/issues/207

oliver-sanders:
Merging with four approvals.

PR #204, opened 2021-04-13
WIP - Authorisation POC

datamel:
### Configurable access 'groups'

Added decorators based on initial thoughts for names in [authorization config files](https://github.com/cylc/cylc-uiserver/issues/10#issuecomment-773752542):

- can_read
- can_write
- can_execute

The verb nature of these is fairly limiting for future possibility of moving actions and mutations between groups. We could consider alternatives such as: users, task-operators, workflow-admins, etc.

I suspect I can refactor the methods I have in the PoC for this. into a single method.

### def _user_action_allowed(username, action)

Shared by all access routes in. 

Note, I've added a dictionary,`roles_dict`, directly to handler to fake up some config data while developing this proof of concept.

- Checks username against configured users for the provided action.
- Fetches all groups (local and LDAP) for the browser user and checks against configured groups for the provided action.

### can_read and _can_read

- `can_read` is a dectorator that can work for both async and non-async handler function), which wraps `_can_read`
- `_can_read` checks if browser user == user spawning the UI Server and calls `_user_action_allowed`

### can_write and _can_write

As the read access function.

### can_execute and _can_execute

The execute access group is thought to be for `play` and `stop` so we need to pick those graphql queries out separately. So for graph POSTS we expect something like:

```python
 @web.authenticated
 @can_write
 @can_execute(mutators=["play", "stop"])
```

If any of these raises a 403, access is forbidden e.g.

- If a user is authenticated and in the writers access group but not in executers they would be denied access
- If a user is authenticated, in writers and executers then they are allowed

Need some thought around inheritance between access groups here.

For security we may need to check the mutator query as well as the `operation_name` as it might be the case that a user could make their browser send a different operation name to that being executed (see code TODO notes).

## Other notes on code

- The override of Execute injects what we need. I don't think we need the `run` added in [202](https://github.com/cylc/cylc-uiserver/pull/202)?
- I did think it was more appropriate to decorate an override of the `Post` method with the authentication and authorisation, at the 'top' of the handler, if you like, before hitting any of our code. We can also override the `Get` and decorate that too but we don't use that (yet?) - a good way to get a distinction between read and write operations we could consider. As it stands, all of our graph calls are mutators and considered writes (not queries) and the stuff I have marked for `can_read` are: loading the UI itself and subscription to updates.

### Note on Config

This will need careful consideration to enable future configuration of mutations (if this is required).
Perhaps something like...

```
[readers]
    [[mutations]]
    ....
    [[users]] 
        alice
        bob
    [[groups]]
        linuxvdi

[writers]
    [[mutations]]
        trigger, pause
    [[users]] 
        claire
        dave
    [[groups]]
        task_administrators
[exectutors]
    [[mutations]]
        start, stop
    [[users]] 
        eddie
        frank
    [[groups]]
        special-group    

```

### Other thoughts / future iterations of work

- Users could move actions and mutations between the access groups
- Users could control 'inheritance' between groups
- Users could 'create' further groups

These changes partially address #xxxx
These changes close #xxxx

**Requirements check-list**

- [ ] Does not need tests (why?).

- [ ] No change log entry required (why? e.g. invisible to users).

- [ ] Created an issue at [cylc-uiserver conda-forge repository](https://github.com/conda-forge/cylc-uiserver-feedstock) with version changes (if you changed dependencies in `setup.py`, see `recipe/meta.yaml`).
- [ ] No dependency changes.


cylc/uiserver/config_defaults.py
```diff
@@ -46,3 +47,10 @@
 c.JupyterHub.cookie_secret_file = f'{RUNTIME_PATH / "cookie_secret"}'
 c.JupyterHub.db_url = f'{RUNTIME_PATH / "jupyterhub.sqlite"}'
 c.ConfigurableHTTPProxy.pid_file = f'{RUNTIME_PATH / "jupyterhub-proxy.pid"}'
+
+# Ensure env variable CYLC_SITE_CONF_PATH is passed to UI Server environment
+c.Spawner.environment = {
+    'CYLC_SITE_CONF_PATH': f'{os.getenv("CYLC_SITE_CONF_PATH")}'}
+
+# remove this before auth goes in...
```
datamel, line 55: I'll remove this before I mark it ready for review. I use it for testing, to ensure I can debug without timeout.

CONTRIBUTING.md
```diff
@@ -41,6 +41,7 @@ below.
  - Stephen McVeigh
  - Ronnie Dutta
  - David Matthews
+ - Mel Hall
```
kinow, line 44: :rocket:

cylc/uiserver/config_defaults.py
```diff
@@ -46,3 +47,7 @@
 c.JupyterHub.cookie_secret_file = f'{RUNTIME_PATH / "cookie_secret"}'
 c.JupyterHub.db_url = f'{RUNTIME_PATH / "jupyterhub.sqlite"}'
 c.ConfigurableHTTPProxy.pid_file = f'{RUNTIME_PATH / "jupyterhub-proxy.pid"}'
+
+# Ensure env variable CYLC_SITE_CONF_PATH is passed to UI Server environment
+c.Spawner.environment = {
+    'CYLC_SITE_CONF_PATH': f'{os.getenv("CYLC_SITE_CONF_PATH")}'}
```
kinow, line 53: Hmmm, I think there's a simpler way to tell JupyterHub what variables must be passed down to the UI Server. Will search their docs...

cylc/uiserver/main.py
```diff
@@ -110,6 +113,100 @@ def validate(self, obj, value):
...
+
+            If this configuration is empty, site authorization defaults to no
+            configuarble authorization and users will be unable to set any
+            authorization.
+        ''' + AUTH_DESCRIPTION + '''
+        Example Configuration:
+
+        .. code-block:: python
+
+        c.UIServer.site_authorisation = {
```
kinow, line 152: Nit-pick, but there's a mix of authorise, authorization, authorisation in our code & docs. Not sure if we should use just one of use them interchangeably.

cylc/uiserver/main.py
```diff
@@ -110,6 +113,100 @@ def validate(self, obj, value):
...
+        "user1": {                      # user1
+            "default": ["!ALL"],        # No privileges for all ui-server
+                                        # owners.
+        },                              # No limit set, so all ui-server owners
+    },                                  # limit is also "!ALL" for user1
+    "server_owner_1": {                 # For specific UI Server owner,
+        "group:group_a": {              # Any user who is a member of group_a
+            "default": "READ",          # Will have default read-only access
+            "limit": ["ALL", "!play"],  # server_owner_1 is able to give away
+        },                              # All privileges, except play.
```
kinow, line 166: Woah, really powerful our authorization here. Adding a note to self to remember to test this :+1: thanks for the great docs @datamel !

cylc/uiserver/main.py
```diff
@@ -110,6 +113,100 @@ def validate(self, obj, value):
...
+                "!stop",
+                "!kill",                # No default, so default is no access
+            ],
+        },
+    },
+}
+
+        ''')
+
+    user_authorization = Dict(
```
kinow, line 182: `site_authorisation`, vs. `user_authorization`, see comment above :point_up:

cylc/uiserver/main.py
```diff
@@ -110,6 +113,100 @@ def validate(self, obj, value):
...
+        config=True,
+        help='''
+            Dictionary containing authorized users and permission levels for
+            authorization. Use this setting to share control of your workflows
+            with other users.
+
+            Note that you are only permitted to give away permissions up to
+            your limit for each user, as defined in the site_authorization
+            configuration.
```
kinow, line 191: :+1: Note to self, verify during review.

cylc/uiserver/main.py
```diff
@@ -110,6 +113,100 @@ def validate(self, obj, value):
...
+
+            ''' + AUTH_DESCRIPTION + '''
+            Example configuaration, residing in `~/.cylc/hub/config.py`:
```
kinow, line 194: s/configuaration/configuration

cylc/uiserver/resolvers.py
```diff
@@ -204,7 +204,8 @@ async def mutator(self, info, *m_args):
             'request_string': req_str,
             'variables': variables,
         }
-        return self.workflows_mgr.multi_request('graphql', w_ids, graphql_args)
+        return await self.workflows_mgr.multi_request(
```
kinow, line 207: :+1:

cylc/uiserver/authorise.py
```diff
@@ -0,0 +1,477 @@
...
+            user_conf_permitted_ops = (
+                self.return_site_auth_defaults_for_access_user(
+                    access_user=access_user_dict))
+        user_conf_permitted_ops = expand_and_process_access_groups(
+            user_conf_permitted_ops)
+        limits_owner_can_give = expand_and_process_access_groups(
+            limits_owner_can_give)
+        allowed_operations = limits_owner_can_give.intersection(
+            user_conf_permitted_ops)
+        print(f"inside this method _get_permitted_ops for {access_user}....")
```
kinow, line 241: Should we use logging here? Or was this used for debugging/development?

cylc/uiserver/authorise.py
```diff
@@ -0,0 +1,477 @@
...
+        if not op_name:
+            self.auth_failed(self.current_user, op_name, 400,
+                             "Operation not in schema.")
+        authorised = self.auth.is_permitted(self.current_user, op_name)
+        if not authorised:
+            self.auth_failed(self.current_user, op_name, 403)
+        if (info.operation.operation in Authorization.ASYNC_OPS
+                or iscoroutinefunction(next_)):
+            return self.async_resolve(next_, root, info, **args)
+        return next_(root, info, **args)
```
kinow, line 369: I will need :eyeglasses: + :coffee: + IDE/debugger to review this function next week

cylc/uiserver/authorise.py
```diff
@@ -0,0 +1,477 @@
...
+            message: Message to log Defaults to None.
+
+        Raises:
+            web.HTTPError
+        """
+        log_message = (f"Authorization failed for {current_user}"
+                       f":requested to {op_name}.")
+        if message:
+            log_message = log_message + " " + message
+        logger.warn(log_message)
```
kinow, line 389: Just wondering if we should log authorization errors a bit different... if I am responsible for a webapp, I like to find a quick and simple way to find all the audit logs, where I can see things like authorization errors (similarly to crash logs, or trace logs, etc...).

Maybe always include some piece of text in each authorization error message, so users can easily grep for it (and we can maybe filter in the future Cylc Review replacement?).

Or maybe include an error number, or log to a different logger?

cylc/uiserver/authorise.py
```diff
@@ -0,0 +1,477 @@
...
+    """
+    groupmax = (os.NGROUPS_MAX)  # type: ignore
+    group_ids = os.getgrouplist(username, groupmax)
+    group_ids.remove(groupmax)
+    # turn list of group_ids into group names with group identifier prepended
+    return list(map(
+        lambda x: f'{Authorization.GRP_IDENTIFIER}{grp.getgrgid(x).gr_name}',
+        group_ids
+        )
+    )
```
kinow, line 434: Note to self: test with that LDAP library, to confirm we can get the users too (NIWA uses it and I think we have a really old issue in cylc-flow related to that...).

kinow (commented):
Just a super quick review done reading in the GitHub UI @datamel

cylc/uiserver/config_defaults.py
datamel: I went with this method as it was recommended on an [issue](https://github.com/jupyterhub/the-littlest-jupyterhub/issues/308). 
From reading the docs we could also use : `c.Spawner.env_keep.append('CYLC_SITE_CONF_PATH')` which may be a bit nicer.

kinow: Ah, I think that's the one I was thinking. If both work the same way, in local and remote environments, then either way is fine by me.

cylc/uiserver/main.py
datamel: Yes oops, thought I'd changed them all to the American spelling for consistency. Wonder if it would be worth supporting both spelling types for the config though?

kinow: I'm more used to seeing the American spelling for color (HTML standard) and authorization (W3C standards, Microsoft Active Directory, AWS, OAuth, SAML, Kubernetes, Docker), but I'm used to seeing both in documentation, websites, etc :man_shrugging: In Cylc 7 I think we only had authorisation, and there might be a Python module authorise or authorisation in cylc-flow. Maybe check with others in Element to see if we have to change, otherwise we can just keep both I guess.

cylc/uiserver/authorise.py
datamel: Excellent, would be great to test the groups as thoroughly as we can.

cylc/uiserver/authorise.py
datamel: Yes, good idea. I'll have a think and do some reading and see if I can come up with an alternative.

cylc/uiserver/authorise.py
datamel: Thanks, will remove. I was testing the caching of the method and forgot to remove.

kinow:
@datamel I've updated the issue description removing the boxes that were not necessary, and linked the issue #10? Not sure if that's the right issue, but since you were assigned I assume it was.

cylc/uiserver/authorise.py
kinow: @datamel looks like `logger.warn` is deprecated (IDE warning about the warn :smile: :drum: ). I think instead we must use `logger.warning`.

cylc/uiserver/authorise.py
```diff
@@ -0,0 +1,476 @@
...
+            operation: operation type
+
+        Returns:
+            operation name
+        """
+        if operation in Authorization.READ_AUTH_OPS:
+            return Authorization.READ_OPERATION
+        else:
+            # Check it is a mutation in our schema
+            if (field_name in Authorization.ALL_OPS):
```
kinow, line 407: Unnecessary parentheses?

cylc/uiserver/authorise.py
```diff
@@ -0,0 +1,476 @@
...
+def get_groups(username: str) -> List[str]:
+    """Return list of system groups for given user
+
+    Args:
+        username: username used to check system groups.
+
+    Returns:
+        list: system groups for username given
+    """
+    groupmax = (os.NGROUPS_MAX)  # type: ignore
```
kinow, line 425: Unnecessary parentheses?

cylc/uiserver/tests/test_authorise.py
```diff
@@ -0,0 +1,261 @@
...
+def test_expand_and_process_access_groups(permission_set, expected):
+    actual = expand_and_process_access_groups(permission_set)
+    assert actual == expected
+
+
+@pytest.mark.parametrize('mut_field_name, operation, expected_op_name', [
+    pytest.param('play',
+                 'mutation',
+                 'play',
+                 id="Muation operation, play field, returns play"),
```
kinow, line 210: s/Muation/Mutation

cylc/uiserver/tests/test_authorise.py
```diff
@@ -0,0 +1,261 @@
...
+    pytest.param('_schema',
+                 'subscription',
+                 'read',
+                 id="Subsciption operation, _schema field, returns read"),
```
kinow, line 214: s/Subsciption/Subscription

cylc/uiserver/tests/test_authorise.py
```diff
@@ -0,0 +1,261 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from getpass import getuser
+from os import path
+from graphql.execution.base import ResolveInfo
+from subprocess import Popen, PIPE
```
kinow, line 20: All imports above appear to be unused?

cylc/uiserver/tests/test_authorise.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+import pytest
+from tornado import web
```
kinow, line 24: Not used.

cylc/uiserver/tests/test_authorise.py
```diff
@@ -0,0 +1,261 @@
...
+
+from unittest.mock import MagicMock, patch
```
kinow, line 26: `MagicMock` is not used.

cylc/uiserver/tests/test_authorise.py
```diff
@@ -0,0 +1,261 @@
...
+
+from cylc.uiserver.authorise import (
+    Authorization,
+    AuthorizationMiddleware,
+    expand_and_process_access_groups,
+    get_groups
```
kinow, line 32: `get_groups` is not used

cylc/uiserver/tests/test_authorise.py
```diff
@@ -0,0 +1,261 @@
...
+)
+from cylc.uiserver.schema import schema
```
kinow, line 34: Not used.

kinow (commented):
@datamel checked out the branch today and remembered I hadn't used the IDE static analyzer yet. Found a few typos and unused imports. Marked them in this last review :+1:

dwsutherland:
> As of current implementation...We do not check auth for return variables, in order for this not to compromise security need confirmation that GraphQL mutations won't be nested (e.g. Can't perform a kill within a play)

No chance of nested here, we've designed the mutations as flat:
```
class Pause(Mutation):
    class Meta:
        description = sstrip('''
            Pause a workflow.
        ''')
        resolver = partial(mutator, command='pause')

    class Arguments:
        workflows = List(WorkflowID, required=True)

    result = GenericScalar()
```

So there's nowhere to nest..
In addition, all field resolving passes through the same middleware, and should share the same operation.

cylc/uiserver/authorise.py
```diff
@@ -0,0 +1,476 @@
...
+    Authorization has access groups: `READ`, `CONTROL`, `ALL` - along with
+    their negations, `!READ`, `!CONTROL` and `!ALL` which indicate removal of
+    the permission groups.
+
+    """
+
+    # config literals
+    DEFAULT = 'default'
+    LIMIT = 'limit'
+    GRP_IDENTIFIER = 'group:'
```
kinow, line 43: Wouldn't it be useful to have identifier for user and group? I'm using the sample configuration you provided, and searching for "group:group1" is quite useful. But if I have a user with name "group1", then searching will find both.

cylc/uiserver/authorise.py
```diff
@@ -0,0 +1,476 @@
...
+    remove = set()
+
+    for perm in permission_set:
+        if perm.startswith("!"):
+            with suppress(KeyError):
+                remove.add(perm.lstrip("!"))
+                remove.add(perm)
+    permission_set.difference_update(remove)
+    permission_set.discard('')
+    return permission_set
```
kinow, line 476: Should we log if the user passes a permission that does not exist? e.g.:

```
"group:group1": {"default": "TRUCK" }
```

Maybe warn too - if possible - if a key is duplicated? (I don't think we can though...), e.g.:

```
"group:group1": {"default": "ALL", "default": "TRUCK" }
```

kinow (commented):
@datamel I finished reading the proposal, and reading the code today :tada: 

Still need to play more with the UIS and authorization schema.

But one thing I couldn't find in the docs, I wonder if you could help me… I saw that we have a site and a user config. So I imagined myself as a system admin looking after a machine with several (100s?) users.

All users having ability to start their own hub (something like `module load cylc` or similar). Now, maybe I don't want users to start Cylc because of GraphQL queries consuming resources, or some other concern.

Maybe I could limit the module loading, but users could still use Conda to install perhaps. That got me thinking in **trying the site configuration to lock my own account**. i.e. set a default to `!ALL` or something like that.

I couldn't get that to work, and I couldn't find anything in the proposal. Is the authorization supposed to work only controlling authorization to other users?

Thanks!

cylc/uiserver/authorise.py
kinow: (Tested and the UIS loaded fine)

kinow (commented):
Found a lazy-way to review it :smile: 

Here's my `~/.cylc/hub/config.py`:

```
c.UIServer.ui_build_dir = '~/Development/python/workspace/cylc-ui/dist'  # path to build
c.JupyterHub.authenticator_class = 'jupyterhub.auth.DummyAuthenticator'

import pwd
def mocked_fn(any_value):
  class Obj(object):
    def __init__(self):
      self.pw_uid = 1000
      self.pw_gid = 1000
      self.pw_dir = '/home/kinow/'
      self.pw_shell = '/bin/bash'
      self.pw_gecos = any_value
      self.pw_name = any_value
      self.pw_passwd = 'x'
   
    def __repr__(self):
      return str(self.__dict__)

  return Obj()
    
pwd.getpwnam = mocked_fn

c.Authenticator.delete_invalid_users = True
c.Authenticator.admin_users = {
  'kinow'
}
c.JupyterHub.api_tokens = {
  'knockknock': 'kinow'
}
c.Application.log_level = 'DEBUG'

c.UIServer.user_authorization = {
    "cylc": ["READ", "pause", "trigger", "message"], 
    "group:group1": ["ALL"],
    "user2": ["READ", "CONTROL", "!trigger", "!edit"],
    "user3": ["READ", "stop", "play"],
    "user4": ["!ALL"]                                    
}
```

JupyterHub uses the `pwd` module to fetch the current user (we use `getpass`, but I think in the end both should access the same system resources and both hopefully work with LDAP/etc). Monkey-patching it and using the `DummyAuthenticator` with no password, that means I can log in as anything.

All the workflows are loaded from the same folder. One strange behaviour is that I can see all workflows, but upon start, the workflow disappears from the list of other users, and only my own user shows the workflow (we are probably filtering it somewhere in UIS), but that doesn't interfere with my testing here I think.

So my `kinow` user has:

image

And the `cylc` user (does not exist in my OS) has:

image

Trying to access my `kinow|five` from this `cylc` user, but no luck.

image

image

So at least I know the new authorization code is working. Now trying to get the hang of the schema and enable/disable permissions.

kinow:
Hmmm, still no luck, when you are back @datamel (no hurry, please enjoy the time off! :beach_umbrella: ) could you help me understand why it's not working, please?

I have the real `kinow` user, starting the hub with

* site config

```
# File: /etc/hub/config.py
c.UIServer.site_authorisation = {
  "*": {
    "*": {
      "default": ["READ"]
    },
    "cylc": {  
      "default": ["READ"]
    }
  }
}
```

* User config (**NB**: I used `user_authorization` as in the top comment of this PR, but it doesn't match the value in the global site config?)

```
# File: /home/kinow/.cylc/hub/config.py
c.UIServer.user_authorization = {
    "cylc": ["ALL"]                                 
}
```

But when I access the Hub with `cylc` and access `kinow|five` I still get 403. :confused: 

Tried with a real user (`cylc1`), and removing the monkey-patch config, same result:

image

cylc/uiserver/handlers.py
```diff
@@ -122,7 +78,6 @@ def set_default_headers(self) -> None:
         self.set_header("Content-Type", 'application/json')
 
     @web.authenticated
-    @authorised
```
oliver-sanders, line 125: This endpoint should be authorised.

The information it contains is not especially sensitive, especially given the authenticated requirement, however, it's still data you wouldn't expect another user to be able to see.

datamel:
Hi @kinow, 
I have a couple of suggestions to try to get this working....

1. Change the  `c.UIServer.site_authorisation` to  `c.UIServer.site_authorization`

With the example above, the `c.UIServer.site_authorisation` would have no impact and `c.UIServer.site_authorization` would be empty (from the uiservers point of view) which means that you would get the 403.

Once this is fixed (apologies for the mix-up with spelling, I will post on element now I am back to get confirmation of spellings), the config you have created will still not work to allow `ALL` operations. This is because there is no set `limit` in the site config, which means that the limit falls back to the default, i.e. READ operations. 

2. Add `limit` to the `site_authorization`

In order for user `cylc` to be allowed `ALL` operations (as intended by the user config) the site config limit must match (or exceed) the user config. Limits tell the uiserver that you (user kinow) is permitted to give away operations (in this case `ALL`) to user `cylc`. 

> * site config
> 
> ```
> # File: /etc/hub/config.py
> c.UIServer.site_authorisation = {                        <- change this to c.UIServer.site_authorization
>   "*": {
>     "*": {
>       "default": ["READ"]
>     },
>     "cylc": {  
>       "default": ["READ"],
         "limit": "ALL"                                <- try adding this (need the comma on the line above too)
>     }
>   }
> }
> ```

> # File: /home/kinow/.cylc/hub/config.py
> c.UIServer.user_authorization = {
>     "cylc": ["ALL"]                                 
> }
> ```

Fingers crossed this will be good to go now. 

If these suggestions fail, I can give you a call on element this evening/your morning to see if I can get you up and running with this.

kinow:
Thanks @datamel ! I will try that first thing in the morning and shout for help if I get stuck again :sweat_smile: There are some conflicts too due to recent PR's merged BTW :+1:

datamel:
> Thanks @datamel ! I will try that first thing in the morning and shout for help if I get stuck again 😅 There are some conflicts too due to recent PR's merged BTW 👍

Great, I'm just rebasing now. Very happy to help getting this tested, feel free to ping me a message on element and I can call if it is easier.
Will also add auth to handlers today, in response to @oliver-sanders comment.

kinow (commented):
Took a bit longer to test it, sorry @datamel .

**#1**

>With the example above, the c.UIServer.site_authorisation would have no impact and c.UIServer.site_authorization would be empty (from the uiservers point of view) which means that you would get the 403.

As predicted! As user `cylc`, trying to access a workflow of `kinow`, 403 error.

image

There is one issue I think, where I am able to figure out whether a user + workflow exists or not:

- `http://localhost:8000/user/kinow/#/workflows/five` is an existing user and an existing workflow, giving me 403
- `http://localhost:8000/user/kinow/#/workflows/six` is an existing user and an non-existing workflow, giving me also 403
- `http://localhost:8000/user/blabla/#/workflows/five` is a non-existing user, which redirects me to my page

So that way I can confirm whether a user exists or not, by simply using any workflow name, with a user that I think has an account in the server. - https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/03-Identity_Management_Testing/04-Testing_for_Account_Enumeration_and_Guessable_User_Account

**#2**

>In order for user cylc to be allowed ALL operations (as intended by the user config) the site config limit must match (or exceed) the user config. Limits tell the uiserver that you (user kinow) is permitted to give away operations (in this case ALL) to user cylc.

Hmmm, this didn't work.

```
# File: /home/kinow/.cylc/hub/config.py
c.UIServer.user_authorization = {
  "cylc": ["ALL"]                                 
}


# File: /etc/cylc/hub/config.py
c.UIServer.site_authorisation = {
  "*": {
    "*": {
      "default": ["READ"]
    },
    "cylc": {  
      "default": ["READ"],
        "limit": "ALL"
    }
  }
}
```

After stopping UIS, then starting it again, I got the same 403 response?

```bash
[I 2021-08-03T14:54:21.174 JupyterHub log:189] 200 GET /hub/api/authorizations/token/[secret] (kinow@127.0.0.1) 10.90ms
2021-08-03 14:54:22,112 cylc.uiserver.authorise INFO     cylc: requested read
2021-08-03 14:54:22,113 cylc.uiserver.authorise INFO     cylc: not authorised to read
2021-08-03 14:54:22,113 cylc.uiserver.authorise WARNING  Authorization failed for cylc:requested to read.
2021-08-03 14:54:23,088 cylc.uiserver.authorise INFO     cylc: requested read
2021-08-03 14:54:23,089 cylc.uiserver.authorise INFO     cylc: not authorised to read
2021-08-03 14:54:23,089 cylc.uiserver.authorise WARNING  Authorization failed for cylc:requested to read.
2021-08-03 14:54:23,113 cylc.uiserver.authorise INFO     cylc: requested read
2021-08-03 14:54:23,113 cylc.uiserver.authorise INFO     cylc: not authorised to read
2021-08-03 14:54:23,113 cylc.uiserver.authorise WARNING  Authorization failed for cylc:requested to read.
```

Any idea what I could be missing @datamel ? 

```bash
(venv) kinow@ranma:~/Development/python/workspace/cylc-uiserver$ git log -n 1
commit 4a9bdd9d07cea124b73ebd91d5ebfafaff4f1797 (HEAD -> pr-204)
Author: Mel Hall <37735232+datamel@users.noreply.github.com>
Date:   Mon Aug 2 11:25:10 2021 +0100

    BK review response

```

Thanks!!!

datamel:
Hi @kinow, thanks for all the testing so far.... 

> 
> Hmmm, this didn't work.
> 
> ```
> # File: /home/kinow/.cylc/hub/config.py
> c.UIServer.user_authorization = {
>   "cylc": ["ALL"]                                 
> }
> 
> 
> # File: /etc/cylc/hub/config.py
> c.UIServer.site_authorisation = {
>   "*": {
>     "*": {
>       "default": ["READ"]
>     },
>     "cylc": {  
>       "default": ["READ"],
>         "limit": "ALL"
>     }
>   }
> }
> ```

The spelling is still wrong on the `c.UIServer.site_authorisation`. It should be  `c.UIServer.site_authorization`.

Unfortunately, spelling errors in config pass silently without any warning at all - this is giving the impression of an empty site_authorization config and so "kinow" does not have the permission to give away access to "cylc".

Try that and see if it works.  Failing that could you could pop a breakpoint in `build_owner_site_auth_conf()` and see what is coming into `site_auth_config`?

oliver-sanders:
> Unfortunately, spelling errors in config pass silently without any warning at all

Once we go Jupyter Server we will get a warning.

datamel:
> There is one issue I think, where I am able to figure out whether a user + workflow exists or not:
> 
> * `http://localhost:8000/user/kinow/#/workflows/five` is an existing user and an existing workflow, giving me 403
> * `http://localhost:8000/user/kinow/#/workflows/six` is an existing user and an non-existing workflow, giving me also 403
> * `http://localhost:8000/user/blabla/#/workflows/five` is a non-existing user, which redirects me to my page
> 
> So that way I can confirm whether a user exists or not, by simply using any workflow name, with a user that I think has an account in the server. - https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/03-Identity_Management_Testing/04-Testing_for_Account_Enumeration_and_Guessable_User_Account
> 

Yes, good spot! I guess we have two options here.
- If someone does not have read access, then redirect them to their own page 
or
- If trying to go to a non-existent user's uiserver, raise a 403. 

I think the first option is probably preferable.

oliver-sanders:
> Yes, good spot! I guess we have two options here.

(I've added some more stuff to my previous comment, this might not be an issue)

I think this problem may be beyond the scope of the UIServer, especially once we go Jupyter Server and the Cylc endpoints are no longer the only endpoints.

kinow:
>The spelling is still wrong on the c.UIServer.site_authorisation. It should be c.UIServer.site_authorization.

Argh, I first fixed the spelling and tested to get the 403. Then I copied the block from another file I had created with the spelling... trying it now.

...

cylc/uiserver/authorise.py
datamel: For the duplicated keys, currently the 2nd "default" will override the first. I think this is probably acceptable to document? 

I have left a `TODO` for more rigorous dictionary validating in `_check_site_auth_dict_correct_format`, we could possibly warn about duplicates and non-existent permissions here. I wanted to do more reading on validating traitlets before embarking on this one!

Passing in non-existent permissions is I guess not a big deal in terms of security (any operations executed by users are checked against our schema mutations, so nothing can slip through here) but it could be frustrating for users if, for example, a typo and they intend to give permissions which are not enacted.

This leads me back to thinking that it would be nice to have an interface whereby users could check what permissions a user has for their uiserver, so, put in a username and it generates a list of permissions for that user. I think this could be helpful due to the support for groups. However, I'm not sure how widely used this feature will be...possibly not worth the time investment!

kinow (approved):
Managed to get it to block another user accesing my `kinow|five`, and then to later share the workflow (after much flailing and head-banging, thanks @datamel !!! :100: ).

Already know a few things we will need to update the UI so that we can access workflows from other users.

As for the OWASP thing I posted, I agree it's not really important since only logged in users will be able to use it (quite sure if we get someone pen-testing it again that might be raised since it's a common test for authx, but we can just explain it's really low risk).

+1, up to @dwsutherland and others now. Thanks again @datamel !

cylc/uiserver/authorise.py
datamel: I can see adding `user:` would be very clear. For info, this was discussed and was ruled out as it is additional input for the user/site which is technically unnecessary.

cylc/uiserver/authorise.py
kinow: >This leads me back to thinking that it would be nice to have an interface whereby users could check what permissions a user has for their uiserver, so, put in a username and it generates a list of permissions for that user. I think this could be helpful due to the support for groups. However, I'm not sure how widely used this feature will be...possibly not worth the time investment!

This tool does sound useful. Maybe create an issue here or in cylc/cylc-ui with low priority?

cylc/uiserver/authorise.py
kinow: I've mocked and created a non-existing user. Anything that returns the user string name should work, including the SSSD/LDAP created users :+1: marking as resolved.

datamel:
> But one thing I couldn't find in the docs, I wonder if you could help me… I saw that we have a site and a user config. So I imagined myself as a system admin looking after a machine with several (100s?) users.
> 
> All users having ability to start their own hub (something like `module load cylc` or similar). Now, maybe I don't want users to start Cylc because of GraphQL queries consuming resources, or some other concern.
> 
> Maybe I could limit the module loading, but users could still use Conda to install perhaps. That got me thinking in **trying the site configuration to lock my own account**. i.e. set a default to `!ALL` or something like that.

I think you have probably this sorted now, but just in case, to block all users - from a site level you would set a `limit` on the permissions the uiserver owner can give away to the access_user, for all users...
```
c.UIServer.site_authorization = {
    "user": {  # user whose uiserver you wish to block access to
        "*": {  # apply this setting to all authenticated users who will try access
            "limit": "[!ALL]"  # "user" is not allowed to grant "*" any permissions
        } 
    }
}

```

`default` is only used for the case where the access_user does not appear in the ui server owner's user config, in which case access falls back to what the site intended.

To block all users from your own uiserver config settings, set 
`c.UIServer.user_authorization= {"*":["!ALL"]}`

This configurable authorization is limited to allowing/denying users access to ui-servers.

cylc/uiserver/handlers.py
datamel: Authorization now added to Main handler and UserProfile handler - https://github.com/cylc/cylc-uiserver/pull/204/commits/baa081a35c58b7716a386c7ecbb3a8934cb8403c
These are now authorized on a `read` access mutation basis (configurable on a per user basis with the same `site_authorization` and `user_authorization` as used in the Middleware).  @kinow, not sure if you want to have a glance at this commit or if you are happy leaving it to @dwsutherland and @oliver-sanders.

datamel:
In order to get configurable authorization working with Jupyter server, I am reviewing @oliver-sanders #230 PR with the aim that it is merged before this one. 

I will then do the rebase of the master with the Jupyter Server work and fix any broken functionality.

So, for reviews of this PR, @dwsutherland and @oliver-sanders, priority is to get #230 in first, and may as well hold fire on the review of this one until after the rebase has gone in.

dwsutherland:
I assume all multi-user interaction with a user's UIS will be through the hub/UIS and not directly (via the CLI/zeroMQ) with the scheduler? i.e. only the UIS owner can interact directly with the Scheduler

If not, then the Scheduler will probably need to use the same authorisation Middleware as the UIS, and we usually import `cylc-flow` -> `cylc-uiserver` .. So we would need to shift that middleware to `cylc/flow/network/graphql.py`

If so, please excuse my ignorance and this comment.

(if there's still time, I'd still like to review this!)

datamel:
@dwsutherland sorry I hadn't noticed this comment in my notifications.
> If not, then the Scheduler will probably need to use the same authorisation Middleware as the UIS, and we usually import `cylc-flow` -> `cylc-uiserver` .. So we would need to shift that middleware to `cylc/flow/network/graphql.py`

I had noticed the IgnoreFieldMiddleware was in `cylc-flow` and double checked with @oliver-sanders at the time, all multi user will go through the hub. We will be passing metadata back to the scheduler with info on the user performing the action.  

> (if there's still time, I'd still like to review this!)
Yes please! Sorry that this PR has been on hold for the Jupyter-Server work but after quite a process of rebasing it now is back in business. I have a broken test (which appears to be flaky - I can get it both passing and failing locally, I'll look more tomorrow.). The functionality should be there 🤞 .

oliver-sanders:
One test left to fix: `cylc/uiserver/tests/test_auth.py::test_unauthenticated[cylc/graphql]`

datamel:
> One test left to fix: `cylc/uiserver/tests/test_auth.py::test_unauthenticated[cylc/graphql]`

Yes, I am also getting some bizarre behaviour from some of them locally, trying to pinpoint the exact problem at the moment.

cylc/uiserver/app.py
```diff
@@ -101,6 +106,87 @@ class CylcUIServer(ExtensionApp):
...
+        config=True,
+        help='''
+            Dictionary containing authorized users and permission levels for
+            authorization. Use this setting to share control of your workflows
+            with other users.
+            Note that you are only permitted to give away permissions up to
+            your limit for each user, as defined in the site_authorization
+            configuration.
+            ''' + AUTH_DESCRIPTION + '''
+            Example configuration, residing in `~/.cylc/hub/config.py`:
```
kinow, line 178: `jupyter_config.py` now?

cylc/uiserver/authorise.py
```diff
@@ -0,0 +1,501 @@
...
+        "releaseholdpoint",
+        "reload",
+        "remove",
+        "resume",
+        "setgraphwindowextent",
+        "setholdpoint",
+        "setoutputs",
+        "setverbosity",
+        "stop",
+        "trigger",
```
kinow, line 117: Do we have a way to generate these lists dynamically?

cylc/uiserver/authorise.py
```diff
@@ -0,0 +1,501 @@
...
+        self.owner_dict = self.build_owner_site_auth_conf()
+        self.get_permitted_operations = (
+            lru_cache(maxsize=128)(self._get_permitted_operations))
+
+    @staticmethod
+    def set_auth_conf(auth_conf: Union[LazyConfigValue, dict]) -> dict:
+        """Resolve lazy config where empty
+
+        Args:
+            auth_conf: Authorization configuration from a config.py
```
kinow, line 139: `jupyter_config`?

dwsutherland:
I'm having much the same trouble as Bruno.. Two users `sutherlander` & `bebop`
I can log and view a users own UIS with:
`/etc/cylc/`
```
print('SITE CONFIG PICKED UP! ---------')

c.CylcUIServer.ui_build_dir = '/opt/cylc/cylc-ui/dist'
c.UIServer.ui_build_dir = '/opt/cylc/cylc-ui/dist'

c.CylcUIServer.site_authorization = {
  "*": {
    "*": {
      "default": ["READ"]
    },
    "bebop": {  
      "default": ["READ"],
        "limit": "ALL"
    }
  }
}
```
In `/home/sutherlander/.cylc/hub/jupyter_config.py`:
```
print('USER CONFIG PICKED UP! ---------')
c.CylcUIServer.ui_build_dir = '/opt/cylc/cylc-ui/dist'

    "bebop": ["ALL"],
}
```
And root user having:
```
print('ROOT CONFIG PICKED UP! ---------')

c.Application.log_level = 'DEBUG'
# WHERE IS THE UI?
c.CylcUIServer.ui_build_dir = '/opt/cylc/cylc-ui/dist'
#c.UIServer.ui_build_dir = '/opt/cylc/cylc-ui/dist'

# FOR TESTING BEFORE 8.0 RELEASE (did not work for an alpha user...)
# c.JupyterHub.bind_url = 'http://127.0.0.1:8000/cylc'
# FOR TESTING ONLY!
#c.JupyterHub.authenticator_class = 'jupyterhub.auth.DummyAuthenticator'
c.Authenticator.delete_invalid_users = True
c.Authenticator.admin_users = {
  'sutherlander'
}
c.JupyterHub.api_tokens = {
  'knockknock': 'sutherlander'
}
# HACK FOR AUTHORIZATION
#import pwd

#def mocked_fn(any_value):
#  class Obj(object):
#    def __init__(self):
#      self.pw_uid = 1000
#      self.pw_gid = 1000
#      self.pw_dir = '/home/sutherlander/'
#      self.pw_shell = '/bin/bash'
#      self.pw_gecos = any_value
#      self.pw_name = any_value
#      self.pw_passwd = 'x'
#
#    def __repr__(self):
#      return str(self.__dict__)
#
#  return Obj()

#pwd.getpwnam = mocked_fn

# FOR SSL TESTING
# c.JupyterHub.ssl_key = '/home/kinow/.cylc/hub/kinow.key'
# c.JupyterHub.ssl_cert = '/home/kinow/.cylc/hub/kinow.crt'
# c.JupyterHub.internal_ssl = True

# File: /home/kinow/.cylc/hub/config.py
# AUTHORIZATION
#c.UIServer.user_authorization = {
#  "cylc": ["!ALL"]
#}

    "sutherlander": ["ALL"],
    "bebop": ["ALL"],
}
```

I can log into them:
image

But when I paste in the URL of `sutherlander` into `bebop` private window I first get:
image

Then a blank page:
image

And this in the UIS log:
```
[I 2021-09-22T17:43:42.296 JupyterHub log:189] 200 GET /hub/api/authorizations/token/[secret] (sutherlander@127.0.0.1) 5.31ms
[I 2021-09-22 17:43:42.296 CylcHubApp auth:992] Logged-in user {'kind': 'user', 'name': 'bebop', 'admin': False, 'groups': [], 'server': '/user/bebop/', 'pending': None, 'created': '2021-09-22T03:00:42.605201Z', 'last_activity': '2021-09-22T05:43:42.293919Z', 'servers': None}
[I 2021-09-22 17:43:42.297 CylcHubApp log:189] 302 GET /user/sutherlander/oauth_callback?code=[secret]&state=[secret] -> /user/sutherlander/cylc/ (@::ffff:127.0.0.1) 27.94ms
[I 2021-09-22 17:43:42.450 CylcHubApp log:189] 101 GET /user/sutherlander/cylc/subscriptions (bebop@::ffff:127.0.0.1) 0.63ms
[W 2021-09-22 17:43:42.484 CylcHubApp handlers:128] Authorisation failed for bebop
[W 2021-09-22 17:43:42.497 CylcHubApp log:189] 403 GET /user/sutherlander/cylc/userprofile (bebop@::ffff:127.0.0.1) 13.58ms
WARNING:cylc.uiserver.authorise:Authorization failed for bebop:requested to read.
[I 2021-09-22 17:43:42.521 CylcHubApp log:189] 200 POST /user/sutherlander/cylc/graphql (bebop@::ffff:127.0.0.1) 23.29ms
```

Any ideas what I'm doing wrong?

dwsutherland (approved):
Functionally:
- Works in restricting access for non-authorized user actions.
- Works on site level without site/root/user config on a user only basis.
- multi-user issues can be punted to follow ons.

Code; followed most of it, couldn't spot any real problems/holes.

Good job 👍

cylc/uiserver/authorise.py
datamel: I think this is a great idea. The mutations could each have the correct auth access group flag and get sent from cylc-flow. I tries this initially and got into a bit of a tangle but I think it is worth pursuing. I will put up a follow up issue for it.

oliver-sanders:
> But when I paste in the URL of sutherlander into bebop private window I first get:

That's as expected, click the "Authorise" button. JupyterHub can work with OAuth systems incl Google, GitHub & OpenID so this stage is for "authorising the authenticator".

oliver-sanders:
> Any ideas what I'm doing wrong?

Not sure, I haven't really played about with the dummy authenticator, we've got a setup with a test user account so we can work with live authorisation.

Tested again, working fine for me, here's my setup:

**User "A"**

`.cylc/hub/jupyter_config.py`
```python
    # allow all users to give away all permissions
    c.CylcUIServer.site_authorization = {    
        '*': {
            '*': {
                'default': [],
                'limit': ['ALL']
            }   
        }   
    }   
    # give user "B" read permissions
        'B': ['READ']
    }
```
Launch the hub:
```console
$ cylc hub
```
Navigate to `https://<hub>/user/a` and authenticate as user "A".

**User "B"**

In a second browser or private tab navigate to the hub and navigate to `https://<hub>/user/a`, authenticate as user "B" and authorise JupyterHub.

Try navigating to a workflow, you should be able to see it, there are currently dev print statements in the authorisation code so you should see messages like this:

```
B: requested read
B: authorized to read
```

Try stopping the workflow you should see messages like this:

```
B: requested stop
B: not authorised to stop
```

And a message in the UI:

Screenshot from 2021-09-22 16-33-28

cylc/uiserver/authorise.py
oliver-sanders: I think this should do the trick:

```python
import graphene

from cylc.uiserver.schema import UISMutations

[
    attr
    for attr in dir(UISMutations)
    if isinstance(getattr(UISMutations, attr), graphene.Field)
]
```

ATM only read mutation is `ping` so we can hardcode that part.

In the long run perhaps we could define a graphene `Meta` field e.g.

```python
class Hold(Mutation, TaskMutation):                                                  
    class Meta:
        description = sstrip('''
            Hold tasks within a workflow.
        ''')
        access = 'CONTROL'

```

oliver-sanders (commented):
Got my test setup sorted today and ran some basic tests, will move onto some heavier testing tomorrow.

Some ReStructuredText comments...

dwsutherland:
> Not sure, I haven't really played about with the dummy authenticator, we've got a setup with a test user account so we can work with live authorisation.

Neither have I, I was using the inbuilt (pam) authenticator.

dwsutherland:
@oliver-sanders - Great! that worked:
image

However, on a site level, I would have thought the root user could also define who can interact with an owner's workflows (not just the owner).. We should be able to have this section:
```
c.CylcUIServer.site_authorization = {
  "*": {
    "*": {
      "default": [],
      "limit": ['ALL']
    }
  }
}
```
defined by the privileged hub user (i.e. `/root/.cylc/hub/jupyter_config.py` and `/etc/cylc/hub/jupyter_config.py`)..

My setup was:
- `root` running the hub
- `sutherlander` has a workflow installed
- `bebop` no workflows, but interacting with `sutherlander` workflows.

The minimum I had to put at the hub/root level was:
```
    "sutherlander": ["ALL"],
    "bebop": ["ALL"],
}
```
So they could see & interact with their own workflows..

datamel:
@dwsutherland and @kinow, for info I have pushed a logging change to help diagnose the problem. Perhaps the site config is not getting picked up correctly and you are getting blocked? 

When you connect to the second user, if you look out for: `User username authorized permissions: ...` in the log. This will determine if the settings are getting through correctly.

On a side note, I'm just responding to suggestions from @oliver-sanders  and will get CI passing asap.

cylc/uiserver/app.py
```diff
@@ -101,6 +106,87 @@ class CylcUIServer(ExtensionApp):
...
+        help='''
+            Dictionary containing site limits and defaults for authorization.
+            This configuration should be placed only in the site set
+            configuration file and not the user configuration file (use
+            ``c.CylcUIServer.user_authorization`` for user defined
+            authorization).
+            If this configuration is empty, site authorization defaults to no
+            configurable authorization and users will be unable to set any
+            authorization.
+        ''' + AUTH_DESCRIPTION + '''
```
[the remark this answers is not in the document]
datamel: I have put it back to `+ AUTH_DESCRIPTION +`, the f string is breaking with the dictionary brackets.

oliver-sanders:
> However, on a site level, I would have thought the root user could also define who can interact with an owner's workflows (not just the owner

Yep, that should work, not sure why it isn't, will take a look at the loading of the site config.

(If you start the server with the `--debug` option, Jupyter Server will log any changes to the configuration made after loading each configuration file which can be useful for debugging.)

wxtim:
I can't give a thorough review, but I have just used the UI served from @datamel 's virtual desktop to start a workflow. I was not able to pause the workflow because she had configured my authorization to prevent me doing this.

kinow:
Ah! @datamel could it that the file was loaded, but I accidentally used `UIServer`, instead of `CylcUIServer`???

>c.UIServer.site_authorization = {

datamel:
@hjoliver I have just pushed changes from @oliver-sanders review. Should be good to go.

hjoliver:
@datamel note I commited a small syntax fix to your branch to make the tests pass (I had ommited a comma in a code change suggestion of mine that you commited, I think).

PR #210, opened 2021-04-16
Revert "[failure] Prepare release: 0.4.0"

oliver-sanders:
Reverts cylc/cylc-uiserver#209

Release failed due to missing PyPi token, reverting to go through the process again at a later date.

(note the tag has not been created and the milestone has not been closed).

> **Note:** We should bump the cylc-flow version to b1 due to the host/platform change before releasing.


MetRonnie (approved):
Might be possible to just re-run the workflow? https://github.com/cylc/cylc-uiserver/pull/209/checks

But if we're going to push a change to setup.py to bump the cylc-flow version, then I guess we shouldn't re-run that

hjoliver (approved):

PR #212, opened 2021-04-25
expanded workflow status deltas

dwsutherland:
These changes partially address https://github.com/cylc/cylc-ui/pull/543

Sibling to https://github.com/cylc/cylc-flow/pull/4206
(that one should go in first)

This PR:
- Adds statuses `installed`/`uninstalled` as seen at the UI Server in the form of deltas.
- Waits until all deltas are send to subscribers on unregistering the workflow and before purging.

Needed for the UI to know when it can remove old/uninstalled workflows.

From Element (Oliver) (note: `held => paused`):
> There are a number of other exotic state transitions that are possible between scans (ran into this with the scan work).
> 
> Here are the rules I would expect and the full state transition matrix (basically the same as the [scan code](https://github.com/cylc/cylc-uiserver/blob/831a528a4c05faa1136a125cc7d43138c2fc140a/cylc/uiserver/workflows_mgr.py#L246-L264)).
> ## The Flow States
> - none (not installed)
> - held (running, task pool held)
> - running (running, task pool unheld)
> - stopping (running, will shutdown soon)
> - stopped (previously run, no active scheduler)
> ## Delta Rules
> 
> <before> -> <after>
> When the state changes from <before> to <after> (over any time period)
> the UI would expect to receive...
> 
> none -> *
> An added delta.
> - -> none
> A pruned delta.
> 
> {held,running,stopping} -> stopped
> A shutdown and an update delta.
> 
> {held,running,stopping,stopped} -> {held,running,stopping,stopped}
> An update delta.
> 
> *(UUID1) -> *(UUID2)
> (scheduler UUID has change i.e. flow has been restarted)
> A shutdown? and an update delta.
> ## All State Transitions
> 
> <before> -> <after>
> (description)
> - delta component 1
> - delta component 2
> - ...
> 
> none -> stopped
> (installed)
> - added(state=stopped)
> 
> none -> held
> (installed and run in held mode)
> - added(state=held)
> 
> none -> running
> (installed and run)
> - added(state=running)
> 
> none -> stopping
> (installed, run and requested to stop)
> - added(state=stopping)
> 
> held -> none
> (was running, deleted)
> - shutdown
> - pruned
> 
> held -> running
> (held held, running)
> - updated(state=running)
> 
> held -> stopping
> (requested to stop)
> - updated(state=stopping)
> 
> held -> stopped
> (stopped)
> - shutdown
> - updated(state=stopped)
> 
> running -> none
> (was running, deleted)
> - shutdown
> - pruned
> 
> running -> held
> (paused)
> - updated(state=held)
> 
> running -> stopping
> (requested to stop)
> - updated(state=stopping)
> 
> running -> stopped
> (shutdown)
> - shutdown
> - updated(state=stopped)
> 
> stopping -> none
> (was running, deleted)
> - shutdown
> - pruned
> 
> stopping -> held
> (pending shutdown canceled)
> - updated(state=held)
> 
> stopping -> running
> (pending shutdown canceled)
> - updated(state=running)
> 
> stopping -> stopped
> (stopped)
> - updated(state=stopped)
> 
> stopped -> none
> (uninstalled)
> - pruned
> 
> stopped -> held
> (running, held mode)
> - updated(state=held)
> 
> stopped -> running
> (running)
> - updated(state=running)
> 
> stopped -> stopping
> (running, requested to stop)
> - updated(state=stopping)

To test this I ran the following:
```
#!/bin/bash

cylc install fox
sleep 6
rm -rf ~/cylc-run/fox
sleep 3
cylc install fox
sleep 6
cylc play --pause fox/run1
sleep 3
cylc play fox/run1
sleep 3
rm -rf ~/cylc-run/fox
sleep 1
pkill -f '/home/sutherlander/.envs/flow/bin/python /home/sutherlander/.envs/flow/bin/cylc play --pause fox/run1'
sleep 5
cylc install fox
sleep 4
cylc play --pause fox/run1
sleep 3
cylc play fox/run1
sleep 3
cylc stop fox/run1
sleep 8
while true; do
    if ! cylc ping $suite 2>/dev/null || \
       $(( $(cylc scan | wc -l) == 0 )); then
        echo 'workflow not running!!'
        break
    else
        echo 'workflow still running'
        sleep 3
    fi
done
# if the suite is in the process of·
sleep 4
rm -rf ~/cylc-run/fox
pkill -f '/home/sutherlander/.envs/flow/bin/python /home/sutherlander/.envs/flow/bin/cylc play --pause fox/run1'
```
workflow_status

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -60,6 +60,7 @@ class DataStoreMgr:
     INIT_DATA_WAIT_TIME = 5.  # seconds
     INIT_DATA_RETRY_DELAY = 0.5  # seconds
     RECONCILE_TIMEOUT = 5.  # seconds
+    PENDING_DELTA_CHECK_INTERVAL = 0.5
 
     def __init__(self, workflows_mgr):
         self.workflows_mgr = workflows_mgr
@@ -70,11 +71,13 @@ class DataStoreMgr:
         self.executors = {}
         self.delta_queues = {}
 
-    def update_contact(self, w_id, contact_data=None):
+    def update_contact(
+            self, w_id, contact_data=None, status=None, pruned=False):
         delta = DELTAS_MAP[ALL_DELTAS]()
         delta.workflow.time = time.time()
         flow = delta.workflow.updated
         flow.id = w_id
+        flow.stamp = f'{w_id}@{delta.workflow.time}'
         if contact_data:
             # update with contact file data
             flow.name = contact_data['name']
@@ -90,7 +93,12 @@ class DataStoreMgr:
             flow.port = 0
             # flow.pub_port = 0
             flow.api_version = 0
-            flow.status = WorkflowStatus.STOPPED.value
+
+        if status is not None:
+            flow.status = status
+        if pruned:
+            flow.pruned = True
+            delta.workflow.pruned = w_id
 
         # Apply to existing workflow data
         if 'delta_times' not in self.data[w_id]:
@@ -141,11 +149,23 @@ class DataStoreMgr:
         self.data[w_id] = data
 
         # create new entry in the delta store
-        self.update_contact(w_id)
+        self.update_contact(w_id, status=WorkflowStatus.INSTALLED.value)
+
+    async def unregister_workflow(self, w_id):
+        logger.debug(f'unregister_workflow({w_id})')
+        self.update_contact(w_id, pruned=True)
+        if w_id in self.delta_queues:
+            while any(
+                not delta_queue.empty()
+                for delta_queue in self.delta_queues[w_id].values()
+            ):
+                await asyncio.sleep(self.PENDING_DELTA_CHECK_INTERVAL)
+        self.purge_workflow(w_id)
 
     def stop_workflow(self, w_id):
         logger.debug(f'stop_workflow({w_id})')
-        self.update_contact(w_id)
+        self.purge_workflow(w_id, data=False)
+        self.update_contact(w_id, status=WorkflowStatus.STOPPED.value)
 
     def purge_workflow(self, w_id, data=True):
         """Purge the manager of a workflow's subscription and data."""
--- a/cylc/uiserver/workflows_mgr.py
+++ b/cylc/uiserver/workflows_mgr.py
@@ -213,14 +213,13 @@ class WorkflowsManager:
 
     async def _unregister(self, wid):
         """Unregister a workflow from the data store."""
-        self.uiserver.data_store_mgr.purge_workflow(wid)
+        await self.uiserver.data_store_mgr.unregister_workflow(wid)
 
     async def _stop(self, wid):
         """Mark a workflow as stopped.
 
         The workflow can't do this itself, because it's not running.
         """
-        self.uiserver.data_store_mgr.purge_workflow(wid, data=False)
         self.uiserver.data_store_mgr.stop_workflow(wid)
 
     async def update(self):
@@ -266,7 +265,8 @@ class WorkflowsManager:
             if before == 'active':
                 self.active.pop(wid)
             elif before == 'inactive':
-                self.inactive.remove(wid)
+                if wid in self.inactive:
+                    self.inactive.remove(wid)
             if after == 'active':
                 self.active[wid] = flow
             elif after == 'inactive':
--- a/cylc/uiserver/tests/test_workflows_mgr.py
+++ b/cylc/uiserver/tests/test_workflows_mgr.py
@@ -376,9 +376,11 @@ async def test_unregister(
     workflow from the data store attributes, and call the necessary
     functions."""
     workflow_name = 'unregister-me'
+    workflow_id = f'{getuser()}{ID_DELIM}{workflow_name}'
+    await uiserver.workflows_mgr._register(workflow_id, None)
 
     uiserver.workflows_mgr._scan_pipe = empty_aiter()
-    uiserver.workflows_mgr.inactive.add(workflow_name)
+    uiserver.workflows_mgr.inactive.add(workflow_id)
     # NOTE: here we will yield a workflow that is not running, it does
     #       not have the contact data and is inactive.
     #       This is what forces the .update() to call unregister()!
@@ -386,7 +388,7 @@ async def test_unregister(
     await uiserver.workflows_mgr.update()
 
     # now the workflow is not active, nor inactive, it is unregistered
-    assert workflow_name not in uiserver.workflows_mgr.inactive
+    assert workflow_id not in uiserver.workflows_mgr.inactive
 
 
 @pytest.mark.asyncio
@@ -400,7 +402,7 @@ async def test_connect(
     If a workflow is running, but in the inactive state,
     then the connect method will be called."""
     workflow_name = 'connect'
-    workflow_id = f'{getuser()}|{workflow_name}'
+    workflow_id = f'{getuser()}{ID_DELIM}{workflow_name}'
     uiserver.workflows_mgr.inactive.add(workflow_id)
 
     assert workflow_id not in uiserver.workflows_mgr.active


cylc/uiserver/data_store_mgr.py
```diff
@@ -141,11 +143,11 @@ async def register_workflow(self, w_id):
         self.data[w_id] = data
 
         # create new entry in the delta store
-        self.update_contact(w_id)
+        self.update_contact(w_id, status='installed')
```
kinow, line 146: Should we update the [`WorkflowStatus`](https://github.com/cylc/cylc-flow/blob/3537c161be43ddae7a1aa5373de14445ce159074/cylc/flow/workflow_status.py#L40) to include `installed` there too?

It's weird that we have `status=WorkflowStatus.STOPPED.value` (below) but not `status=WorkflowStatus.INSTALLED.value` here.

cylc/uiserver/workflows_mgr.py
```diff
@@ -213,7 +214,21 @@ async def _disconnect(self, wid):
...
+        store_mgr = self.uiserver.data_store_mgr
+        if wid in store_mgr.data:
+            store_mgr.update_contact(wid, status='uninstalled')
+        while True:
+            has_queues = []
+            if wid in store_mgr.delta_queues:
+                has_queues = [
+                    True
+                    for delta_queue in store_mgr.delta_queues[wid].values()
+                    if not delta_queue.empty()]
```
kinow, line 226: I think the above can be simplified with `any()`?

kinow (commented):
Only had time to review the code today. Going home earlier due to our evening meeting. Couple comments. One about `any`… I think we don't need the `has_queues` array. It could be a `bool` instead, and just be set to `True` if any `delta_queue` is not empty? Brain is in the javascript-mode, so I couldn't write Python to test it now, but can try it another day with :coffee: feel free to ignore if that doesn't make sense :+1:

cylc/uiserver/workflows_mgr.py
kinow: Best I could come up with, but not tested :laughing: just to show what I mean.

```py
has_queues = any(not delta_queue.empty for delta_queue in store_mgr.delta_queues[wid].values())
```

cylc/uiserver/data_store_mgr.py
dwsutherland: Good point, I'll actually subclass it at the UIS (unless there's some reason to have it at the Scheduler.

cylc/uiserver/workflows_mgr.py
dwsutherland: Oh yeah, true.

dwsutherland: What's faster; calling `any` or list comprehension? XD

cylc/uiserver/data_store_mgr.py
dwsutherland: Moved to `cylc.flow` .. Sibling PR https://github.com/cylc/cylc-flow/pull/4206

dwsutherland:
Tests won't pass until https://github.com/cylc/cylc-flow/pull/4206 is in.

cylc/uiserver/workflows_mgr.py
dwsutherland: Ok made it better, used that as the while condition.

oliver-sanders:
I don't quite get the `uninstalled` state, shouldn't it be a `pruned` delta?

kinow:
Using the Cylc UI GScan branch to test this PR now. The first thing I noticed was that the only workflows displayed using this branch are the running ones. Quickly debugging it, looks like the GScan filters isn't happy with the new states. I'll update that branch to work with this PR so that it displays the same as on Cylc UI's `master` branch first. Then will try to use the new states — which should fix the issue highlighted in that pull request, when a workflow is renamed/removed.

kinow:
Adding new states for the UI, with new icons. Chose folder-plus and folder-remove, but these icons may change later; we can also change states here later but updating the Cylc UI branch will be very simple.

image

kinow (commented):
@dwsutherland I had a workflow/run `five-running/run19` which I simply `rm -rf ~/cylc-run/five-running/run19`. Then I received two deltas:

```js
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"id": "kinow|five-running/run19", "shutdown": false, "added": {"workflow": {"id": "kinow|five-running/run19", "name": "five-running/run19", "status": "uninstalled", "owner": "kinow", "host": "", "port": 0, "__typename": "Workflow"}, "__typename": "Added"}, "updated": {"workflow": {"id": "kinow|five-running/run19", "__typename": "Workflow"}, "__typename": "Updated"}, "__typename": "Deltas"}}}}

{"id": "1", "type": "data", "payload": {"data": {"deltas": {"id": "kinow|five-running/run19", "shutdown": false, "added": {"workflow": {"id": "kinow|five-running/run19", "name": "five-running/run19", "status": "uninstalled", "owner": "kinow", "host": "", "port": 0, "__typename": "Workflow"}, "__typename": "Added"}, "updated": {"workflow": {"id": "kinow|five-running/run19", "__typename": "Workflow"}, "__typename": "Updated"}, "__typename": "Deltas"}}}}
```

They appear to have the exact same content. Any idea why it would result in two identical deltas for the same query (you can see it's the same query as it has the same UI subscription id `1`).

Also, I was expecting to see the workflow updated to have the state `uninstalled`. But instead I have the workflow added with the state `uninstalled`. Is that expected?

dwsutherland:
> Also, I was expecting to see the workflow updated to have the state uninstalled. But instead I have the workflow added with the state uninstalled. Is that expected?

No point in showing this state.

kinow:
So instead of the unistalled state we will get a pruned delta? I just need something yhat I can use to remove the workflow from gscan if it was deleted or renamed.

dwsutherland:
> So instead of the unistalled state we will get a pruned delta? I just need something yhat I can use to remove the workflow from gscan if it was deleted or renamed.

No, use it as the prune indicator.. but just don't show it.

kinow:
> > So instead of the unistalled state we will get a pruned delta? I just need something yhat I can use to remove the workflow from gscan if it was deleted or renamed.
> 
> No, use it as the prune indicator.. but just don't show it.

Aaaahh! D'oh moment :rofl:  will update that branch tomorrow handling the uninstalled state. Great spot David! :man_facepalming:

dwsutherland:
> I don't quite get the `uninstalled` state, once uninstalled there is nothing there to track, shouldn't it be a `pruned` delta?

It does make sense (in my head anyway) to say "the workflow is uninstalled" in the same way you'd say "the workflow is running/installed/paused"..

It essentially is the pruned delta (what's the difference?), and it is the state of the workflow... (we can put it anyway)
And besides, the `delta.pruned`/`delta` is only one type of subscription, so this way it'll show up in the gscan subscription also.

oliver-sanders:
> It essentially is the pruned delta (what's the difference?)

Once a workflow is uninstalled it doesn't exist any more so there is nothing for the UIS/UI to track. So there shouldn't be an entry in the UIS or UI data store for it any more. So I think we want to send a "pruned" message telling remote data stores that the workflow has been removed.

The UI shouldn't have to interpret the updated deltas to find out whether or not it needs to purge its data store:

```js
if ((delta.updated && delta.status && delta.status == WorkflowState.UNINSTALLED) || delta.pruned) {
    prune(delta.id)
}
```

hjoliver:
Looks good!

dwsutherland:
Changed the `delta.pruned` entry to be an ID:
image

oliver-sanders:
Run a few tests:

* The workflow pruned deltas are looking good!
* I haven't seen the workflow status flickering :+1:.
* There are still a couple of oddities in the deltas but I think this is for future work.

Here's the result of a test I performed using the following subscription:

```graphql
subscription {
  deltas {
    added {
      workflow(stripNull:true) {
        id
        status
      }
    }
    updated {
      workflow(stripNull:true) {
        id
        status
      }
    }
    pruned {
      workflow
    }
  }
}
```

Installing/playing/cleaning a workflow resulted in the following deltas:

```console
$ cylc install one --no-run-name
added(one, installed)

$ cylc play one
added(one, running)  # should be updated(one, running)
updated(one)  # empty delta not needed for this sub caused by task pool changes
updated(one, stopped)

$ cylc clean one
updated(one)  # should not be an updated delta here
pruned(one)
```

Here's the returned JSON in full:

```console
$ cylc install one --no-run-name
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {"workflow": {"id": "oliver|one", "status": "installed"}}, "updated": {}, "pruned": {}}}}}

$ cylc play one
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {"workflow": {"id": "oliver|one", "status": "running"}}, "updated": {}, "pruned": {}}}}}
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {}, "updated": {"workflow": {"id": "oliver|one"}}, "pruned": {}}}}}
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {}, "updated": {}, "pruned": {}}}}}
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {}, "updated": {"workflow": {"id": "oliver|one", "status": "stopped"}}, "pruned": {}}}}}

$ cylc clean one
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {}, "updated": {"workflow": {"id": "oliver|one"}}, "pruned": {"workflow": "oliver|one"}}}}}
```

hjoliver:
Tested with same result as @oliver-sanders 

I applied the same subscription above in GraphiQL, then watched the network response in Firefox Dev Tools, because you can't tell in GraphiQL if the same response comes back multiple times.
- we get an empty updated delta whenever the task pool changes, containing just the workflow ID. 
- another empty updated delta appears on removing the workflow, along with the correct pruned delta.

@dwsutherland - can the server strip / not send these empty (ID only) deltas?

kinow:
Just wondering if `stripNull` would help here @hjoliver ? In @oliver-sanders ' example query I think it doesn't have `stripNull: true` at the top level. Did you try that?

oliver-sanders:
> Just wondering if stripNull would help here

Worth a shot sticking it a level up on the deltas bit:

```graphql
subscription {
  deltas(stripNull:true) {
    added {
      workflow(stripNull:true) {
        id
        status
      }
    }
    updated {
      workflow(stripNull:true) {
        id
        status
      }
    }
    pruned {
      workflow
    }
  }
}
```

```console
$ cylc install one --no-run-name
added(installed)

$ cylc play one
added(running)       # should be updated(running)
updated()            # empty updated delta
null                 # empty delta
updated(stopped)

$ cylc clean one
updated()            # empty updated delta
pruned()
```

```console
$ cylc install one --no-run-name
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {"workflow": {"id": "oliver|one", "status": "installed"}}, "updated": {}, "pruned": {}}}}}

$ cylc play one
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {"workflow": {"id": "oliver|one", "status": "running"}}, "updated": {}, "pruned": {}}}}}
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {}, "updated": {"workflow": {"id": "oliver|one"}}, "pruned": {}}}}}
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {}, "updated": {}, "pruned": {}}}}}
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {}, "updated": {"workflow": {"id": "oliver|one", "status": "stopped"}}, "pruned": {}}}}}

$ cylc clean one
{"id": "1", "type": "data", "payload": {"data": {"deltas": {"added": {}, "updated": {"workflow": {"id": "oliver|one"}}, "pruned": {"workflow": "oliver|one"}}}}}
```

dwsutherland:
It's inherited so you can declare at the top:
```
subscription {
  deltas (stripNull: true) {
    added {
      workflow {
        id
        status
      }
    }
    updated {
      workflow {
        id
        lastUpdated
        status
      }
    }
    pruned {
      workflow
    }
  }
}
```
However, the reason added, updated, and pruned aren't stripped (yet) is because they aren't protobuf fields (this data structure is a mix of dictionary and protobuf)..

> @dwsutherland - can the server strip / not send these empty (ID only) deltas?

(as mentioned on riot) It should be the topic of a separate PR, this PR is only providing the `installed` and `pruned` messages to the client. The fixes included in this PR is to make sure it's a clean signal.

It may take some work to figure out how to make some graphql fields show up only when other fields are not empty.

Additionally, I don't think I can have strip null as an argument to the parent/root `subscription` or `query`..
Which leads to the next question; can we not push after the stripping? (i.e. we strip down to nothing to send)

hjoliver:
>  Which leads to the next question; can we not push after the stripping? (i.e. we strip down to nothing to send)

Presumably we do push after stripping?  (There's no point in pushing before stripping!)  ... I must be misunderstanding you?

hjoliver:
@dwsutherland so you've explained the empty (ID only) deltas.  But I don't think you commented on this?:

```
$ cylc install one --no-run-name
added(installed)     # <--- good

$ cylc play one
added(running)       # <---- should be updated(running)
```

kinow:
Started reviewing this one today with the sibling cylc-flow, and the UI gscan deltas branches, but it will take a few more days to complete the review & tests :+

dwsutherland:
> @dwsutherland so you've explained the empty (ID only) deltas. But I don't think you commented on this?:
> 
> ```
> $ cylc install one --no-run-name
> added(installed)     # <--- good
> 
> $ cylc play one
> added(running)       # <---- should be updated(running)
> ```

Tricky one, because to the scheduler it is `added`..
Hence, that `running` would have come through with the initial burst, any subsequent status changes will be in `updated`..

The only `status` changes in `added` should be on initial burst. 

So I'd suggest leaving it as is and have the UI treat this transition the same as a `reload` (where you'd also get a new `added` to overwrite/recreate the previous UI store)...

dwsutherland:
> > Which leads to the next question; can we not push after the stripping? (i.e. we strip down to nothing to send)
> 
> Presumably we do push after stripping? (There's no point in pushing before stripping!) ... I must be misunderstanding you?

Yes, what I meant was: If there's nothing to push after stripping, can we back out of yielding a result to the web socket?

oliver-sanders:
> So I'd suggest leaving it as is and have the UI treat this transition the same as a reload (where you'd also get a new added to overwrite/recreate the previous UI store)...

We should be able to turn that added into an updated before broadcast somehow?

Going to bump to a new issue to avoid convoluting this one - https://github.com/cylc/cylc-uiserver/issues/221

oliver-sanders:
So I'd suggest leaving it as is and have the UI treat this transition the same as a reload (where you'd also get a new added to overwrite/recreate the previous UI store)...
Wasn't aware of the added delta on reload (and I doubt the UI is either). We should be able to turn that added into an updated before broadcast somehow? Really don't want to have to make the UI interpret the information it's being provided in order to sync it's internal data store (the UI doesn't know much about Cylc).

Going to bump to a new issue to avoid convoluting this one - #221

oliver-sanders (approved):
Tested the pruned delta, works fine for me, bumped the other stuff into an issue for debate to allow us to focus on one thing at a time.

oliver-sanders:
> Yes, what I meant was: If there's nothing to push after stripping, can we back out of yielding a result to the web socket?

Is this in the async generator `while True` loop? If so I think `continue` would do it?

hjoliver:
@dwsutherland - a few tests failing here.

dwsutherland:
> @dwsutherland - a few tests failing here.

Yes, it will fail until the `cylc-flow` sibling is in.

hjoliver:
Oh, of course - sorry :face_with_head_bandage:

hjoliver (approved):
No problems here on re-testing. :+1:  Let's get this in ...

PR #213, opened 2021-04-29
load balanced spawner

### LINKED ISSUE #137: create a remote spawner with load-balancing capability
[discussed at CylcCon2020]

We want the Cylc Hub to be able to start UI Servers on any one of a group of hosts.

There is already [at least one] JupyterHub spawner which is able to start servers on remote hosts, the remote spawner.

We would like to add basic load-balancing to this. Cylc has load-balancing built-in, see the `cylc.flow.host_select` module.

The idea is to try and build a new JupyterHub spawner which handles load balancing to work out where to start the server, then uses the configured spawner (e.g. the remote spawner) to actually start the server.

This would give sites a lot of flexibility with which spawners they install.

As, depending on the nature of the installation this spawner may be running as a privileged user the code should be kept to a minimum.

oliver-sanders:
closes #137 

Turned out that having a spawner as a plugin to a spawner was a little tricky as there's all sorts of setup Jupyterhub does.

Simple solution, runs the command through an SSH connection, similar to how `cylc play` does it.

> **Note:** JupyterHub spawners run as the user so this spawner, the SSH connection and the UIS is spawns are all in userland even if the hub is not.

Here's the configuration to use it:

```python
c.JupyterHub.spawner_class = 'cylc.uiserver.distributed_spawner.DistributedSpawner'
c.DistributedSpawner.hosts = ['localhost']  
```

Unfortunately the UIS doesn't like being spawned onto other hosts, I think we need to patch the handlers with the location of the Hub. I've tried patching the `MainHandler` but that leaves me with the following error (Hub stdout):

```
tornado.application WARNING  oauth state '<key>' != None
```

So ATM this can't be used to spawn UIS on other hosts which makes it a bit of a chocolate teapot.

Been looking at the JupyterHub `single-user` thinggy trying to figure out how to fix it. They configure their handlers quite differently (redirect rather than render) and seem to provide `hub-host` and `hub-prefix` CLI opts?

https://github.com/jupyterhub/jupyterhub/blob/3fec19d191119e4203c243f44c43e74c8fe62f04/jupyterhub/singleuser/mixins.py#L141-L155

**Requirements check-list**

- [ ] Does not need tests (why?).

- [ ] No change log entry required (why? e.g. invisible to users).

- [ ] Created an issue at [cylc-uiserver conda-forge repository](https://github.com/conda-forge/cylc-uiserver-feedstock) with version changes (if you changed dependencies in `setup.py`, see `recipe/meta.yaml`).
- [ ] No dependency changes.


oliver-sanders:
Will rebase after #230 at which point this should work,

cylc/uiserver/distributed_spawner.py
```diff
@@ -0,0 +1,203 @@
...
+        self._proc = None
+
+    def choose_host(self):
+        return select_host(self.hosts, self.ranking)[1]
+
+    def get_env(self):
+        return {
+            **Spawner.get_env(self),
+            'CYLC_VERSION': CYLC_VERSION,
+            'JUPYTERHUB_SERVICE_PREFIX': '/user/osanders/'
```
datamel, line 124: Having a brief glance at this with #230, and not marked ready for review but this line will need changing.
Nice looking pr though, happy to review when it is ready.

oliver-sanders: Yeah this was hacked a bit, should be much nicer with Jupyter Server :crossed_fingers:.

oliver-sanders:
Rebased this now we are using Jupyter Server.

The spawner works, the server is started on the remote host and the polling works correctly, but, from the hub's URL I get a 503 error and from the servers URL, a 404 so it's still not useable.

I think it's missing some of the config/initialisation required to work with the hub's remote API? More investigation required.

PR #222, opened 2021-05-26
Update for cylc flags changes.

hjoliver:
Trivial fix post recent(ish) cylc-flow change.  One review will do.


kinow (approved):
Confirmed the flags value changed in Cylc Flow https://github.com/cylc/cylc-flow/blob/bdb51e2d8c7cc741a690a9aef396418f1fd619d2/cylc/flow/flags.py#L21

Looks good to me! :+1:

PR #234, opened 2021-08-11
data-store sync reconciliation fixed

dwsutherland:
These changes partially address https://github.com/cylc/cylc-ui/issues/667

The reason why the reload broke the UI is because the UIS data-store would get out of sync with reload, and the sync reconciliation was broken (causing an error to be thrown in the subscription thread).

This PR fixes the sync reconciliation.. Will also put a message in `--debug`, i.e,:
```
2021-08-11 12:43:32,864 cylc.uiserver.data_store_mgr DEBUG    Out of sync with jobs of sutherlander|five/run1... Reconciling.
```
A sibling PR (https://github.com/cylc/cylc-flow/pull/4349) fixes the broken sync.
Another sibling PR (https://github.com/cylc/cylc-ui/pull/729) changes what delta content the UI considers an initial burst.

**Requirements check-list**
- [x] Does need tests, too hard basket for now (as requires for Cylc stack).
- [x] No change log entry (bug fix).
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -228,6 +228,12 @@ class DataStoreMgr:
         self.apply_all_delta(w_id, delta)
         self.delta_store_to_queues(w_id, topic, delta)
 
+    def clear_data_field(self, w_id, field_name):
+        if field_name == WORKFLOW:
+            self.data[w_id][field_name].Clear()
+        else:
+            self.data[w_id][field_name].clear()
+
     def apply_all_delta(self, w_id, delta):
         """Apply the AllDeltas delta."""
         for field, sub_delta in delta.ListFields():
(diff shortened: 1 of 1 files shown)


cylc/uiserver/data_store_mgr.py
```diff
@@ -230,23 +230,31 @@ def update_workflow_data(self, topic, delta, w_id):
...
-                    self.data[w_id][field.name].clear()
+            if sub_delta.reloaded and False:
+                self.clear_data_field(w_id, field.name)
                 self.data[w_id]['delta_times'][field.name] = 0.0
-            # Apply the delta if newer than the previously applied.
-            if delta_time >= self.data[w_id]['delta_times'][field.name]:
-                apply_delta(field.name, sub_delta, self.data[w_id])
-                self.data[w_id]['delta_times'][field.name] = delta_time
-                self.reconcile_update(field.name, sub_delta, w_id)
+            # hard to catch errors in a threaded async app, so use try-except.
```
kinow, line 248: :+1:

kinow (approved):
Looks good to me! Tested with a little modification in Cylc UI, and now it appears to work fine! Thanks for fixing it so quickly @dwsutherland ! There are no tests included, but I think changes like this are not easily tested with unit/functional/etc tests.

hjoliver (approved):

kinow (approved):
Had another look at the code, looks good.

Using UI `master`, and UIS `master`, after `cylc reload five`, the tree view is blank.

image

Using this branch, the UI is also blank for a few seconds. But eventually I can see the messages:

```bash
[D 2021-09-09 16:33:32.124 CylcUIServer] Out of sync with family_proxies of kinow|five... Reconciling.
[D 2021-09-09 16:33:32.126 CylcUIServer] Out of sync with task_proxies of kinow|five... Reconciling.
[D 2021-09-09 16:33:32.128 CylcUIServer] Out of sync with edges of kinow|five... Reconciling.
[D 2021-09-09 16:33:33.132 CylcUIServer] Out of sync with jobs of kinow|five... Reconciling.
```

And after that the UI displays the tree with no issues. There's an issue in the UI for adding a no-state, or no-workflow-found, etc, view states. After that instead of a blank display we can show something else, like a loading bar, or a message that it's reloading the workflow, etc.

Thanks @dwsutherland !

dwsutherland:
>  There's an issue in the UI for adding a no-state, or no-workflow-found, etc, view states. After that instead of a blank display we can show something else, like a loading bar, or a message that it's reloading the workflow, etc.

This is addressed by they sibling (there's no blank after that's in)

PR #252, opened 2021-09-27
config: fix config file load order

oliver-sanders:
* Turns out Jupyter Server config_file_paths are loaded in reverse order.
* Fixes the CYLC_SITE_CONF_PATH env var override (which wasn't adding
  `/hub` to the path.

**Requirements check-list**
- [x] No changelog (bug has not been released).
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/__init__.py
+++ b/cylc/uiserver/__init__.py
@@ -15,6 +15,9 @@
 
 __version__ = "0.5.0"
 
+import os
+from typing import Dict
+
 from cylc.uiserver.app import CylcUIServer
 
 
@@ -29,3 +32,21 @@ def _jupyter_server_extension_points():
             'app': CylcUIServer
         }
     ]
+
+
+def getenv(*env_vars: str) -> Dict[str, str]:
+    """Extract env vars if set.
+
+    Returns a dict containing key:value pairs of environment variables
+    defined in env_vars providing they are present in the environment.
+
+    Examples:
+        >>> getenv('HOME', 'NOT_AN_ENVIRONMENT_VARIABLE')
+        {'HOME': ...}
+
+    """
+    env: Dict[str, str] = {}
+    for env_var in env_vars:
+        if env_var in os.environ:
+            env[env_var] = os.environ[env_var]
+    return env
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -252,10 +252,6 @@ class CylcUIServer(ExtensionApp):
         ioloop.IOLoop.current().add_callback(
             self.workflows_mgr.update
         )
-        ioloop.PeriodicCallback(
-            self.workflows_mgr.update,
-            self.scan_interval * 1000
-        ).start()
 
     def initialize_settings(self):
         """Update extension settings.
@@ -274,6 +270,11 @@ class CylcUIServer(ExtensionApp):
                 for key, value in self.config['CylcUIServer'].items()
             )
         )
+        # configure the scan
+        ioloop.PeriodicCallback(
+            self.workflows_mgr.update,
+            self.scan_interval * 1000
+        ).start()
 
     def initialize_handlers(self):
         self.handlers.extend([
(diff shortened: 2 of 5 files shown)


datamel (commented):
I am still having problems with getting config items to load... It is possible my test is flawed, I will detail it below.

- I have `CYLC_SITE_CONF_PATH` exported to: `~/etc/cylc` in my .bashrc,
- In the file `~/etc/cylc/hub/jupyter_config.py` I have `c.CylcUIServer.scan_interval = 60  # PT60S`
- In line 256 of `app.py`, I have added `print(f".!!!!!!!...scan interval is ...{self.scan_interval}")`
- I then run `cylc hub --debug`.
The output is `".!!!!!!!...scan interval is ...5.0"`, which is the default rather than the 60 I have set in the config file.

Can you reproduce this or spot any flaw in my test?

MetRonnie:
I can reproduce that. It looks like no exported env vars (including `CYLC_SITE_CONF_PATH`) are getting through to `app.py`? I put
```
print(f".!!!!!!!...ENV:\n{os.environ}")
```
in `__init__` instead of the scan interval

MetRonnie:
I used Mel's suggestion of putting
```python
c.Spawner.environment = {
    'CYLC_SITE_CONF_PATH': f'{os.getenv("CYLC_SITE_CONF_PATH")}'
}
```
in `~/.cylc/hub/jupyter_config.py` and now the `CYLC_SITE_CONF_PATH` is coming through successfully, however `self.scan_interval` is still `5.0`, not `60`

oliver-sanders:
Can't replicate that I'm afraid.

`c.Spawner` only applies in hub mode, you should be able to test this in hubless mode which reduces the number of moving parts.

I've added some default config to automatically pass through the relevant `CYLC_*` environment variables and added a debug statement in the CylcUIServer which lists all of the configuration that have been set to make testing a little easier.

Here's how I tested it:

```
.cylc/
├── hub
│   └── jupyter_config.py
└── site_conf
    └── hub
        └── jupyter_config.py
```

`.cylc/hub/jupyter_config.py`:

```python
c.CylcUIServer.scan_interval = 3
```

`.cylc/site_conf/hub/jupyter_config.py`:

```python
c.CylcUIServer.scan_interval = 42
```

```
$ cylc gui --debug
...
  * scan_interval = 3
$ CYLC_SITE_CONF_PATH=$HOME/.cylc/site_conf cylc gui --debug
...
  * scan_interval = 42
```

To test in hub mode I had to patch the `--debug` arg onto the spawner command for some reason:

```
$ cylc hub --debug
# refresh browser tab to spawn UIS
...
  * scan_interval = 3
$ CYLC_SITE_CONF_PATH=$HOME/.cylc/site_conf cylc hub --debug
# refresh browser tab to spawn UIS
...
  * scan_interval = 42
```

oliver-sanders:
> Can you reproduce this or spot any flaw in my test?

Ah, yes...

You were adding your print statement into the `__init__` method, however, the config has not yet been loaded at this stage (as the object loads the config for itself) so were always getting the default value.

This was perfectly sensible as the periodic callback was started there using the scan interval config which is a bug 😠.

oliver-sanders:
Hopefully that should do the trick.

MetRonnie:
This is working for me with `cylc gui --debug` as advertised. However, the `cylc hub --debug`  
>``` 
>$ cylc hub --debug
># refresh browser tab to spawn UIS
>...
>  * scan_interval = 3
>$ CYLC_SITE_CONF_PATH=$HOME/.cylc/site_conf cylc hub --debug
># refresh browser tab to spawn UIS
>...
>  * scan_interval = 42
>```
is not working for me -  I don't see the configuration printed out

oliver-sanders:
>  However, the cylc hub --debug

When I tried this the `--debug` context wasn't being carried through the spawner properly for some reason so I appended the `--debug` argument to `c.Spawner.command`.

> To test in hub mode I had to patch the --debug arg onto the spawner command for some reason:

(otherwise just bodge the `log.debug` to `log.info` 😁)

MetRonnie (approved):
Am happy this is working as expected

Issue #242, opened 2021-09-02
Ping broken on the gui


datamel:
This appears with master `cylc-uiserver`, `cylc-ui` and `cylc-flow` installed.
Steps to reproduce: 

- Start either `cylc gui`, `jupyter-cylc` or `cylc hub` - issue is present with all three.
- Play a workflow, useful to run it in `--no-detach` mode to see the error on screen. 
- From the gui `ping` the workflow (with the ping button on the menu or from GraphIQL:
```
mutation {
  ping (workflows: ["workflow_name"]) {
    result
  }
}
``` 

You get a traceback error of....

```
Traceback (most recent call last):
  File "~/miniconda3/envs/cylc-8/lib/python3.7/site-packages/promise/promise.py", line 844, in handle_future_result
    resolve(future.result())
  File "~/cylc-flow/cylc/flowwork/graphql.py", line 426, in async_resolve
    return await next_(root, info, **args)
  File "~/cylc-uiserver/cylc/uiserver/authorise.py", line 416, in async_resolve
    return await next_(root, info, **args)
  File "~/miniconda3/envs/cylc-8/lib/python3.7/site-packages/promise/iterate_promise.py", line 10, in iterate_promise
    yield from promise.future  # type: ignore
  File "~/miniconda3/envs/cylc-8/lib/python3.7/site-packages/promise/promise.py", line 844, in handle_future_result
    resolve(future.result())
  File " ~/cylc-flow/cylc/flowwork/schema.py", line 1277, in mutator
    res = await resolvers.mutator(info, command, w_args, args)
  File " ~/cylc-uiserver/cylc/uiserver/resolvers.py", line 201, in mutator
    'graphql', w_ids, graphql_args
  File " ~/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 305, in multi_request
    for msg_core in list(val.values())[0].get('result')
graphql.error.located_error.GraphQLLocatedError: 'list' object has no attribute 'values'
```

kinow:
@datamel I just tried calling it from GraphiQL and didn't get any error?

```graphql
mutation {
  ping (workflows: ["workflow_name"]) {
    result
  }
}
```

Are you able to confirm the issue is still occurring, please? 

Thanks!

kinow:
Oh, broken indeed. Just needed to run that against a running workflow :man_facepalming: 

[...]

PR #270, opened 2021-10-28
Remove ping from the UI server

MetRonnie:
These changes partially address #242. Sibling to https://github.com/cylc/cylc-flow/pull/4481 (see for why remove ping) and https://github.com/cylc/cylc-ui/pull/823

In terms of the traceback encountered in #242, that is a separate issue

**Requirements check-list**
- [x] Does not need tests 
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/authorise.py
+++ b/cylc/uiserver/authorise.py
@@ -83,11 +83,7 @@ class Authorization:
 
     # Access Groups
 
-    READ_OPS = [
-        "ping",
-        READ_OPERATION,
-    ]
-
+    READ_OPS = {READ_OPERATION}
     ASYNC_OPS = {"query", "mutation"}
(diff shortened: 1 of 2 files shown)


datamel (commented):
Not sure what the future plans for `ping` are? Is it staying on the cli? I would have expected it to, since some users will be unable to use the ui? Unless maybe we are getting rid of ping and expect users to use `cylc scan` instead?

There is an intention for users to be able to interact with other's flows on the cli in future, leaving this code in will have it in place for then (the mutation for ping will be required in this case cylc-flow side)?

datamel (approved):
In response to my comment above, as discussed in team meeting. Ping will be authorised with the `read` operation.
Thanks @MetRonnie.

Issue #269, opened 2021-10-26
Mutations intended for one workflow are received by others


MetRonnie:
Mutations intended for a particular workflow are received by other running workflows and they respond to them.

With the following mutation in graphiQL (or presumably by doing hold from the UI too):
```graphql
mutation {
  hold (
    workflows: ["temp4"]
    tasks: ["foo.3"]
  ) {
    result
  }
}
```

With just the hub running and no workflows running, the response is:
```json
{
  "data": {
    "hold": {
      "result": []
    }
  }
}
```
With the workflows `tutorial-family-triggers` and `temp4` running:
```json
{
  "data": {
    "hold": {
      "result": [
        "E",
        "r",
        "r",
        "o",
        "r",
        ":",
        " ",
        "N",
        "o",
        " ",
        "m",
        "a",
        "t",
        "c",
        "h",
        "i",
        "n",
        "g",
        " ",
        "W",
        "o",
        "r",
        "k",
        "f",
        "l",
        "o",
        "w",
        {
          "id": "rdutta|temp4",
          "response": [
            true,
            "Command queued"
          ]
        }
      ]
    }
  }
}
```

If you replace
https://github.com/cylc/cylc-flow/blob/c01278beed55b3275c2cd836dc20c6ecc907435a/cylc/flow/network/resolvers.py#L463

with `raise ValueError("somehting")` then you get a `GraphQLLocatedError` traceback in the terminal running the other workflow (`tutorial-family-triggers` in this case).

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

MetRonnie:
Note I am looking into the separate issue of why the error message gets broken up into individual characters as part of #271

dwsutherland:
Yeah, it's likely that the error message is in a different structure than the non-error.. The reason for this discrepancy in output formats is the attempted mapping of outputs, i.e.:

`scheduler-graphql` => `uis-graphql` => UI
`uis-graphql` => UI

So the mutation at the UIS might get a result from the UIS internals or the scheduler graphql output..

Issue #250, opened 2021-09-14
Different results for the same stop mutation


kinow:
**Describe the bug**

Some really strange behaviour in GraphiQL & UIS when testing the following mutation.
```
mutation {
  stop (workflows: ["user|cylc"]) { # <---- INVALID WORKFLOW ID
    result
  }
}
```

It returns different results, and a incorrectly displayed error message, but it depends on when you execute the query?

**Release version(s) and/or repository branch(es) affected?**

`master` everywhere.

**Steps to reproduce the bug**

1. Start UIS, don't run any workflows, run the mutation in GraphiQL. Reply (in less than 1 second):
```graphql
{
  "data": {
    "stop": {
      "result": []
    }
  }
}
```
2. Start `five`. Run the exact mutation (i.e. don't change the workflow ID). Reply (again, less than 1 second):
```graphql
{
  "data": {
    "stop": {
      "result": [
        "E",
        "r",
        "r",
        "o",
        "r",
        ":",
        " ",
        "N",
        "o",
        " ",
        "m",
        "a",
        "t",
        "c",
        "h",
        "i",
        "n",
        "g",
        " ",
        "W",
        "o",
        "r",
        "k",
        "f",
        "l",
        "o",
        "w"
      ]
    }
  }
}
```
3. It gets stranger. Now stop `five` and run the exact mutation yet again really fast! Don't wait too long! Reply (in ~2 seconds):
```graphql
{
  "errors": [
    {
      "message": "'NoneType' object has no attribute 'values'",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "stop",
        "result"
      ]
    }
  ],
  "data": {
    "stop": {
      "result": null
    }
  }
}
```

This error is similar to the error from: https://github.com/cylc/cylc-uiserver/issues/242
4. Oh, one more scenario. Instead of stopping and executing the GraphQL immediately, if you instead wait a few seconds after you stopped five, then you get the first reply again (<1 sec):
```graphql
{
  "data": {
    "stop": {
      "result": []
    }
  }
}
```

😵 any idea what could be wrong? Looks like there's some state in the UIS or flow that is not going well with the resolvers?

**Expected behavior**

Consistent responses, in a way that a user/dev would expect them to happen.

**Screenshots**

**Additional context**

Asked @dwsutherland over Element about it.

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
[...]

PR #274, opened 2021-11-09
Reinstate/fix multi-workflow request filter

dwsutherland:
These changes close https://github.com/cylc/cylc-uiserver/issues/269
Need to retest for https://github.com/cylc/cylc-uiserver/issues/250

The fix was to use the matching workflows and test against active workflows (not just send to all active workflows).

**no workflows:**
mutation_no_workflow

**matching workflow**
mutation_send_matching_workflow

**wildcard**
mutation_wildcards_still_work

**Requirements check-list**
- [ ] Cross project tests needed at some point.
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -190,6 +190,9 @@ class Resolvers(BaseResolvers):
         w_ids = [
             flow[WORKFLOW].id
             for flow in await self.get_workflows_data(w_args)]
+        if not w_ids:
+            return [{
+                'response': (False, 'No matching workflows')}]
         # Pass the request to the workflow GraphQL endpoints
         req_str, variables, _, _ = info.context.get('graphql_params')
         graphql_args = {
@@ -215,7 +218,8 @@ class Resolvers(BaseResolvers):
             flow[WORKFLOW].id
             for flow in await self.get_workflows_data(w_args)]
         if not w_ids:
-            return 'Error: No matching Workflow'
+            return [{
+                'response': (False, 'No matching workflows')}]
         # Pass the multi-node request to the workflow GraphQL endpoints
         req_str, variables, _, _ = info.context.get('graphql_params')
         graphql_args = {
--- a/cylc/uiserver/workflows_mgr.py
+++ b/cylc/uiserver/workflows_mgr.py
@@ -284,7 +284,8 @@ class WorkflowsManager:
                 command,
                 multi_args.get(w_id, args),
                 timeout,
-            ) for w_id in self.active
+            ) for w_id in workflows
+            if w_id in self.active
         }
         gathers = [
             workflow_request(req_context=info, *request_args, log=self.log)


cylc/uiserver/resolvers.py
```diff
@@ -215,7 +217,7 @@ async def nodes_mutator(self, info, *m_args):
             flow[WORKFLOW].id
             for flow in await self.get_workflows_data(w_args)]
         if not w_ids:
-            return 'Error: No matching Workflow'
+            return [{'error': 'No matching workflows'}]
```
oliver-sanders, line 220: I think this should be "error**s**" with the message in a map:

```suggestion
            return [{
                'errors': [{'message': 'No matching workflows'}]
            }]
```

http://spec.graphql.org/October2021/#sec-Errors

(If that doesn't play nicely with the UI [most likely my fault] we can fix later).

cylc/uiserver/resolvers.py
```diff
@@ -190,6 +190,8 @@ async def mutator(self, info, *m_args):
         w_ids = [
             flow[WORKFLOW].id
             for flow in await self.get_workflows_data(w_args)]
+        if not w_ids:
+            return [{'error': 'No matching workflows'}]
```
oliver-sanders, line 194: I think it should be:

```suggestion
            return [{
                'errors': [{'message': 'No matching workflows'}]
            }]
```

oliver-sanders (commented):
Makes sense, looks good, will test shortly.

cylc/uiserver/resolvers.py
MetRonnie: I think this will require some follow-up work.

At the moment this response is [going through the mutator function twice](https://matrix.to/#/!SDKYnTgYyOXNBpJySp:matrix.org/$1636136658135167ecqpg:matrix.org?via=matrix.org), so even if you apply that suggestion we get
```json
{
    "data": {
        "<mutation>": {
            "result": [
                {
                    "errors": [{"message": "No matching workflows"}]
                }
            ]
        }
    }
}
```
Whereas what we should be getting is just the `errors`, no `data`:
```json
{
    "errors": [{"message": "No matching workflows"}]
}
```
Another possibility is we treat the no-match result as a non-error, e.g.  
```json
{
    "data": {
        "<mutation>": {
            "result": [
                {
                    "id": "whatever",
                    "response": [false, "No matching workflows"]
                }
            ]
        }
    }
}
```
or something like that

dwsutherland: Yeah, we could raise a GraphQL error.. But I'm not sure this warrants it, it's just information about the result of a command (it's not like the command itself needs to fail, although you could argue it).

You wouldn't get an ID back (unless you use your args, but that could be more than one)..

oliver-sanders: Ah, ok, this is the generic response and not a GraphQL error.

In which case shouldn't it be a string rather than a map?

MetRonnie (approved):
Apart from https://github.com/cylc/cylc-uiserver/pull/274/files#r745478859 which could be follow-up work, have tested and am happy with this fix.

Only thing is, when sending the mutation for a existing but stopped workflow, I get no response
```json
{
  "data": {
    "hold": {
      "result": []
    }
  }
}
```

oliver-sanders:
The `GenericReponse` interface now has multiple formats:

Running workflow:

```json
 "result": [
   {       
     "id": "osanders|generic",     
     "response": [     
       true,     
       "Command queued"     
     ]     
   }     
 ]
```

No matching workflow(s):

```json
 "result": [     
   {     
     "error": "No matching workflows"     
   }     
 ]
```

No matching running workflow(s):

```json
 "result": [] 
```

It should be `[success: bool, message: str]`. The UI uses the `success` part to detect when mutations fail.

cylc/uiserver/resolvers.py
dwsutherland: It won't result in the format we want.. We'd have to raise or return some sort of `ExecutionResult` class or something

dwsutherland:
Ok changed to a `(False, "message")` type
Screenshot from 2021-11-10 00-28-11
:

dwsutherland:
In the future we can change it to w/e we like..

MetRonnie:
> In the future we can change it to w/e we like..

Instead of having the single `result` generic scalar output on mutations, could we instead split it into 2 outputs:
- The boolean, named something like `ok` or `success`
- The string, named `message`

dwsutherland:
> > In the future we can change it to w/e we like..
> 
> Instead of having the single `result` generic scalar output on mutations, could we instead split it into 2 outputs:
> 
> * The boolean, named something like `ok` or `success`
> * The string, named `message`

Yes, some of this stuff I just ported from the old API  (as I didn't know what the UI would want)..

oliver-sanders:
> could we instead split it into 2 outputs:

Yes we could, however, we can do something much cooler and track the command until complete. This was originally scheduled for 8.0.0.

oliver-sanders (approved):

MetRonnie:
> track the command until complete

I was wondering if that was deliberately not done to avoid waiting for the Scheduler command queue to process

oliver-sanders:
Yes, however, we can turn this into an event stream using a pub-sub like model allowing clients to decide whether they want to wait around or exit.

The cylc-flow issue is here: https://github.com/cylc/cylc-flow/issues/3329

oliver-sanders:
(note we couldn't really do this until recently as Cylc used to use HTTPS communications which aren't good for sending small chatty updates like this).

PR #277, opened 2021-11-10
Fix log bug

hjoliver:
@wxtim - I ran into a minor bug introduced by a recent PR of yours.

### DIFF
--- a/cylc/uiserver/authorise.py
+++ b/cylc/uiserver/authorise.py
@@ -451,7 +451,6 @@ class AuthorizationMiddleware:
                        f":requested to {op_name}.")
         if message:
             log_message = log_message + " " + message
-        Authorization.log.warning(log_message)
         raise web.HTTPError(http_code, reason=message)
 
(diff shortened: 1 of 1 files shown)


cylc/uiserver/authorise.py
```diff
@@ -451,7 +451,6 @@ def auth_failed(self, current_user: str, op_name: str,
                        f":requested to {op_name}.")
         if message:
             log_message = log_message + " " + message
-        Authorization.log.warning(log_message)
```
hjoliver, line 454: Authorization instances have a `log` attribute, but the class doesn't.

I guess you just wanted to the log the message to the server console?  Maybe not necessary, at `raise web.HTTPError()` (on the next line) seems to do that anyway?

wxtim: This replaces some logging here, which I don't think worked either. IMO this change is fine.

wxtim (approved):

PR #284, opened 2021-11-11
actions: test build on project metadata change

oliver-sanders:
~Markdown is fairly non-standard. PyPi and GitHub use different "flavours".~

~Adapt our markdown to make it PyPi compliant and add a test.~

~GitHub seems happy with the PyPi flavour.~

Build the project when its configuration changes. This can help to identify packaging errors which might only turn up at release time early on.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,41 @@
+name: Build
+
+# build the project whenever the configuration is changed
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'README.md'       # check markdown is valid
+      - 'MANIFEST.in'     # check packaging
+      - 'pyproject.toml'  # check build config
+      - 'setup.cfg'       # check deps and project config
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        os: ['ubuntu-latest']
+        python: ['3.7', '3.8', '3.9']
+        include:
+          - os: 'macos-latest'
+            python: '3.7'
+    steps:
+      - name: Checkout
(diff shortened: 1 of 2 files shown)


oliver-sanders:
Actually, lets generalise this into a build test...

oliver-sanders:
Done.

MetRonnie (commented):
The PyPI documentation says GitHub flavoured markdown is supported. I found that there is a simpler fix: see #285

I have also added a `twine check` step to the release-actions: https://github.com/cylc/release-actions/pull/25

oliver-sanders:
Leaving the build test as this is something we should have.

There's not really enough here to justify shipping it to release actions?

.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: 'ubuntu-latest'
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
```
MetRonnie, line 22: There isn't a `matrix` set in the workflow

MetRonnie:
> There's not really enough here to justify shipping it to release actions?

Not sure what you mean

oliver-sanders:
The common parts of the workflow only add up to a few lines so easy enough to copy.

Can centralise into release-actions if you like.

oliver-sanders:
The common parts of the workflow only add up to a few lines so easy enough to copy.

Can centralise into release-actions if you like.

oliver-sanders:
Moved the build into release-actions, the remainder of the workflow is mostly specific to the project.

MetRonnie (approved):

PR #285, opened 2021-11-11
Fix long description content type setting in setup.cfg

MetRonnie:
See https://github.com/pypa/twine/issues/479#issuecomment-514575729


oliver-sanders:
Wow, the quoted string was taken from a comment posted by the author of the metadata checking!

Ok, whatever.

oliver-sanders (approved):

PR #288, opened 2021-11-17
Pass authenticated user to cylc-flow

datamel:
This is necessary for logging which authenticated user executes the command on the UI.
Sibling pr: https://github.com/cylc/cylc-flow/pull/4522
These changes close https://github.com/cylc/cylc-uiserver/issues/224 and https://github.com/cylc/cylc-flow/issues/4294

- [x] Does not need tests - can not test this currently due to no multi user testing in the automated test battery.
- [x] No change log entry required -invisible to users.
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -227,31 +227,3 @@ class Resolvers(BaseResolvers):
             self.workflows_mgr,
             log=self.log
         )
-
-    async def nodes_mutator(self, info, *m_args):
-        """Mutate node items of associated workflows."""
-        _, _, w_args, _ = m_args
-        w_ids = [
-            flow[WORKFLOW].id
-            for flow in await self.get_workflows_data(w_args)]
-        if not w_ids:
-            return [{
-                'response': (False, 'No matching workflows')}]
-        auth_user = info.context.get('current_user', 'unknown user')
-
-        # Pass the multi-node request to the workflow GraphQL endpoints
-        _, variables, _, _ = info.context.get('graphql_params')
-
-        # Create a modified request string,
-        # containing only the current mutation/field.
-        operation_ast = deepcopy(info.operation)
-        operation_ast.selection_set.selections = info.field_asts
-
-        graphql_args = {
-            'request_string': print_ast(operation_ast),
-            'variables': variables,
-        }
-        multi_args = {w_id: graphql_args for w_id in w_ids}
-        return await self.workflows_mgr.multi_request(
-            'graphql', w_ids, multi_args=multi_args, auth_user=auth_user
-        )
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -194,8 +194,10 @@ class Resolvers(BaseResolvers):
     # Mutations
     async def mutator(self, info, *m_args):
         """Mutate workflow."""
+        req_meta = {}
         _, w_args, _, _ = m_args
-        auth_user = info.context.get('current_user', 'unknown user')
+        req_meta['auth_user'] = info.context.get(
+            'current_user', 'unknown user')
         w_ids = [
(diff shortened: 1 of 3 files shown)


wxtim (approved):

wxtim (commented):
- [x] Code makes sense
- [x] Demoed
- [ ] Passing tests

datamel:
Note test failure expected as it requires sibling cylc-flow to be checked out.

> * [x]  Code makes sense
> * [x]  Demoed
> * [ ]  Passing tests

Ah, tests will only pass if checked out with sibling on cylc-flow.

wxtim:
I can confirm that tests pass if Sibling PR is checked out. I will leave you to merge this and the sibling at the same time so as not to break master.

wxtim (approved):

cylc/uiserver/resolvers.py
```diff
@@ -189,7 +189,10 @@ def __init__(self, data, log, **kwargs):
     # Mutations
     async def mutator(self, info, *m_args):
         """Mutate workflow."""
-        _, w_args, _ = m_args
+        req_meta = {}
+        _, w_args, _, _ = m_args
```
MetRonnie, line 193: Tuple unpacking has gained an extra member by mistake?
```suggestion
        _, w_args, _ = m_args
```

datamel: I know the test is failing here due to this but you should find that if you have the sibling checked out on cylc-flow then the test will pass (with the correct unpacking). The unpacking accounts for the meta.

cylc/uiserver/resolvers.py
```diff
@@ -224,29 +227,3 @@ async def service(self, info, *m_args):
             self.workflows_mgr,
             log=self.log
         )
-
-    async def nodes_mutator(self, info, *m_args):
```
datamel, line 228: @dwsutherland, any chance you could double check that this is unused code. I couldn't get any breakpoints stopping here, can't find any code calls. Both myself and @oliver-sanders think it is not used and should be removed? Also, removed its cylc-flow partner in https://github.com/cylc/cylc-flow/pull/4522/files.

cylc/uiserver/resolvers.py
MetRonnie: Oh, it's also done in https://github.com/cylc/cylc-flow/pull/4522. It's generally a good idea to have the same people review sibling PRs

wxtim: I'm tempted to ask GH for a multi-repo PR feature - surely we aren't the only people doing this sort of thing....

datamel: It seems strange GH actions can't check out different branches. Maybe something to look into when time permits.

oliver-sanders: Kinda doable, we could put a special comment in the PR description that tells GHA what base branch to use, however, the failing tests are also a good indicator not to merge until the dependent branch is in.

oliver-sanders (approved):
(tests pass locally with the cylc-flow branch checked out)

MetRonnie (approved):

cylc/uiserver/resolvers.py
dwsutherland: It was in use, until:
https://github.com/cylc/cylc-flow/pull/3469
specifically:
https://github.com/cylc/cylc-flow/pull/3469/files#diff-97f4d62be41218e2d70a203edfca6d44d7c0046b6908cefa8e3af7c02cd31920

And the idea behind it was that mutations being multi-workflow could be given node ids from different workflows, and then these ids would be parsed and mutations sent to the appropriate workflow..

The fact that we moved away from this, means we are happy with not using this functionality... i.e. a node mutation will always be associated with only one workflow and/or the workflow parsing happens somewhere else.

PR #293, opened 2021-12-02
multi-mutation request filtered reconstruction

### LINKED ISSUE #289: Multi Request Mutations
On master of `cylc-ui`, `cylc-uiserver` and `cylc-flow`, running a multi mutation request on GraphiQL mutations are coming through the incorrect number of times.

**Steps to reproduce the bug**
1. Start a workflow (mine was called `test_auth`)
2. On the GraphiQL interface in the UI, 
```
 mutation {
    pause (workflows: ["test_auth"]) {
    result
  }
 stop (workflows: ["test_auth"]) {
    result
  }

}
```
This results in a workflow log of:

image

Note that this is not just a logging error, since `Workflow is already paused` appears in the log and the queue has each mutation twice (see code diff below to reproduce this logging)

**Expected behavior**
Mutations only come through once each, as requested by the user.

**Additional context**
- I added debugging to inspect the queue (to confirm my suspicions): my diff for cylc-flow is as follows:
```
diff --git a/cylc/flow/scheduler.py b/cylc/flow/scheduler.py
index af62bada8..c8271c8a0 100644
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -825,6 +825,9 @@ class Scheduler:
         if qsize <= 0:
             return
         LOG.info(f"Processing {qsize} queued command(s)")
+        import queue
+        for n in list(self.command_queue.queue):
+            LOG.warning(f"Queued item: {n}")
         while True:
             try:
                 name, args, kwargs = self.command_queue.get(False)

```
- Note that this may be a problem solved in `cylc-flow` so this issue may need transferring if the root cause is there.

- I also ran a 3 mutation request, using :
```
mutation {
    pause (workflows: ["test_auth"]) {
    result
  }
 resume (workflows: ["test_auth"]) {
    result
  }
 stop (workflows: ["test_auth"]) {
    result
  }

}

```
these came through as follows:  

image
 
Tagged against 1.x for now, I expect most users will use the gui buttons rather than this interface. 

**Pull requests welcome!**

dwsutherland:
These changes close #289

Each root-entry/mutation-field in a GraphQL request document would pass on the whole document to the Scheduler.

Which means two things for multi mutation-field requests:
- duplicates would be sent to the scheduler (for every mutation-field)
- one mutation-field would be carried through on the authorization of another.

This has been fixed, by using existing `python-graphql` tools to reconstruct a stripped down version of the request-doc.
A request to the UIS of the form:
```
mutation {
  pause (
    workflows: ["sutherlander|generic/run1"]
  ) {
    result
  }
  stop (
    workflows: ["sutherlander|generic/run1"]
  ) {
    result
  }
}
```
Results in two requests, like before, but the requests only have one mutation each:
```
DOC AST
Document(definitions=[OperationDefinition(operation='mutation', name=None, variable_definitions=[], directives=[], selection_set=SelectionSet(selections=[Field(alias=None, name=Name(value='pause'), arguments=[Argument(name=Name(value='workflows'), value=ListValue(values=[StringValue(value='sutherlander|generic/run1')]))], directives=[], selection_set=SelectionSet(selections=[Field(alias=None, name=Name(value='result'), arguments=[], directives=[], selection_set=None)]))]))])

RESULT
mutation {
  pause(workflows: ["sutherlander|generic/run1"]) {
    result
  }
}
```
```
DOC AST
Document(definitions=[OperationDefinition(operation='mutation', name=None, variable_definitions=[], directives=[], selection_set=SelectionSet(selections=[Field(alias=None, name=Name(value='stop'), arguments=[Argument(name=Name(value='workflows'), value=ListValue(values=[StringValue(value='sutherlander|generic/run1')]))], directives=[], selection_set=SelectionSet(selections=[Field(alias=None, name=Name(value='result'), arguments=[], directives=[], selection_set=None)]))]))])

RESULT
mutation {
  stop(workflows: ["sutherlander|generic/run1"]) {
    result
  }
}
```

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -16,7 +16,10 @@
 """GraphQL resolvers for use in data accessing and mutation of workflows."""
 
 import os
+from copy import deepcopy
 from subprocess import Popen, PIPE, DEVNULL
+
+from graphql.language.base import print_ast
 
 from cylc.flow.network.resolvers import BaseResolvers
 from cylc.flow.data_store_mgr import WORKFLOW
@@ -194,9 +197,15 @@ class Resolvers(BaseResolvers):
             return [{
                 'response': (False, 'No matching workflows')}]
         # Pass the request to the workflow GraphQL endpoints
-        req_str, variables, _, _ = info.context.get('graphql_params')
+        _, variables, _, _ = info.context.get('graphql_params')
+
+        # Create a modified request string,
+        # containing only the current mutation/field.
+        operation_ast = deepcopy(info.operation)
+        operation_ast.selection_set.selections = info.field_asts
+
         graphql_args = {
-            'request_string': req_str,
+            'request_string': print_ast(operation_ast),
             'variables': variables,
         }
         return await self.workflows_mgr.multi_request(
@@ -221,9 +230,15 @@ class Resolvers(BaseResolvers):
             return [{
                 'response': (False, 'No matching workflows')}]
         # Pass the multi-node request to the workflow GraphQL endpoints
-        req_str, variables, _, _ = info.context.get('graphql_params')
+        _, variables, _, _ = info.context.get('graphql_params')
+
+        # Create a modified request string,
+        # containing only the current mutation/field.
+        operation_ast = deepcopy(info.operation)
+        operation_ast.selection_set.selections = info.field_asts
+
         graphql_args = {
-            'request_string': req_str,
+            'request_string': print_ast(operation_ast),
             'variables': variables,
         }
         multi_args = {w_id: graphql_args for w_id in w_ids}


cylc/uiserver/resolvers.py
```diff
@@ -194,9 +197,15 @@ async def mutator(self, info, *m_args):
...
+        # Create a modified request string,
+        # containing only the current mutation/field.
+        operation_ast = deepcopy(info.operation)
+        operation_ast.selection_set.selections = info.field_asts
+
         graphql_args = {
-            'request_string': req_str,
+            'request_string': print_ast(operation_ast),
             'variables': variables,
         }
```
MetRonnie, line 210: Won't this have to be done for `nodes_mutator()` below too? If so it's probably worth extracting this out into a helper function?

MetRonnie (commented):
Nice and simple!

cylc/uiserver/resolvers.py
dwsutherland: (not the helper function bit.. but it's only 2 places, a small section, and I don't think it'll be used elsewhere)

datamel (approved):
Great fix, thanks @dwsutherland. I have re run my test (as in the issue) and requests are coming through correctly now: 
image

Not sure if it would be worth putting a test around `print_ast`, although perhaps since it is not private, they are unlikely to change the output.

MetRonnie (approved):
Have tested; confirmed it fixes the bug. Still think a helper function would be useful (would have prevented missing out `nodes_mutator` in this case, for example), but happy to merge without it.

Out of interest, where did you find out about `print_ast`? Is there a good docs source for graphql stuff like that (closest I've found is graphql-core-3.readthedocs.io/en/stable which I think is a major version up from what we're using)?

oliver-sanders (approved):
Great.

Found a way to test this, will raise in another PR.

PR #294, opened 2021-12-06
tests: add fixures and basic tests for graphql functionality

oliver-sanders:
* Test a basic {workflows{id,status}} query.
* Test multiple queries are properly split up and reissued.

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/tests/test_graphql.py
+++ b/cylc/uiserver/tests/test_graphql.py
@@ -1,0 +1,161 @@
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import json
+from textwrap import dedent
+
+import pytest
+
+
+@pytest.fixture
+def gql_query(jp_fetch):
+    """Perform a GraphQL request.
+
+    E.G:
+
+    response = await gql_query(
+        *('cylc', 'graphql'),
+        query='''
+            query {
+                workflows {
+                    id
+                    status
+                }
+            }
+        ''',
+    )
+
+    """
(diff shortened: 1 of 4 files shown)


oliver-sanders:
Dammit, tests passed locally, `pip install --upgrade`ed my packages, now I get this error, investigating.

oliver-sanders:
Fixed, the jupyterhub stuff was being imported by `--doctest-modules` which was messing with the environment causing the server to think auth was provided by the hub.

Added the hub files to the `ignore_collect` list which has to be implemented in a special way because pytest-doctest doesn't appear to read in `collect_ignore` items from the standard locations.

dwsutherland (approved):
I don't really understand how `jp_fetch` relates to `jp_server_config`... Some test magic/black-box tying this all together I assume... 🤯 

I do, however, understand what the new tests are testing.

The tests work locally, and coverage has increased.. 👍

oliver-sanders:
> I don't really understand how jp_fetch relates to jp_server_config... Some test magic/black-box tying this all together I assume... exploding_head

Magic!

We are borrowing from the Jupyter Server test framework which provides a fixture called jp_serverapp which launches a new server.

https://github.com/jupyter-server/jupyter_server/blob/82cd79181f10bf6447a4262674cea9420c0140f9/jupyter_server/pytest_plugin.py#L300-L306

The `jp_fetch` fixture depends on `jp_serverapp`:

https://github.com/jupyter-server/jupyter_server/blob/3af6396d800a803eed8c71339db3eb271ed13020/jupyter_server/pytest_plugin.py#L327-L328

So whenever you see `jp_fetch` it is spinning up a server under the hood.

Note that the `jp_serverapp` fixture has function level scoping so you get a *new* server instance for every `def test`.

The Cylc integration test battery has module scoped variants of heavy-weight fixtures e.g. `mod_flow` and `mod_run` to allow multiple tests in the same module to share the same fixture.

cylc/uiserver/tests/test_graphql.py
```diff
@@ -0,0 +1,161 @@
...
+                pause(workflows: ["*"]) {
+                    result
+                }
+                stop(workflows: ["*"]) {
+                    result
+                }
+            }
+        ''',
+    )
+    assert response.code == 200
```
wxtim, line 134: consider 
```
if response.code != 200:
    raise SomePytestError(response.reason)

wxtim (approved):
Nothing problematic. Happy for you to merge if you wish to disregard my single comment.

cylc/uiserver/tests/test_graphql.py
oliver-sanders: If the test fails the full response (including the `response.reason`) and server logs are dumped in the pytest output so this is already well captured.

Issue #291, opened 2021-11-29
configuration path root


oliver-sanders:
Follow on from https://github.com/cylc/cylc-uiserver/issues/189

This project is called cylc-uiserver, however, it configures both the hub and the UIS.

At present the config path is `~/.cylc/hub`, should this be changed to `~/.cylc/uiserver` to match `~/.cylc/flow`?

oliver-sanders:
Agreed, rename to `ui-server`.

hjoliver:
Somewhat relatedly, the server config file can be called:
- `jupyter_config.py`
- `jupyter_cylc_config.py`
- (`jupyterhub_config.py` for hub-launched only)

Should we document/recommend use of `jupyter_cylc_config.py` for maximum clarity?  

Even though `cylc` is in the path name for installed site and user configs, it would makes sense for command line use of temporary config files: `cylc gui --config=jupyter_cylc_config.py`.

oliver-sanders:
> Should we document/recommend use of jupyter_cylc_config.py for maximum clarity?

I think this file targets the Cylc extension so would not be loaded by other Jupyter components. These targeted files should be used to configure the one specific component (server, extension, hub, lab, etc) whereas we use it to configure multiple.

I think using the generic `jupyter_config.py` as documented is the clearest way.

PR #301, opened 2022-01-28
Adds version hierarchy to config files

### LINKED ISSUE #199: conf: hierarchical configuration path
Cylc Flow has a configuration file hierarchy which allows us to provide different configuration for different installed versions of Cylc E.G:

* `~/.cylc/flow/global.cylc`
* `~/.cylc/flow/8/global.cylc`
*`~/.cylc/flow/8.0/global.cylc`

The UIS/Hub could probably do with the same treatment.

I would make more sense to use the Cylc Flow (meta) version for this hierarchy because:

* It is already used for this purpose.
* The UI Server version isn't user-facing (visible via `cylc version --long`).
* The Hub version is obscure (implementation detail visible via `pip list`).

**Pull requests welcome!**

datamel:
For hub_version `0.6.0`,  the load order can be seen below: 
image

These changes close #199 and #291 

- [x] No dependency changes.

### DIFF
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -90,7 +90,7 @@ from cylc.uiserver.workflows_mgr import WorkflowsManager
 
 # UIS configuration dirs
 UISERVER_DIR = 'uiserver'
-USER_CONF_ROOT: Path = Path(f'~/.cylc/{UISERVER_DIR}').expanduser()
+USER_CONF_ROOT = Path.home() / '.cylc' / UISERVER_DIR
 SITE_CONF_ROOT: Path = Path(
     os.getenv('CYLC_SITE_CONF_PATH')
     or GlobalConfig.DEFAULT_SITE_CONF_PATH,
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -41,7 +41,6 @@ Cylc specific configurations are documented here.
 """
 
 import getpass
-import os
 from pathlib import Path, PurePath
 import sys
 from typing import List
@@ -62,7 +61,6 @@ from traitlets import (
 
 from jupyter_server.extension.application import ExtensionApp
 
-from cylc.flow.cfgspec.globalcfg import GlobalConfig
 from cylc.flow.network.graphql import (
     CylcGraphQLBackend, IgnoreFieldMiddleware
 )
@@ -81,21 +79,15 @@ from cylc.uiserver.handlers import (
     UIServerGraphQLHandler,
     UserProfileHandler,
 )
-from cylc.uiserver.jupyterhub_config import get_conf_dir_hierarchy
+from cylc.uiserver.jupyterhub_config import (
+    get_conf_dir_hierarchy,
+    SITE_CONF_ROOT,
+    USER_CONF_ROOT
+)
 from cylc.uiserver.resolvers import Resolvers
 from cylc.uiserver.schema import schema
 from cylc.uiserver.websockets.tornado import TornadoSubscriptionServer
 from cylc.uiserver.workflows_mgr import WorkflowsManager
-
-
-# UIS configuration dirs
-UISERVER_DIR = 'uiserver'
-USER_CONF_ROOT = Path.home() / '.cylc' / UISERVER_DIR
-SITE_CONF_ROOT: Path = Path(
-    os.getenv('CYLC_SITE_CONF_PATH')
-    or GlobalConfig.DEFAULT_SITE_CONF_PATH,
-    UISERVER_DIR
-)
 
 
 class PathType(TraitType):
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -88,11 +88,12 @@ from cylc.uiserver.workflows_mgr import WorkflowsManager
 
 
 # UIS configuration dirs
-USER_CONF_ROOT: Path = Path('~/.cylc/hub').expanduser()
+UISERVER_DIR = 'uiserver'
+USER_CONF_ROOT: Path = Path(f'~/.cylc/{UISERVER_DIR}').expanduser()
 SITE_CONF_ROOT: Path = Path(
     os.getenv('CYLC_SITE_CONF_PATH')
     or GlobalConfig.DEFAULT_SITE_CONF_PATH,
-    'hub'
+    UISERVER_DIR
 )
 
 
@@ -227,7 +228,7 @@ class CylcUIServer(ExtensionApp):
             ''' + AUTH_DESCRIPTION + '''
 
             Example configuration, residing in
-            ``~/.cylc/hub/jupyter_config.py``:
+            ``~/.cylc/uiserver/jupyter_config.py``:
 
             .. code-block:: python
 
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -17,9 +17,9 @@
 """
 Cylc UI Server can be configured using a ``jupyter_config.py`` file, loaded
 from a hierarchy of locations. This hierarchy includes the prepackaged
-configuration, the site directory, which defaults to ``/etc/cylc/uiserver`` but can be set with the
-environment variable ``$CYLC_SITE_CONF_PATH`` ) and the user directory
-(``~/.cylc/uiserver``).
+configuration, the site directory, which defaults to ``/etc/cylc/uiserver`` but
+can be set with the environment variable ``$CYLC_SITE_CONF_PATH`` ) and
+the user directory (``~/.cylc/uiserver``).
 For example, at Cylc UI Server version 0.6.0, the hierarchy (highest priority
 at the bottom) would be:
 
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -16,12 +16,14 @@
 
 """
 Cylc UI Server can be configured using a ``jupyter_config.py`` file, loaded
-from a hierarchy of locations. Including the site directory,
-(defaults to ``/etc/cylc/uiserver``) and the user directory
+from a hierarchy of locations. Including the site directory, which defaults
+to ``/etc/cylc/uiserver`` but can be set with the environment variable
+``$CYLC_SITE_CONF_PATH`` ) and the user directory
 (``~/.cylc/uiserver``).
-For example, at Cylc UI Server version 0.6.0, the hierarchy would be:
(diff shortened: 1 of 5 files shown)


cylc/uiserver/jupyterhub_config.py
```diff
@@ -49,21 +55,35 @@ def _load(path):
         exec(path.read_text())
 
 
+def get_conf_dir_hierarchy(config_paths: List[Path]):
```
MetRonnie, line 58:
```suggestion
def get_conf_dir_hierarchy(config_paths: List[Path]) -> List[Path]:
```

cylc/uiserver/jupyterhub_config.py
```diff
@@ -49,21 +55,35 @@ def _load(path):
...
+    """Takes list of config paths, adds version and filename to the path"""
+    conf_hierarchy = []
+    version_hierarchy = get_version_hierarchy(hub_version or __version__)
+    for x in product(config_paths, version_hierarchy):
+        conf_hierarchy.append(Path(x[0], x[1], 'jupyter_config.py'))
+    return conf_hierarchy
+
+
 def load():
```
MetRonnie, line 67: (Type annotations in a function body aren't checked unless the definition has type annotations)

```suggestion
def load() -> None:
```

cylc/uiserver/jupyterhub_config.py
```diff
@@ -49,21 +55,35 @@ def _load(path):
...
     if os.getenv('CYLC_SITE_CONF_PATH'):
         site_conf_path: Path = Path(
             os.environ['CYLC_SITE_CONF_PATH'],
-            'hub/jupyter_config.py'
+            UISERVER_DIR
         )
     else:
-        site_conf_path: Path = SITE_CONF_PATH
-    config_paths = [DEFAULT_CONF_PATH, site_conf_path, USER_CONF_PATH]
+        site_conf_path: Path = SITE_CONF_ROOT
```
MetRonnie, line 75: Already declared
```suggestion
        site_conf_path = SITE_CONF_ROOT
```

MetRonnie (commented):
Should these two be deleted?
https://github.com/cylc/cylc-uiserver/blob/0c17a396c4f2cee33b9459a3f29a5639168ceee8/cylc/uiserver/jupyterhub_config.py#L45-L48

MetRonnie:
For some reason it's not using the `jupyter_config.py` I have moved from `~/.cylc/hub/` to `~/.cylc/uiserver/0.6/`. I have verified that it gets loaded using breakpoints, but the UI build being used is the bundled one instead of the one specified
```
c.CylcUIServer.ui_build_dir = '<home>/github/cylc-ui/dist'
```

datamel:
> For some reason it's not using the `jupyter_config.py` I have moved from `~/.cylc/hub/` to `~/.cylc/uiserver/0.6/`. I have verified that it gets loaded using breakpoints, but the UI build being used is the bundled one instead of the one specified
> 
> ```
> c.CylcUIServer.ui_build_dir = '<home>/github/cylc-ui/dist'
> ```
I'll look into this now, thanks @MetRonnie

MetRonnie:
Here is the callstack for that breakpoint, in case it's any use
image

oliver-sanders:
> I have verified that it gets loaded using breakpoints

It could be loading other config files which reset it back, try running in debug mode, you should see a log entry every time something changed.

MetRonnie:
```
cylc hub --debug
```

I have found no real difference in the log output with this PR or master checked out, other than:
- the expected `~/.cylc/hub -> ~/.cylc/uiserver`
-
    ```diff
    -[I CylcUIServer] Serving UI from: ~/github/cylc-ui/dist
    +[I CylcUIServer] Serving UI from: ~/github/cylc-uiserver/cylc/uiserver/ui/0.6.0
   ```

In neither was there an indication of what `jupyter_config.py` was loaded.

MetRonnie:
Neither
https://github.com/cylc/cylc-uiserver/blob/aa31dea3df52dbf08226c66acc37cf8c67364694/cylc/uiserver/app.py#L379-L384

nor
https://github.com/cylc/cylc-uiserver/blob/5ce8fa6a8fa7eecde9225f8cb05fac5d79c1eaee/cylc/uiserver/jupyterhub_config.py#L48

are logging anything for me with `cylc hub --debug`

oliver-sanders:
```
cylc gui --debug
```

cylc/uiserver/jupyterhub_config.py
oliver-sanders, line 58: This file is only loaded by JupyterHub, to open it up to Jupyter Server too need to add it to the list of `config_file_paths` in `cylc.uiserver.CylcUIServer`

oliver-sanders (commented):
Should do the trick for JupyterHub, needs a small addition to make it work for the UI Server too.

cylc/uiserver/jupyterhub_config.py
MetRonnie: Need to update? https://github.com/cylc/cylc-uiserver/blob/1b2a5993d41e03e6291e917c91d4d09ee7feeb9c/cylc/uiserver/app.py#L103-L115

MetRonnie:
It's not loading my `~/.cylc/uiserver/0.6/jupyter_config.py` for `cylc gui` either

cylc/uiserver/app.py
```diff
@@ -128,16 +122,18 @@ class CylcUIServer(ExtensionApp):
...
+                [
+                    # user configuration
+                    USER_CONF_ROOT,
+                    # site configuration
+                    SITE_CONF_ROOT,
+                ], filename=False
+            )
         )
     )
+    config_file_paths.append(str(Path(uis_pkg).parent))
```
oliver-sanders, line 135: We should add the path hierarchy here rather than loading the hub config, this lets Jupyter Server take care of loading the files correctly.

MetRonnie (commented):
hub -> uiserver here?
https://github.com/cylc/cylc-uiserver/blob/3dade03dbc88074c23f9ed06b3c05c1925b7ae01/cylc/uiserver/app.py#L20-L22

cylc/uiserver/jupyterhub_config.py
```diff
@@ -49,21 +58,40 @@ def _load(path):
         exec(path.read_text())
 
 
-def load():
+def get_conf_dir_hierarchy(
+        config_paths: List[Path], filename: Optional[bool] = True
```
MetRonnie, line 62: `Optional[X]` means `Union[X, None]`. Not needed for default args unless the default is `None`

```suggestion
        config_paths: List[Path], filename: bool = True
```

MetRonnie (approved):
It is correctly loading the moved `jupyter_config.py` now

MetRonnie (changes requested):
Ah wait, forgot about the requested change in my earlier comment

> hub -> uiserver here?
> https://github.com/cylc/cylc-uiserver/blob/3dade03dbc88074c23f9ed06b3c05c1925b7ae01/cylc/uiserver/app.py#L20-L22

datamel:
> Ah wait, forgot about the requested change in my earlier comment
> 
> > hub -> uiserver here?
> > https://github.com/cylc/cylc-uiserver/blob/3dade03dbc88074c23f9ed06b3c05c1925b7ae01/cylc/uiserver/app.py#L20-L22

Great, well spotted, that just came in on a rebase.

cylc/uiserver/app.py
```diff
@@ -81,21 +79,17 @@
     UIServerGraphQLHandler,
     UserProfileHandler,
 )
+from cylc.uiserver.jupyterhub_config import (
+    get_conf_dir_hierarchy,
+    SITE_CONF_ROOT,
+    USER_CONF_ROOT
+)
```
oliver-sanders, line 86: Because of the way JupyterHub config files work the `jupyterhub_config` file auto loads so we should not import it.

Can move the function and vars into this file or put them in another module.

cylc/uiserver/jupyterhub_config.py
```diff
@@ -49,21 +58,40 @@ def _load(path):
         exec(path.read_text())
 
 
-def load():
+def get_conf_dir_hierarchy(
+        config_paths: List[Path], filename: bool = True
+) -> List[Path]:
+    """Takes list of config paths, adds version and filename to the path"""
+    conf_hierarchy = []
+    version_hierarchy = get_version_hierarchy(hub_version or __version__)
```
oliver-sanders, line 66: Use the package version here (no need for the switch):

```suggestion
    version_hierarchy = get_version_hierarchy(__version__)
```

cylc/uiserver/jupyterhub_config.py
```diff
@@ -49,21 +58,40 @@ def _load(path):
...
+    for x in product(config_paths, version_hierarchy):
+        if filename:
+            conf_hierarchy.append(Path(x[0], x[1], 'jupyter_config.py'))
+        else:
+            conf_hierarchy.append(Path(x[0], x[1]))
+    return conf_hierarchy
```
oliver-sanders, line 72: Can avoid the `map(str(` in `cylc.uiserver.app` by:

```suggestion
    return str(conf_hierarchy)
```

oliver-sanders (commented):
Sorry, could do with shifting the logic out of the hub config, this file is only intended for JupyterHub and is not supposed to be imported directly for normal use because it auto-loads.

cylc/uiserver/app.py
```diff
@@ -15,11 +15,24 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Cylc UI Server can be configured using a ``jupyter_config.py`` file located in:
+Cylc UI Server can be configured using a ``jupyter_config.py`` file, loaded
+from a hierarchy of locations. This hierarchy includes the prepackaged
+configuration, the site directory, which defaults to ``/etc/cylc/uiserver`` but
+can be set with the environment variable ``$CYLC_SITE_CONF_PATH`` ) and
```
oliver-sanders, line 21: MismatchedBracketError:

```suggestion
configuration, the site directory (which defaults to ``/etc/cylc/uiserver`` but
can be set with the environment variable ``$CYLC_SITE_CONF_PATH``) and
```

oliver-sanders (approved):
Looks good to me, tested as working, just need to get the test passing.

datamel:
> Looks good to me, tested as working, just need to get the test passing.

Yep thanks! On it now. I can't reproduce locally but I expect a path expansion problem.

MetRonnie (approved):

PR #320, opened 2022-02-17
setup: remove dependency caps

oliver-sanders:
I don't think there is a good reason for these dependency caps.

Only JupyterHub has the potential to be a breaking change at the moment, will need to test out v2.

(sites will likely install jupyterhub separately so we need to ensure compatibility in the 1.0 series anyway).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -74,7 +74,7 @@ cylc.command =
 
 [options.extras_require]
 hub =
-    jupyterhub==1.4.*
+    jupyterhub>=1.4.0
 tests =
     coverage>=5.0.0
     flake8-broken-line>=0.3.0


oliver-sanders:
Going to back out of unpinning the graphql dependencies due to sustainability issues with those packages.

Because JupyterHub may be installed in a separate env to the UIS we need to maintain a wide compatibility bracket with it.

Tested spawning/stopping with JupyterHub 2.2.2, all good.

datamel (commented):
Tried with jupyterhub 2.2.2, go the following error with failure to spawn:
```
Found database schema version 4dc2d5a8c53c != 833da8570507. Backup your database and run `jupyterhub upgrade-db` to upgrade to the latest schema.
```
I ran the recommended  `jupyterhub upgrade-db`, but the error remains.



Update: https://jupyterhub.readthedocs.io/en/stable/admin/upgrading.html?highlight=upgrade#upgrade-jupyterhub-database
I needed to be in `.cylc/uiserver` before running the command. 
I have now got the hub running. Will run some more tests before approving.

datamel (approved):
Running smoothly (aside from the need to update the db) on 2.2.2.

oliver-sanders:
The issue of migrating a JupyterHub DB from v1 to v2 is a JupyterHub one (I just deleted my DB), so long as it works after that's sorted we're good.

oliver-sanders:
(rebased onto upstream/master so the tests pass)

Issue #308, opened 2022-02-11
Traceback in Logs


datamel:
I am getting transient traceback on uiserver (all branches master), I think the steps below reproduce:

**Steps to reproduce the bug**

1. Have some pre-run (on cli) workflows in cylc-run directory (I had run rose test battery recently)

2. Start the hub: `cylc hub`.

I get traceback as follows:
```
[D 2022-02-11 10:18:29.780 CylcUIServer] sync_workflow(~<username>/rtb.20220211T094740Z/23-app-fcm-make-dest-cont-alt-with-share/-kpDDb)
[E 2022-02-11 10:18:29.788 CylcHubApp ioloop:761] Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7f76fb4c3e50>>, <Task finished name='Task-520' coro=<WorkflowsManager.update() done, defined at ~/cylc-uiserver/cylc/uiserver/workflows_mgr.py:224> exception=KeyError('req_client')>)
    Traceback (most recent call last):
      File “~/.conda/envs/cylc-8.0rc1.dev/lib/python3.9/site-packages/tornado/ioloop.py", line 741, in _run_callback
        ret = callback()
      File “~/.conda/envs/cylc-8.0rc1.dev/lib/python3.9/site-packages/tornado/ioloop.py", line 765, in _discard_future_result
        future.result()
      File “~/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 252, in update
        if not await self._connect(wid, flow):
      File “~/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 197, in _connect
        await self.uiserver.data_store_mgr.sync_workflow(
      File “~/cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 154, in sync_workflow
        await self.entire_workflow_update(ids=[w_id])
      File "~/cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 382, in entire_workflow_update
        gathers = [
      File "~/cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 382, in <listcomp>
        gathers = [
      File "~ /cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 377, in <genexpr>
        {'client': info['req_client'],
    KeyError: 'req_client'	
```

3. I then delete that folder and get a load of error messages like the following (One per workflow):

```
[...]

PR #324, opened 2022-02-25
Workflows manager

### LINKED ISSUE #223: scan: check the uuid for stopped workflows
~Blocked by: https://github.com/cylc/cylc-flow/pull/4292~

Small bug that can cause the UI to maintain data in its data store which shouldn't be there. As we accumulate offline data in the UIS/UI data stores this will become more of an issue.

If the database of a stopped workflow has been removed then it is no longer a continuation of the same run as before, the UIS should wipe any cached data about the flow and send a `prune` and `added` delta to the UI to allow it to keep its data store in sync.

I can think of two ways the database might be removed:

* Purposeful removal to permit a re-run:

   ```console
   $ cylc play a
   $ cylc stop a
   $ cylc clean a --rm .service/db
   $ cylc play a
   ```

* Re-installation (but not reinstallation :smile:):

   ```console
   $ cylc play a
   $ cylc stop a
   $ cylc clean a
   $ cylc install a
   ```

Should be an easyish fix, means checking the workflow db for stopped workflows for each scan interval.

**Pull requests welcome!**

### LINKED ISSUE #310: traceback: active.pop
Quick traceback to suppress:

```
[E 2022-02-15 11:06:28.010 ServerApp] Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7fc4ea6800d0>>, <Task finished coro=<WorkflowsManager.update() done, defined at /Users/user/cylc-uiserver/cylc/uiserver/workflows_mgr.py:224> exception=KeyError('~user/cit-20220215T204121+0935/integration.test_workflow_files/test_detect_old_contact_file_running/4ea15b34-8e4f-11ec-b6d7-acde48001122')>)
      File "/Users/user/mambaforge/envs/cylc8/lib/python3.7/site-packages/tornado/ioloop.py", line 741, in _run_callback
        ret = callback()
      File "/Users/user/mambaforge/envs/cylc8/lib/python3.7/site-packages/tornado/ioloop.py", line 765, in _discard_future_result
        future.result()
      File "/Users/user/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 267, in update
        self.active.pop(wid)
    KeyError: '~user/cit-20220215T204121+0935/integration.test_workflow_files/test_detect_old_contact_file_running/4ea15b34-8e4f-11ec-b6d7-acde48001122'
```
**Pull requests welcome!**

### LINKED ISSUE #311: workflow state can be incorrect
* The workflow manager scans the filesystem detecting state changes.
* The workflow manager  invokes the appropriate register/connect/disconnect/purge interfaces in the data store manager on state changes.
* Both components are async.
* Updates can be performed whilst the data store manager is `await`ing something.
* Sometimes things go wrong in the data store manager as a result of this.
* This cases some of the UIS tracebacks.
* It also causes the workflow and data store managers to fall out of sync over workflow states.

To reproduce, try stopping, deleting, installing and running a bunch of workflows in sequence:

```console
$ cylc install <whatever> --flow-name=x
$ for i in $(seq 1 10); do id="x-$i"; cylc stop --now --now "$id"; rm -rf "$id"; cp -r "x" "$id"; cylc play --pause "$id"; done
```

(note i'm using `cp` rather than `cylc install` because it's faster which helps replicate the bug)

After each run of the above command you should see 10 running workflows in the GUI. Every now and again you will see one or more stopped workflows. If you run the UIS in `--debug` mode you will be able to see from the logs that `sync_workflow` was attempted, however, the workflow still shows as stopped because something went wrong.

It's ok for things to go wrong in these circumstances. If 10 workflows suddenly start up it might take a while to crunch through the list, in the mean time if anything happens to any of the workflows yet to be processed things could go wrong. The UIS should be resistant to this, I think the workflow manager should compare scan results to the workflows in the data store rather than the workflows from the last scan. That way stuck workflows would be pushed back through the pipe.

In the mean time the workaround is to turn it off and on again.

**Pull requests welcome!**

### LINKED ISSUE #312: prevent parallel scans
Somewhat related to https://github.com/cylc/cylc-uiserver/issues/311

The filesystem scan runs to a user-defined schedule. As a result it is possible for two scans to happen in parallel.

Whilst this is unlikely under normal circumstances, on startup if there is a large backlog of workflows to chew through it might be possible. Also client timeouts could cause issues too.

Not sure what the best way to avoid this is?

1) We could have each scan schedule its next instance.
  * Works fine in practice (this is how Tui currently works).
  * However, if the chain breaks it's done for.
2) Is there some tornado / asyncio functionality (tornado is now implemented with asyncio) that can help?
3) Other ideas?

**Pull requests welcome!**

oliver-sanders:
Closes #223, #310, #311, #312 and most of #308

* Check the presence of the workflow database for workflows which have previously run.
   *  (closes #223).
  * When workflows stop we keep their data in the store.
  * In the UI you will still see the state of the task pool when the workflow shut down.
  * If the user deletes the workflow database the data store needs to be wiped because they have deleted the continuity of the workflow run, i.e. it is effectively a new workflow run now.
* Ensure that one scan finishes before the next begins.
  * (closes #312).
  * Scans are scheduled to an interval.
  * When changes in workflow state are detected the workflow manager calls the relevant methods in the data store manager to adapt to this change.
  * Because these operations are await'ed in the scan they hold up the scan.
  * This means it is possible two scans to overlap even if you set a sensible interval between them.
  * Especially likely on startup if you have loads of workflows.
* Ensure workflow disconnect happens before re-connect.
  * (closes #311).
  * Previously the workflow manager was correctly telling the data store to connect to running workflows.
  * However, if the disconnect hadn't been performed, the new connect would just return rather than doing anything.
  * Ensure disconnect is called when necessary.

Starting and stopping a batch of workflows, before some would occasionally get stuck in the stopped state:

Taking a single workflow through a full cycle:

cycle

Removing the workflow database now causes the workflow to be re-registered (data from the old run will disappear):

db-remove

**Requirements check-list**
- [x] No dependency changes (one dep change but tests only)

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -141,9 +141,9 @@ class DataStoreMgr:
                 contact_data[CFF.PUBLISH_PORT]
             )
         )
-        sucessfull_updates = await self._entire_workflow_update(ids=[w_id])
+        successful_updates = await self._entire_workflow_update(ids=[w_id])
 
-        if w_id not in sucessfull_updates:
+        if w_id not in successful_updates:
             # something went wrong, undo any changes to allow for subsequent
             # connection attempts
             self.log.debug(f'failed to connect to {w_id}')
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -150,29 +150,17 @@ class DataStoreMgr:
         Call this when a workflow has stopped.
         """
         self.log.debug(f'disconnect_workflow({w_id})')
+        self._update_contact(
+            w_id,
+            status=WorkflowStatus.STOPPED.value,
+            status_msg=self._get_status_msg(w_id, False),
+        )
         if w_id in self.w_subs:
             self.w_subs[w_id].stop()
             del self.w_subs[w_id]
         if w_id in self.executors:
             self.executors[w_id].shutdown(wait=True)
             del self.executors[w_id]
-
-    def stop_workflow(self, w_id):
-        """Inform the data store that a workflow has stopped.
-
-        Call this when the data store thinks a workflow is running but it
-        isn't. Normally the data store will be subscribed to the workflow
-        when it shuts down so will be sent a shutdown message. However, in
-        some circumstances the data store will not be subscribed so will not
-        receive this message and we have to call this method manually.
-        """
-        self.log.debug(f'stop_workflow({w_id})')
-        self.disconnect_workflow(w_id)
-        self._update_contact(
-            w_id,
-            status=WorkflowStatus.STOPPED.value,
-            status_msg=self._get_status_msg(w_id, False),
-        )
 
     def get_workflows(self):
         """Return all workflows the data store is currently tracking.
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -128,7 +128,6 @@ class DataStoreMgr:
             return
 
         self.delta_queues[w_id] = {}
-        self._update_contact(w_id, contact_data)
 
         # Might be options other than threads to achieve
         # non-blocking subscriptions, but this works.
@@ -142,7 +141,17 @@ class DataStoreMgr:
                 contact_data[CFF.PUBLISH_PORT]
             )
         )
-        await self._entire_workflow_update(ids=[w_id])
+        sucessfull_updates = await self._entire_workflow_update(ids=[w_id])
+
+        if w_id not in sucessfull_updates:
+            # something went wrong, undo any changes to allow for subsequent
+            # connection attempts
+            self.log.debug(f'failed to connect to {w_id}')
+            self.disconnect_workflow(w_id)
+            return False
+        else:
+            # don't update the contact data until we have successfully updated
+            self._update_contact(w_id, contact_data)
 
     def disconnect_workflow(self, w_id):
         """Terminate workflow subscriptions.
@@ -356,6 +365,8 @@ class DataStoreMgr:
             if not ids or kwargs['req_context'] in ids
         ]
         items = await asyncio.gather(*gathers, return_exceptions=True)
+
+        successes = set()
         for item in items:
             if isinstance(item, Exception):
                 self.log.exception(
@@ -378,6 +389,13 @@ class DataStoreMgr:
                             continue
                         new_data[field.name] = {n.id: n for n in value}
                     self.data[w_id] = new_data
+                    successes.add(w_id)
+                else:
+                    self.log.error(
+                        f'Error: communicating with {w_id} - {result}'
+                    )
+
+        return successes
 
     def _update_contact(
         self,
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -74,13 +74,330 @@ class DataStoreMgr:
         self.executors = {}
         self.delta_queues = {}
 
-    def update_contact(
-            self,
+    async def register_workflow(self, w_id: str, is_active: bool) -> None:
+        """Register a new workflow with the data store.
+
+        Call this when a new workflow is discovered on the file system
+        (e.g. installed).
+        """
+        self.log.debug(f'register_workflow({w_id})')
+        self.delta_queues[w_id] = {}
+
+        # create new entry in the data store
+        data = deepcopy(DATA_TEMPLATE)
+        self.data[w_id] = data
+
+        # create new entry in the delta store
+        self._update_contact(
             w_id,
-            contact_data=None,
-            status=None,
-            status_msg=None,
-            pruned=False,
+            status=WorkflowStatus.STOPPED.value,
+            status_msg=self._get_status_msg(w_id, is_active),
+        )
+
+    async def unregister_workflow(self, w_id):
+        """Remove a workflow from the data store entirely.
+
+        Call this when a workflow is deleted.
+        """
+        self.log.debug(f'unregister_workflow({w_id})')
+        if w_id in self.data:
+            self._update_contact(w_id, pruned=True)
+        while any(
+            not delta_queue.empty()
+            for delta_queue in self.delta_queues.get(w_id, {}).values()
+        ):
+            await asyncio.sleep(self.PENDING_DELTA_CHECK_INTERVAL)
+        self._purge_workflow(w_id)
+
+    async def connect_workflow(self, w_id, contact_data):
+        """Initiate workflow subscriptions.
+
+        Call this when a workflow has started.
+
+        Subscriptions and sync management is instantiated and run in
+        a separate thread for each workflow. This is to avoid the sync loop
+        blocking the main loop.
+
+        """
+        self.log.debug(f'connect_workflow({w_id})')
+        if self.loop is None:
+            self.loop = asyncio.get_running_loop()
+
+        # don't sync if subscription exists
+        if w_id in self.w_subs:
+            return
+
+        self.delta_queues[w_id] = {}
+        self._update_contact(w_id, contact_data)
+
+        # Might be options other than threads to achieve
+        # non-blocking subscriptions, but this works.
+        self.executors[w_id] = ThreadPoolExecutor()
+        self.executors[w_id].submit(
+            partial(
+                self._start_subscription,
+                w_id,
+                contact_data['name'],
+                contact_data[CFF.HOST],
+                contact_data[CFF.PUBLISH_PORT]
+            )
+        )
+        await self._entire_workflow_update(ids=[w_id])
+
+    def disconnect_workflow(self, w_id):
+        """Terminate workflow subscriptions.
+
+        Call this when a workflow has stopped.
+        """
+        self.log.debug(f'disconnect_workflow({w_id})')
+        if w_id in self.w_subs:
+            self.w_subs[w_id].stop()
+            del self.w_subs[w_id]
+        if w_id in self.executors:
+            self.executors[w_id].shutdown(wait=True)
+            del self.executors[w_id]
+
+    def stop_workflow(self, w_id):
+        """Inform the data store that a workflow has stopped.
+
+        Call this when the data store thinks a workflow is running but it
+        isn't. Normally the data store will be subscribed to the workflow
+        when it shuts down so will be sent a shutdown message. However, in
+        some circumstances the data store will not be subscribed so will not
+        receive this message and we have to call this method manually.
+        """
+        self.log.debug(f'stop_workflow({w_id})')
+        self.disconnect_workflow(w_id)
+        self._update_contact(
+            w_id,
+            status=WorkflowStatus.STOPPED.value,
+            status_msg=self._get_status_msg(w_id, False),
+        )
+
+    def get_workflows(self):
+        """Return all workflows the data store is currently tracking.
+
+        Returns:
+            (active, inactive)
+                active: Set of active (running, paused, stopping) workflows.
+                inactive: Set of (stopped) workflows.
+        """
+        active = set()
+        inactive = set()
+        for w_id, workflow in self.data.items():
+            status = getattr(workflow.get('workflow'), 'status', 'stopped')
+            if status == 'stopped':
+                inactive.add(w_id)
+            else:
+                active.add(w_id)
+        return active, inactive
+
+    def _purge_workflow(self, w_id):
+        """Purge the manager of a workflow's subscription and data."""
+        self.log.debug(f'delete_workflow({w_id})')
+        self.disconnect_workflow(w_id)
+        if w_id in self.data:
+            del self.data[w_id]
+        if w_id in self.delta_queues:
+            del self.delta_queues[w_id]
+
+    def _start_subscription(self, w_id, reg, host, port):
+        """Instantiate and run subscriber data-store sync.
+
+        Args:
+            w_id (str): Workflow external ID.
+            reg (str): Registered workflow name.
+            host (str): Hostname of target workflow.
+            port (int): Port of target workflow.
+
+        """
+        self.w_subs[w_id] = WorkflowSubscriber(
+            reg,
+            host=host,
+            port=port,
+            context=self.workflows_mgr.context,
+            topics=self.topics
+        )
+        self.w_subs[w_id].loop.run_until_complete(
+            self.w_subs[w_id].subscribe(
+                process_delta_msg,
+                func=self._update_workflow_data,
+                w_id=w_id))
+
+    def _update_workflow_data(self, topic, delta, w_id):
+        """Manage and apply incoming data-store deltas.
+
+        Args:
+            topic (str): topic of published data.
+            delta (object): Published protobuf message data container.
+            w_id (str): Workflow external ID.
+
+        """
+        # wait until data-store is populated for this workflow
+        if w_id not in self.data:
+            loop_cnt = 0
+            while loop_cnt < self.INIT_DATA_WAIT_TIME:
+                if w_id in self.data:
+                    break
+                time.sleep(self.INIT_DATA_RETRY_DELAY)
+                loop_cnt += 1
+                continue
+        if topic == 'shutdown':
+            self.log.debug(f'shutdown({w_id})')
+            self._delta_store_to_queues(w_id, topic, delta)
+            # update the status to stopped and set the status message
+            self._update_contact(
+                w_id,
+                status=WorkflowStatus.STOPPED.value,
+                status_msg=self._get_status_msg(w_id, False),
+            )
+            # close connections
+            self.disconnect_workflow(w_id)
+            return
+        self._apply_all_delta(w_id, delta)
+        self._delta_store_to_queues(w_id, topic, delta)
+
+    def _clear_data_field(self, w_id, field_name):
+        if field_name == WORKFLOW:
+            self.data[w_id][field_name].Clear()
+        else:
+            self.data[w_id][field_name].clear()
+
+    def _apply_all_delta(self, w_id, delta):
+        """Apply the AllDeltas delta."""
+        for field, sub_delta in delta.ListFields():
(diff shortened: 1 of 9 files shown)


oliver-sanders:
@dwsutherland if you get the chance to take a look at this one that would be great.

cylc/uiserver/workflows_mgr.py
```diff
@@ -89,7 +93,28 @@ async def workflow_request(
         return (req_context, None)
 
 
-class WorkflowsManager:
+async def wrap(*coros):
```
MetRonnie, line 96: Might I suggest a more descriptive name like `run_coros`

cylc/uiserver/workflows_mgr.py
```diff
@@ -293,12 +361,15 @@ async def multi_request(
...
                 command,
                 multi_args.get(w_id, args),
                 timeout,
-            ) for w_id in workflows
-            if w_id in self.active
+            )
+            for w_id in workflows
+            if w_id in self.workflows
+            # skip stopped workflows
+            and self.workflows.get(w_id, {}).get('req_client')
```
MetRonnie, line 372: Think the first condition can be skipped
```suggestion
            # skip stopped workflows
            if self.workflows.get(w_id, {}).get('req_client')
```

MetRonnie (commented):
Partial review (6/9 files viewed)

MetRonnie (approved):

datamel (approved):
I have run the ui-server with branch since yesterday without any traceback, deleting stale workflows and have seen no traceback. I have started and stopped multiple workflows, workflow state appears to be working correctly.

I have also read the code, only one query.
Thanks @oliver-sanders

cylc/uiserver/workflows_mgr.py
```diff
@@ -188,31 +228,31 @@ async def _register(self, wid, flow, is_active):
...
     async def _disconnect(self, wid):
         """Disconnect from a running workflow."""
-        try:
-            client = self.active[wid]['req_client']
-        except KeyError:
-            pass
-        else:
-            with suppress(IOError):
-                client.stop(stop_loop=False)
+        self.uiserver.data_store_mgr.disconnect_workflow(wid)
```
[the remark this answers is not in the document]
MetRonnie: It's not an async method though?

oliver-sanders:
Keep this one open for a mo, I'll slide in an extra commit to fix the issue Ronnie recently reported.

oliver-sanders:
Last commit combines the stop and disconnect methods. The stop method contained the logic for updating the contact data. In the event a workflow went from running to stopped unexpectedly the stop method was not called leaving behind broken contact data and an incorrect status string.

newnew

cylc/uiserver/workflows_mgr.py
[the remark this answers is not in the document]
datamel: Sorry what I meant was should we be marking `disconnect_workflow` as async and so await this?
Not entirely sure why `register_workflow` is async but this one isn't when neither look IO-bound.
Not fully got my head around using asyncio for multi-threading CPU-bound operations yet. Will have to use some training time on this.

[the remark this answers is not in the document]
MetRonnie: Ah, not sure why `register_workflow` is defined `async`

MetRonnie (approved):

MetRonnie:
Hmm, I was still able to reproduce the mismatch between `status` and `statusMsg` (`stopped` vs `Running`, discussed on Teams chat) for a failed integration test workflow. Does it need a cylc-ui fix?

oliver-sanders:
Don't think it's related to the UI, if you run the UIS in debug mode (`cylc gui --debug`) the output should be enough to work out what's going wrong, will take another look at the logic.

oliver-sanders:
Couldn't replicate with 10 runs of the integration tests.

datamel:
> Couldn't replicate with 10 runs of the integration tests.

I'm having a go at replicating now. Will report back in a bit.

oliver-sanders:
ok, what does the debug output say?

datamel:
> ok, what does the debug output say?

Debug output looked normal as far as I could see. It is a very transient problem as have only seen it once. I have closely inspected the status_msg code and can't spot anything obvious. Only thing I can think of is nfs maybe causing problems? 
@MetRonnie have you been able to reliably reproduce?

oliver-sanders:
Can you send me the debug output, it contains a record of every method called on the data store which should reveal the issue.

datamel:
> Can you send me the debug output, it contains a record of every method called on the data store which should reveal the issue.
```
[D 2022-03-03 14:09:18.510 CylcUIServer] register_workflow(~/cit-20220303T234411+0935/integration.test_examples/test_logging/834bfec6-9afb-11ec-93eb-0050569a149e)
[D 2022-03-03 14:09:18.518 CylcUIServer] Could not connect to ~/cit-20220303T234411+0935/integration.test_examples/test_logging/834bfec6-9afb-11ec-93eb-0050569a149e: cit-20220303T234411+0935/integration.test_examples/test_logging/834bfec6-9afb-11ec-93eb-0050569a149e is not running
```

oliver-sanders:
Brill, that focuses me on connection failure, there should be only one side-effect of connection failure which is the workflow getting added to `self.workflows`. Changed that so it only gets added if connection succeeded.

MetRonnie:
Turns out my reproduceable way of reproducing the statusMsg mismatch is not as reproduceable as I thought when I first reproduced it

MetRonnie:
Unfortunately I am still able to reproduce (though not with 100% reliability). Only thing in the uiserver logs:
```
[D 2022-03-03 17:27:49.998 CylcUIServer] register_workflow(~rdutta/cit-20220304T030009+0935/integration.test_examples/test_run/e02ef01a-9b16-11ec-97a2-0050569a0f64)
[D 2022-03-03 17:27:50.009 CylcUIServer] connect_workflow(~rdutta/cit-20220304T030009+0935/integration.test_examples/test_run/e02ef01a-9b16-11ec-97a2-0050569a0f64)
cit-20220304T030009+0935/integration.test_examples/test_run/e02ef01a-9b16-11ec-97a2-0050569a0f64 is not running
```

oliver-sanders:
Cheers, will have a crack at reproducing tomorrow.

datamel:
I'm still (occasionally) hitting the running state issue, with some traceback in the log now: 

```
[E 2022-03-04 11:11:29.735 CylcUIServer] Failed to update entire local data-store of a workflow
      File "/~/cylc-flow/cylc/flow/workflow_files.py", line 520, in detect_old_contact_file
        os.unlink(fname)
    FileNotFoundError: [Errno 2] No such file or directory: '/~/cylc-run/cit-20220304T204549+0935/integration.test_scan_api/test_scan_cleans_stuck_contact_files/-crashed-/.service/contact'
    
      File "/~/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 79, in workflow_request
        result = await client.async_request(
      File "/~/cylc-flow/cylc/flow/network/client.py", line 198, in async_request
        self.timeout_handler()
      File "/~/cylc-flow/cylc/flow/network/client.py", line 325, in _timeout_handler
        detect_old_contact_file(workflow, contact_data)
      File "/~/cylc-flow/cylc/flow/workflow_files.py", line 523, in detect_old_contact_file
        raise CylcError(
    cylc.flow.exceptions.CylcError: Failed to remove old contact file: "/~/cylc-run/cit-20220304T204549+0935/integration.test_scan_api/test_scan_cleans_stuck_contact_files/-crashed-/.service/contact"
    [Errno 2] No such file or directory: '/~/cylc-run/cit-20220304T204549+0935/integration.test_scan_api/test_scan_cleans_stuck_contact_files/-crashed-/.service/contact'


[E 2022-03-04 11:13:54.202 ServerApp] Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7f3f1c048d30>>, <Task finished name='Task-1' coro=<WorkflowsManager.run() done, defined at /~/cylc-uiserver/cylc/uiserver/workflows_mgr.py:397> exception=gaierror(-2, 'Name or service not known')>)
      File "/~/miniconda3/envs/cylc8/lib/python3.9/site-packages/tornado/ioloop.py", line 741, in _run_callback
        ret = callback()
      File "/~/miniconda3/envs/cylc8/lib/python3.9/site-packages/tornado/ioloop.py", line 765, in _discard_future_result
        future.result()
      File "/~/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 410, in run
        await self.update()
      File "/~/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 336, in update
        await asyncio.gather(*tasks)
      File "/~/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 99, in run_coros_in_order
        await coro
      File "/~/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 232, in _connect
        flow['req_client'] = WorkflowRuntimeClient(flow['name'])
      File "/~/cylc-flow/cylc/flow/network/client.py", line 131, in __init__
        host, port, _ = get_location(workflow)
      File "/~/cylc-flow/cylc/flow/network/__init__.py", line 83, in get_location
        host = get_fqdn_by_host(host)
      File "/~/cylc-flow/cylc/flow/hostuserutil.py", line 265, in get_fqdn_by_host
        return HostUtil.get_inst().get_fqdn_by_host(target)
      File "/~/cylc-flow/cylc/flow/hostuserutil.py", line 171, in get_fqdn_by_host
        return self._get_host_info(target)[0]
      File "/~/cylc-flow/cylc/flow/hostuserutil.py", line 135, in _get_host_info
        self._host_exs[target] = socket.gethostbyname_ex(target)
    socket.gaierror: [Errno -2] Name or service not known: 'not-a-host.no-such.domain'
```

oliver-sanders:
The first traceback looks like a possible bug in cylc-flow due to the malformed path:

```
Errno 2] No such file or directory: '/~/cylc-run/cit-20220304T204549+0935/integration.test_scan_api
```

The second traceback is to be expected:

```
socket.gaierror: [Errno -2] Name or service not known: 'not-a-host.no-such.domain'
```

Since `not-a-host.no-such.domain` does not exist (likely an integration test simulating an unreachable host).

oliver-sanders:
https://github.com/cylc/cylc-flow/issues/4734

oliver-sanders:
Thanks @MetRonnie that diff helped pin down the error.

The issue occurred when the contact file was present but the workflow was not contactable resulting in a timeout.

Such timeouts are currently ignored by the data store and do not result in traceback. Since the contact file is still present the workflow manager will not send a disconnect request (which is correct).

This results in the mixed status because the presence of the contact file says the workflow is running (used for the status string), however, the connection to the workflow is what sets the workflow status and this has failed so the status remains as it was before (stopped).

Now we know what the problem is this diff should result in the error every time.

```diff
diff --git a/cylc/flow/network/client.py b/cylc/flow/network/client.py
index ae917c41a..2781fe156 100644
--- a/cylc/flow/network/client.py
+++ b/cylc/flow/network/client.py
@@ -114,7 +114,7 @@ class WorkflowRuntimeClient(ZMQSocketBase):
 
     """
 
-    DEFAULT_TIMEOUT = 5.  # 5 seconds
+    DEFAULT_TIMEOUT = 0.  # 5 seconds
 
     def __init__(
         self,
```

datamel:
> The first traceback looks like a possible bug in cylc-flow due to the malformed path:

I had replaced my usual $HOME expansion in the traceback with a `~` (for security), so not sure it is a malformed path.
Unless you think the `-crashed-` could be causing problems?

Sorry for not making that clear, https://github.com/cylc/cylc-flow/issues/4734 can probably just be closed.

oliver-sanders:
Ah, I thought we had failed to expand `~` resulting in `/~` causing the problem.

In that case it's still a cylc-flow issue but presumably caused by the script trying to remove the contact file only to find that either the contact file or .service dir is no longer there?

It is perfectly possible that the contact file could have been removed since the client was created (even if that's a short time window another client started at the same time could have beat this one to removing the file, or a clean in progress at the time or whatever), so probably just need to quietly suppress this traceback.

oliver-sanders:
PR up https://github.com/cylc/cylc-flow/pull/4735

oliver-sanders:
Ok, last commit should do it, really hard to test properly without running a workflow so came up with a kinda half-test which at least ensures that `disconnect_workflow` is called in the event `connect_workflow` fails.

datamel (approved):
Looks good. Read code, checked out and tested as running smoothly. 
Minor typo, not a blocker though. Thanks @oliver-sanders 
____
Note there is some new warnings on this branch on running pytest:
```
Task was destroyed but it is pending!
task: <Task pending name='Task-29' coro=<WorkflowsManager.run() running at ~/cylc-uiserver/cylc/uiserver/workflows_mgr.py:412> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f43db947bb0>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at ~/miniconda3/envs/cylc8/lib/python3.9/site-packages/tornado/ioloop.py:688]>
Task was destroyed but it is pending!
task: <Task pending name='Task-64' coro=<WorkflowsManager.run() running at ~/cylc-uiserver/cylc/uiserver/workflows_mgr.py:412> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f43db98d310>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at ~/miniconda3/envs/cylc8/lib/python3.9/site-packages/tornado/ioloop.py:688]>
```

oliver-sanders:
Will take a look at those warnings...

oliver-sanders:
The scan task was being sent the stop signal, however, the UIS was shutting down before the stop had completed. Strangely, even when the scan task has completed Tornado needs an `asyncio.sleep` to detect that it has completed. Next commit should address this.

datamel (commented):
Very strange. This latest commit has not solved the problem for me, I have run the tests as a set and am still getting the warnings. I also have run `cylc/uiserver/tests/test_auth.py` and `cylc/uiserver/tests/test_graphql.py` individually - these seem to be the source of the error. However on master they run without warning. Can you reproduce? Perhaps a race condition if not.

oliver-sanders:
I get lots of other warnings but not the specific `task: <Task pending name='Task-29' coro=<WorkflowsManager.run()` one.

datamel:
> I get lots of other warnings but not the specific `task: <Task pending name='Task-29' coro=<WorkflowsManager.run()` one.
> 
> (python 3.7)

Ahha I expect my python version is too high - Python 3.9.10

oliver-sanders:
I'll see if I can replicate with 3.9

oliver-sanders:
Managed to replicate, it's to do with the jupyter_server version. It looks like more recent jupyter_server is not calling the `stop_extension` endpoint (hopefully only in tests?).

oliver-sanders:
The warnings are due to a regression in jupyter_server:

https://github.com/jupyter-server/jupyter_server/pull/588/files#diff-3cd116aa8a4bea11f638c9209aa3b58e13a3a0aad127d310b0198b4931ca962cR305-R306

cylc/uiserver/workflows_mgr.py
```diff
@@ -115,10 +136,14 @@ def __init__(self, uiserver, log, context=None, run_dir=None):
             | api_version(f'=={API}')
         )
 
-    def spawn_workflow(self):
-        """Start/spawn a workflow."""
-        # TODO - Spawn workflows
-        pass
+        # queue for requesting new scans
+        self.queue = asyncio.Queue()
```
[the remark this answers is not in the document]
oliver-sanders: * Converted to private attr.
* Added type hint.
* Added comments.

MetRonnie (approved):
Seems to shrug off my attempts to create horrible workflow/contact file states

oliver-sanders:
Good I think that's everything addressed?

cylc/uiserver/data_store_mgr.py
```diff
@@ -408,3 +389,76 @@ async def entire_workflow_update(self, ids=None):
...
+        flow = delta.workflow.updated
+        flow.id = w_id
+        flow.stamp = f'{w_id}@{delta.workflow.time}'
+        if contact_data:
+            # update with contact file data
+            flow.name = contact_data['name']
+            flow.owner = contact_data['owner']
+            flow.host = contact_data[CFF.HOST]
+            flow.port = int(contact_data[CFF.PORT])
+            # flow.pub_port = int(contact_data[CFF.PUBLISH_PORT])
```
datamel, line 419: Is this needed?

cylc/uiserver/data_store_mgr.py
```diff
@@ -408,3 +389,76 @@ async def entire_workflow_update(self, ids=None):
...
+            flow.api_version = int(contact_data[CFF.API])
+        else:
+            # wipe pre-existing contact-file data
+            w_tokens = Tokens(w_id)
+            flow.owner = w_tokens['user']
+            flow.name = w_tokens['workflow']
+            flow.host = ''
+            flow.port = 0
+            # flow.pub_port = 0
```
datamel, line 428: ditto

datamel (commented):
Just a quick query about the commenting of the publish port.

cylc/uiserver/data_store_mgr.py
oliver-sanders:
```suggestion
```

cylc/uiserver/data_store_mgr.py
oliver-sanders: No, this attribute is unused, I had commented it out but not removed it, removed now.

cylc/uiserver/data_store_mgr.py
oliver-sanders: Ditto.

MetRonnie (approved):

datamel (approved):

PR #329, opened 2022-03-22
fix bugbear

wxtim:
Equivelent to https://github.com/cylc/cylc-flow/pull/4746

### DIFF
--- a/cylc/uiserver/authorise.py
+++ b/cylc/uiserver/authorise.py
@@ -225,7 +225,7 @@ class Authorization:
         allowed_operations.discard("")
         return allowed_operations
 
-    @lru_cache(maxsize=128)
+    @lru_cache(maxsize=128)  # noqa B019 TODO?
     def get_permitted_operations(self, access_user: str):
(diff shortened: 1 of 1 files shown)


cylc/uiserver/authorise.py
```diff
@@ -225,7 +225,7 @@ def get_access_user_permissions_from_owner_conf(
         allowed_operations.discard("")
         return allowed_operations
 
-    @lru_cache(maxsize=128)
+    @lru_cache(maxsize=128)  # noqa B019 TODO?
```
oliver-sanders, line 228: TODO: this cache can keep instances of the `Authorization` class alive after the instances would otherwise have been garbage collected.

Since only one `Authorisation` instance is created per UIS and since it lives for the lifetime of the UIS this is not a problem, however, we may as well do it properly.

oliver-sanders: https://github.com/cylc/cylc-uiserver/issues/330

oliver-sanders (approved):

PR #336, opened 2022-03-28
Fix lack of upper pin for `rx` in `graphql-core<2.3.0`

MetRonnie:

I tried (in fresh conda env)
```
$ conda install 'cylc-flow==8.0rc2'
$ pip install cylc-uiserver
```
And when I ran `cylc gui` I got
```
ImportError: cannot import name 'extensionmethod' from 'rx.internal'
```
(`rx` is a dependency of [`graphql-core`](https://github.com/graphql-python/graphql-core-legacy) which is a dependency of `graphene`)

Turns out with the conda installation of `cylc-flow` I ended up with `rx 3.2.0`, because I ended up with `graphql-core 2.2` and that version does not have an upper pin on `rx` (just a lower pin `rx>=1.6`). (The latest `graphene 2.1.9` has a pin `graphql-core>=2.1,<3`) It was not until `graphql-core 2.3.0` that the pin `rx>=1.6,<2` was introduced.

**Requirements check-list**
- [x] Does not need tests
- [x] No change log entry required (users don't need to know the pain I've been through).
- [x] This dependency is already included in [cylc-uiserver conda-forge repository](https://github.com/conda-forge/cylc-uiserver-feedstock)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,6 +58,10 @@ install_requires =
     tornado>=6.1.0  # matches jupyter_server value
     traitlets>=5
 
+    # Fix lack of upper pin for rx in graphql-core<2.3.0 (remove when
+    # upgrading graphene to 3.0):
+    rx<2
+
 [options.packages.find]
 include = cylc*
 
--- a/cylc/uiserver/authorise.py
+++ b/cylc/uiserver/authorise.py
@@ -420,15 +420,17 @@ class AuthorizationMiddleware:
         op_name = self.get_op_name(info.field_name, info.operation.operation)
         # It shouldn't get here but worth checking for zero trust
         if not op_name:
-            self.auth_failed(current_user, op_name, 400,
-                             "Operation not in schema.")
+            self.auth_failed(
+                current_user, op_name, http_code=400,
+                msg="Operation not in schema."
+            )
         try:
             authorised = self.auth.is_permitted(current_user, op_name)
         except Exception:
             # Fail secure
             authorised = False
         if not authorised:
-            self.auth_failed(current_user, op_name, 403)
+            self.auth_failed(current_user, op_name, http_code=403)
         if (info.operation.operation in Authorization.ASYNC_OPS
                 or iscoroutinefunction(next_)):
             return self.async_resolve(next_, root, info, **args)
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -58,6 +58,9 @@ install_requires =
     tornado>=6.1.0  # matches jupyter_server value
     traitlets>=5
 
+    # Fix lack of upper pin for rx in graphql-core<2.3.0:
+    rx<2
```
MetRonnie, line 62: The other option is to pin `graphql-core>=2.3.0`?

oliver-sanders:
Note we have been pinning Rx in the conda-forge feedstock for a while:

https://github.com/conda-forge/cylc-uiserver-feedstock/blob/8659d2aa62649dc3c692aae7ea92ee1b1c695c26/recipe/meta.yaml#L47-L48

https://github.com/conda-forge/cylc-feedstock/pull/3#issuecomment-660716268

setup.cfg
```diff
@@ -58,6 +58,10 @@ install_requires =
     tornado>=6.1.0  # matches jupyter_server value
     traitlets>=5
 
+    # Fix lack of upper pin for rx in graphql-core<2.3.0 (remove when
+    # upgrading graphene to 3.0):
+    rx<2
```
MetRonnie, line 63: The other option is to pin `graphql-core>=2.3.0`?

oliver-sanders: graphql-core 2.3.2 was the last v2 release, due to graphene-tornado and graphql-ws (both stuck on v2) we cannot move up to v3.

oliver-sanders: The conda-forge feedstock lower-pins the dep at 1.6, should this be done here too?

https://github.com/conda-forge/cylc-uiserver-feedstock/blob/d45ddfd4cb0acf9b4a30a50e6f3294caea16dfc0/recipe/meta.yaml#L48

MetRonnie: - There is already [a requirement on `graphql-core>=2.0,<3` in `graphene 2.1.0 ` and above](https://github.com/graphql-python/graphene/blob/05d7e61b844b574f59d68ead86ed40431843d51d/setup.py#L90), so raising the lower boundary would give us `graphql-core>=2.3.0,<3`

- There is already [a requirement on `rx>=1.6` in `graphql-core 2.2` and above](https://github.com/graphql-python/graphql-core-legacy/blob/4b1f697ac068a5ff83c35cd818a430e35a0b14b4/setup.py#L24), so we shouldn't need to duplicate the lower boundary necessarily

oliver-sanders: ok

oliver-sanders (approved):

PR #344, opened 2022-04-22
Workaround lru_cache decorator on method garbage collection problem

### LINKED ISSUE #330: Fix Authorization class caching
This cache can keep instances of the `Authorization` class alive after the instances would otherwise have been garbage collected.

Since only one `Authorisation` instance is created per UIS and since it lives for the lifetime of the UIS this is not a problem, however, we may as well do it properly.

_Originally posted by @oliver-sanders in https://github.com/cylc/cylc-uiserver/pull/329#discussion_r831966835_

MetRonnie:
These changes close #330 

`@lru_cache`, if used on a method, retains the class instance beyond the point it would normally be garbage collected. This workaround is from https://stackoverflow.com/a/68550238/3217306

This PR also includes some minor GH Actions test workflow updates and removes pytest-asyncio (see https://github.com/pytest-dev/pytest-asyncio/issues/257#issuecomment-1017566870)

**Requirements check-list**
- [x] Does not need tests
- [x] No change log entry required (invisible to users)
- [x] No core dependency changes.

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -407,6 +407,24 @@ class DataStoreMgr:
         status_msg=None,
         pruned=False,
     ):
+        """Update the data store with information from the contact file.
+
+        Args:
+            w_id: Workflow ID.
+            contact_data: Contact file data dictionary.
+            status: Workflow status (e.g. "running").
+            status_msg: Workflow status message (e.g. "will stop at 2000").
+            pruned: ?
+
+        Returns:
+            True if the contact data is successfully updated or False if
+            the workflow is no longer in the store (e.g. has been removed).
+
+        """
+        if w_id not in self.data:
+            # workflow has been removed - do nothing
+            return False
+
         delta = DELTAS_MAP[ALL_DELTAS]()
         delta.workflow.time = time.time()
         flow = delta.workflow.updated
@@ -442,6 +460,8 @@ class DataStoreMgr:
         self._apply_all_delta(w_id, delta)
         # Queue delta for subscription push
         self._delta_store_to_queues(w_id, ALL_DELTAS, delta)
+
+        return True
 
     def _get_status_msg(self, w_id: str, is_active: bool) -> str:
         """Derive a status message for the workflow.
--- a/cylc/uiserver/tests/test_data_store_mgr.py
+++ b/cylc/uiserver/tests/test_data_store_mgr.py
@@ -145,8 +145,18 @@ async def test_update_contact_no_contact_data(
     w_id = Tokens(user='user', workflow='workflow_id').id
     api_version = 0
     await data_store_mgr.register_workflow(w_id=w_id, is_active=False)
-    data_store_mgr._update_contact(w_id=w_id, contact_data=None)
+    assert data_store_mgr._update_contact(w_id=w_id, contact_data=None)
     assert api_version == data_store_mgr.data[w_id]['workflow'].api_version
+
+
+async def test_update_contact_no_workflow(
+    data_store_mgr: DataStoreMgr
+):
+    """Ensure _update_contact doesn't error if the workflow is missing.
+
+    This can happen if the workflow is removed.
+    """
+    assert not data_store_mgr._update_contact(w_id='elephant')
 
 
 async def test_update_contact_with_contact_data(


MetRonnie:
MacOS test failure is pretty strange, doesn't *seem* related to this change, but doesn't fail on master though

MetRonnie:
Well I don't know why it's resolved itself. Possibly needed `asyncio_mode=strict` in `pytest.ini`

cylc/uiserver/workflows_mgr.py
```diff
@@ -436,6 +436,7 @@ async def run(self):
             except RuntimeError:
                 # RuntimeError may be raised if the event loop is closed
                 break
+            await asyncio.sleep(0)
```
MetRonnie, line 439: Is this correct? Or should it deliberately block the event loop?

datamel:
> Well I don't know why it's resolved itself. Possibly needed `asyncio_mode=strict` in `pytest.ini`
> 
> However, still getting these printed out after the end of the pytest output
> 
> ```
> Task was destroyed but it is pending!
> task: <Task pending name='Task-57' coro=<WorkflowsManager.run() running at
>     /home/runner/work/cylc-uiserver/cylc-uiserver/cylc/uiserver/workflows_mgr.py:435>
>     wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f79f6f8ce50>()]>
>     cb=[IOLoop.add_future.<locals>.<lambda>() at
>     /opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/tornado/ioloop.py:688]>
> ```

I have raised this with @oliver-sanders before, from memory, it only occurs when the python version is >=3.9. See conversation after https://github.com/cylc/cylc-uiserver/pull/324#pullrequestreview-903049598

MetRonnie:
It's happening for Python 3.7 on this branch and master: https://github.com/cylc/cylc-uiserver/runs/6160205704?check_suite_focus=true#step:7:80

I saw this comment on #324:

> The warnings are due to a regression in jupyter_server:
>
> https://github.com/jupyter-server/jupyter_server/pull/588/files#diff-3cd116aa8a4bea11f638c9209aa3b58e13a3a0aad127d310b0198b4931ca962cR305-R306

Is it worth raising an issue there?

cylc/uiserver/workflows_mgr.py
oliver-sanders: I don't think so, the `await` in `await self._queue.get()` should yield control in the same way as `await asyncio.sleep(0)`.

pyproject.toml
```diff
@@ -22,3 +22,4 @@ testpaths = [
 markers = [
     'integration: tests which run servers and try to connect to them'
 ]
+asyncio_mode = 'strict'  # 'auto' seems to conflict with pytest-tornasync
```
oliver-sanders, line 25: Makes sense, this async means tests need an explicit `pytest.mark.asyncio`? Need to check that relevant tests are decorated correctly otherwise they will be skipped from here on.

MetRonnie:
Hmm, adding `@pytest.mark.asyncio` to the async tests that were missing them brings back this error:
```
FAILED cylc/uiserver/tests/test_auth.py::test_cylc_handler -
    tornado.simple_httpclient.HTTPTimeoutError: Timeout during request
```
Seems to be only the tests that use the `jp_fetch` fixture

MetRonnie:
> After a brief look at `pytest-tornasync` it seems the plugin has a similar behaviour to `pytest-asyncio` with `async_mode=auto`: It tries to run all tests that are coroutines in a `tornado.IoLoop`. When both plugins try to do the same thing, unexpected things happen.

https://github.com/pytest-dev/pytest-asyncio/issues/257#issuecomment-1017566870

oliver-sanders:
> When both plugins try to do the same thing, unexpected things happen.

Makes sense, thanks for investigating this.

oliver-sanders (approved):

datamel (commented):
I have manually tested this, working well. This was the initial implementation, @oliver-sanders just checking there is not a reason for the requested change, I am struggling to remember... https://github.com/cylc/cylc-uiserver/pull/204/commits/be7d04dfd643491da078175474a1a489639dd77e#r728858282

MetRonnie:
https://github.com/cylc/cylc-uiserver/pull/204#discussion_r728858282 - that was before we realised about the problem of using `@lru_cache` to decorate methods: https://stackoverflow.com/q/33672412/3217306

datamel:
This branch has conflicts now. If you want to get this rebased I can give it a final check before it goes in @MetRonnie?

MetRonnie:
Will merge master into this branch rather than rebase - makes the conflict resolution diff easy to see

PR #348, opened 2022-05-16
data-store: log workflow events to info level

oliver-sanders:
* To assist with debugging the UIS under normal operation bump the logging of workflow events (registered, connected, disconnected, etc events) from debug to info levels.
* Centralised workflow event logging, included more argument info & gave all events a prefix to help us to `grep` them out when needed.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Does not need tests (why?).
- [x] No change log entry required (why? e.g. invisible to users).
- [x] No documentation update required.
- [x] No dependency changes.

PR #349, opened 2022-05-17
logging: configure default logging config

oliver-sanders:
Provide a default logging setup for the UIS:

* Bump Traitlets to >=5.2.1 (needed for logging_config) trait.
* Add default logging config to write to ~/.cylc/uiserver/uiserver.log at the INFO level.
* Document in the readme for now.

Note this only captures the CylcUIServer part of the log, the base ServerApp log (i.e. the jupyter_server log) and any extensions (e.g. jupyter_lab) are excluded (can be easily added if desired but not much use from the Cylc perspective).

In combination with #348 this should capture workflow state transitions and tracebacks which are useful for debugging and monitoring for issues.

**Requirements check-list**
- [x] Does not need tests (please test manually)
- [x] No documentation update required (waiting on feature to be documented in jupyter_server)
- [x] No dependency changes.


oliver-sanders:
(rebased)

oliver-sanders:
@wxtim pointed out that uncaptured tracebacks go to the ServerApp log rather than the CylcUIServer log so I've added logging for the ServerApp too (pointing at a different file).

cylc/uiserver/jupyter_config.py
```diff
@@ -57,8 +57,64 @@
...
+c.CylcUIServer.logging_config = {
+    'version': 1,
+    'handlers': {
+        'file': {
+            'class': 'logging.handlers.RotatingFileHandler',
+            'level': 'INFO',
+            'filename': str(USER_CONF_ROOT / 'uiserver.log'),
+            'mode': 'a',
+            'backupCount': 5,
+            'maxBytes': 10485760,
```
oliver-sanders, line 78: Copied from Cylc, perhaps we should reduce this a little?

oliver-sanders:
Could potentially bump this to the 1.0.2 milestone (only adds logging) but no pressure from our side (happy to use dev builds for the moment).

cylc/uiserver/jupyter_config.py
wxtim: Given the amount of logging the UIS doesn't produce I feel that this is likely to be overkill.

oliver-sanders:
Rebased, added changelog.

wxtim (approved):
- [x] Code looks sane
- [x] Working as advertised

PR #368, opened 2022-06-22
clean: fix clean mutation

oliver-sanders:
* fix --rm-dirs option (should be None if unset, not ['']).
* Log clean outcome.
* Fix post-clean & post-play scan (should use .scan not .update).

**Requirements check-list**


cylc/uiserver/resolvers.py
```diff
@@ -51,7 +51,7 @@ class InvalidSchemaOptionError(CylcError):
 CLEAN = 'clean'
 OPT_CONVERTERS: Dict[str, Dict[str, Union[Callable, None]]] = {
     CLEAN: {
-        'rm': lambda opt, value: ('rm_dirs', [value]),
+        'rm': lambda opt, value: ('rm_dirs', [value] if value else None),
```
oliver-sanders, line 54: This resulted in `['']` which is "Truthy" effectively disabling the clean operation.

cylc/uiserver/resolvers.py
```diff
@@ -235,17 +235,20 @@ async def clean(cls, workflows, args, workflows_mgr, executor, log):
...
             else:
-                if future == WORKFLOW_RUNNING_MSG:
+                if string == WORKFLOW_RUNNING_MSG:
                     log.error(ServiceFileError(WORKFLOW_RUNNING_MSG))
                     return cls._error(WORKFLOW_RUNNING_MSG)
+                log.info(string)
 
         # trigger a re-scan
-        await workflows_mgr.update()
+        await workflows_mgr.scan()
```
oliver-sanders, line 251: This is due to a change in the scanning logic, the `.scan` interface requests rather than performs a scan preventing concurrent scan operations from fighting over the same workflows.

PR #367, opened 2022-06-17
logging: redirect cylc-flow log to uis log file

oliver-sanders:
Output of cleaning a workflow:

**Before**

```log
2023-02-16T13:12:17 INFO     Starting Cylc UI Server    
2023-02-16T13:12:17 INFO     Serving UI from: ~/cylc-uiserver/cylc/uiserver/ui/1.4.0    
2023-02-16T13:12:21 INFO     [data-store] register_workflow('~/generic/run1', False)    
2023-02-16T13:12:24 INFO     Cleaning ~/generic/run1    
2023-02-16T13:12:25 INFO     [data-store] unregister_workflow('~/generic/run1')    
2023-02-16T13:12:25 INFO     [data-store] _purge_workflow('~/generic/run1')    
2023-02-16T13:12:25 INFO     [data-store] disconnect_workflow('~/generic/run1', update_contact=False)  
```

**After**

```log
2023-02-16T13:10:50 INFO     Starting Cylc UI Server
2023-02-16T13:10:50 INFO     Serving UI from: /net/home/h06/osanders/cylc-uiserver/cylc/uiserver/ui/1.4.0
2023-02-16T13:11:25 INFO     [data-store] register_workflow('~osanders/generic/run1', False)
2023-02-16T13:11:31 INFO     Cleaning ~osanders/generic/run1
2023-02-16T13:11:31 INFO     No workflow database for generic/run1 - will only clean locally
2023-02-16T13:11:31 INFO     Removing symlink and its target directory: /home/h06/osanders/cylc-run/generic/run1/work -> /net/spice/scratch/osanders/cylc-run/generic/run1/work
2023-02-16T13:11:31 INFO     Removing symlink and its target directory: /home/h06/osanders/cylc-run/generic/run1/share -> /net/data/users/osanders/cylc-run/generic/run1/share
2023-02-16T13:11:31 INFO     Removing symlink and its target directory: /home/h06/osanders/cylc-run/generic/run1/log -> /net/data/users/osanders/cylc-run/generic/run1/log
2023-02-16T13:11:31 INFO     Removing directory: /home/h06/osanders/cylc-run/generic/run1
2023-02-16T13:11:31 INFO     Removing directory: /home/h06/osanders/cylc-run/generic/_cylc-install
2023-02-16T13:11:31 INFO     Removing directory: /home/h06/osanders/cylc-run/generic
2023-02-16T13:11:31 INFO     Removing directory: /net/spice/scratch/osanders/cylc-run/generic/run1
[...]

PR #323, opened 2022-02-24
Add Cylc Clean Support

wxtim:
These changes close #299

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Appropriate tests are included (unit and/or functional).
- [x] Appropriate change log entry included.
- [x] No documentation update required.
- [x] No dependency changes.

PR #369, opened 2022-06-22
Fix cylc clean & tidy implementation

### LINKED ISSUE #366: Clean fails silently
Caused by my suggestions in #323

**Pull requests welcome!**

MetRonnie:
These changes close #366, follow-up to #323, supersedes #368 

- Fix cylc clean mutation functionality (bug caused by incorrect handling of "rm" field)
- Tidy implementation of cylc clean resolver
- Fix lack of default run mode in cylc play mutation form
- Fix post-clean & post-play scan (should use `.scan()` not `.update()`) (cherry-picked from #368)

There is no logging of what gets cleaned, but this will come in #367 

**Requirements check-list**
- [x] No change log entry required (follow-up to unreleased #323)

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -231,7 +231,7 @@ class Services:
                 return cls._error(msg)
 
         # trigger a re-scan
-        await workflows_mgr.update()
+        await workflows_mgr.scan()
         return cls._return("Workflow(s) cleaned")
 
     @classmethod
@@ -304,7 +304,7 @@ class Services:
                     'Workflow started'
                 )
         # trigger a re-scan
-        await workflows_mgr.update()
+        await workflows_mgr.scan()
         return response
 
 
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -178,10 +178,10 @@ def _schema_opts_to_api_opts(
 
 
 def _clean(workflow_ids, opts):
-    """Helper funciton to call `cylc clean`.
+    """Helper function to call `cylc clean`.
 
     Execute this function inside of an "executor" (note this is why we have
-    to set up asynio here).
+    to set up asyncio here).
     """
     loop = asyncio.new_event_loop()
     asyncio.set_event_loop(loop)
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -33,7 +33,7 @@ from cylc.flow.exceptions import (
 )
 from cylc.flow.network.resolvers import BaseResolvers
 from cylc.flow.scripts.clean import CleanOptions
-from cylc.flow.workflow_files import init_clean
+from cylc.flow.scripts.clean import run
 
 if TYPE_CHECKING:
     from concurrent.futures import Executor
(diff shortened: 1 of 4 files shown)


cylc/uiserver/resolvers.py
```diff
@@ -19,48 +19,42 @@
...
+    from cylc.flow.option_parsers import Options
     from cylc.uiserver.workflows_mgr import WorkflowsManager
 
 
 # show traceback from cylc commands
 DEBUG = True
-CLEAN = 'clean'
-OPT_CONVERTERS: Dict[str, Dict[str, Union[Callable, None]]] = {
-    CLEAN: {
-        'rm': lambda opt, value: ('rm_dirs', [value]),
```
MetRonnie, line 54: As Oliver commented in #368 
> This resulted in `['']` which is "Truthy" effectively disabling the clean operation.

cylc/uiserver/resolvers.py
```diff
@@ -172,80 +166,72 @@ def func(schema_key, schema_value):
...
+                elif isinstance(exc, WorkflowFilesError):  # Expected error
+                    msg = str(exc)
+                else:  # Unexpected error
+                    msg = f"{type(exc).__name__}: {exc}"
+                    log.exception(msg)
+                return cls._error(msg)
 
         # trigger a re-scan
-        await workflows_mgr.update()
+        await workflows_mgr.scan()
```
MetRonnie, line 234: As Oliver said in #368 
> This is due to a change in the scanning logic, the `.scan` interface requests rather than performs a scan preventing concurrent scan operations from fighting over the same workflows.

cylc/uiserver/resolvers.py
```diff
@@ -172,80 +166,72 @@ def func(schema_key, schema_value):
     converters = OPT_CONVERTERS[schema]
     api_opts = {}
     for opt, value in schema_opts.items():
-        # All valid options should be in SCHEMA_TO_API:
-        if opt not in converters:
-            raise InvalidSchemaOptionError(
-                f'{opt} is not a valid option for Cylc Clean'
-            )
```
MetRonnie, line 179: Instead of this we can rely on the Python API `Options` class from `cylc.flow.optparse` (in this instance, `CleanOptions`)

wxtim (approved):

oliver-sanders:
Fond another issue :(

Providing workflow IDs as globs doesn't work e.g: `cylc clean '*'` 

I think we should use the `cylc.flow.scripts.clean.run` interface rather than the more internal `init_clean` which will solve this.

https://github.com/MetRonnie/cylc-uiserver/pull/2

cylc/uiserver/resolvers.py
hjoliver: That's vaguely amusing. A bag of nothing is falsey, but a bag containing a single nothing is truthy.

cylc/uiserver/resolvers.py
```diff
@@ -172,80 +166,86 @@ def func(schema_key, schema_value):
...
+    return schema(**api_opts)
 
 
-def _clean(tokens, opts):
-    """Run Cylc Clean using `cylc.flow.workflow_files` api.
+def _clean(workflow_ids, opts):
+    """Helper funciton to call `cylc clean`.
+
+    Execute this function inside of an "executor" (note this is why we have
+    to set up asynio here).
```
hjoliver, line 184:
```suggestion
    to set up asyncio here).
```

hjoliver (approved):
LGTM, but I didn't have time to be very thorough. Another sanity check wouldn't go amiss.

cylc/uiserver/resolvers.py
MetRonnie: At least it's not JS where a bag of nothing is truthy!

oliver-sanders: Because:

```python
>>> len('')
0
>>> len([])
0
>>> len([''])
1
```

Issue #240, opened 2021-08-31
cylc gui: detect if a server is already running


oliver-sanders:
As of #230 we have reincarnated the `cylc gui` command.

If a server is already running and the user runs the `cylc gui` command again, a second server will be started which is likely not what they wanted.

This is likely to be a bugbear for Cylc 7 users who may type `cylc gui` automatically.

We could detect when a server is running using a flag file (e.g. in the `~/.cylc/hub` dir) and:

* Open a new browser tab pointing at the running server rather than starting a new one.
* Or send a message to the UI app telling it to change its focus to the selected workflow:
  * See #239

We should take a look at Jupyter Server to see if it has any related functionality.

Related to #238, #239

oliver-sanders:
Closed by #370

Issue #239, opened 2021-08-31
cylc gui: add workflow ID argument


oliver-sanders:
As of #230 we have reincarnated the `cylc gui` command.

In Cylc 7 the `cylc gui` command took a single (optional) argument, the workflow ID, which, if supplied, opened the GUI ready to view the provided workflow.

We could add this argument back and have it open the URL `<server>/workflows/<workflow>` as the default URL.

Related to: #238, #240

PR #370, opened 2022-06-23
Add gui workflow arg

datamel:
This adds functionality for opening the gui at a specific workflow when running `cylc gui <workflow_name>`

These changes close [#239](https://github.com/cylc/cylc-uiserver/issues/239) and [#240](https://github.com/cylc/cylc-uiserver/issues/240)

This PR also restricts the number of gui instances to one, by searching for pid file created by jupyter server.

- [x] Any dependency changes applied to `setup.cfg`. No dependency changes.
- [x] Does not need tests (This is inbuilt Jupyter tech (default url)).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,9 @@ ones in. -->
 
 ### Enhancements
 
+[#370](https://github.com/cylc/cylc-uiserver/pull/370) - Gui command adapted:
+`cylc gui workflow_id` is now supported and will open gui at that workflow.
+
 [#376](https://github.com/cylc/cylc-uiserver/pull/376) -
 UIServer logs are now archived. The five most recent logs are retained (located
 in `~/.cylc/uiserver/log/`). A new log is created with each UIServer instance.
@@ -45,16 +48,6 @@ Traitlets `logging_config` "trait" for more information.
 
 [#323](https://github.com/cylc/cylc-uiserver/pull/323) -
 `cylc clean` made available for runs and files within runs.
-
-### Enhancements
-
-[#370](https://github.com/cylc/cylc-uiserver/pull/370) - Gui command adapted:
-`cylc gui workflow_name` is now supported and will open gui at that workflow.
-
-### Enhancements
-
-[#370](https://github.com/cylc/cylc-uiserver/pull/370) - Gui command adapted:
-`cylc gui workflow_name` is now supported and will open gui at that workflow.
 
 -------------------------------------------------------------------------------
 ## __cylc-uiserver-1.0.3 (<span actions:bind='release-date'>Released 2022-05-31</span>)__
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,6 +51,11 @@ Traitlets `logging_config` "trait" for more information.
 [#370](https://github.com/cylc/cylc-uiserver/pull/370) - Gui command adapted:
 `cylc gui workflow_name` is now supported and will open gui at that workflow.
 
+### Enhancements
+
+[#370](https://github.com/cylc/cylc-uiserver/pull/370) - Gui command adapted:
+`cylc gui workflow_name` is now supported and will open gui at that workflow.
+
 -------------------------------------------------------------------------------
 ## __cylc-uiserver-1.0.3 (<span actions:bind='release-date'>Released 2022-05-31</span>)__
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,11 @@ Traitlets `logging_config` "trait" for more information.
 [#323](https://github.com/cylc/cylc-uiserver/pull/323) -
 `cylc clean` made available for runs and files within runs.
 
+### Enhancements
+
+[#370](https://github.com/cylc/cylc-uiserver/pull/370) - Gui command adapted:
+`cylc gui workflow_name` is now supported and will open gui at that workflow.
+
 -------------------------------------------------------------------------------
 ## __cylc-uiserver-1.0.3 (<span actions:bind='release-date'>Released 2022-05-31</span>)__
 
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -54,6 +54,7 @@ Cylc specific configurations are documented here.
 """
 
 from concurrent.futures import ProcessPoolExecutor
+import contextlib
 import getpass
 from pathlib import Path, PurePath
 import sys
@@ -525,3 +526,12 @@ class CylcUIServer(ExtensionApp):
         self.data_store_mgr.executor.shutdown(wait=False)
         # Destroy ZeroMQ context of all sockets
         self.workflows_mgr.context.destroy()
+        self.clean_pid_file()
+
+    @staticmethod
+    def clean_pid_file():
+        """Remove pid file if it exists."""
+        pid_file = Path("~/.cylc/uiserver/pid").expanduser()
+        print(f"Removing Process Id file from {pid_file.parent}.")
+        with contextlib.suppress(FileNotFoundError):
+            pid_file.unlink()
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -132,7 +132,7 @@ class CylcUIServer(ExtensionApp):
     Cylc - A user interface for monitoring and controlling Cylc workflows.
     '''  # type: ignore[assignment]
     examples = '''
-        cylc gui    # start the cylc GUI
+        cylc gui [workflow]   # start the cylc GUI [at the workflow page]
     '''  # type: ignore[assignment]
     config_file_paths = get_conf_dir_hierarchy(
         [
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -126,7 +126,7 @@ class CylcUIServer(ExtensionApp):
 
     name = 'cylc'
     app_name = 'cylc-gui'
-    default_url = "/cylc"
+    subcommands = []
     load_other_extensions = True
     description = '''
     Cylc - A user interface for monitoring and controlling Cylc workflows.
@@ -506,7 +506,11 @@ class CylcUIServer(ExtensionApp):
         """Change the jinja templating environment."""
 
     @classmethod
-    def launch_instance(cls, argv=None, **kwargs):
+    def launch_instance(cls, argv=None, workflow_id=None, **kwargs):
+        if workflow_id:
+            cls.default_url = f"/cylc/#/workflows/{workflow_id}"
+        else:
+            cls.default_url = "/cylc"
         if argv is None:
             # jupyter server isn't expecting to be launched by a Cylc command
             # this patches some internal logic
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -54,8 +54,8 @@ Cylc specific configurations are documented here.
 """
 
 from concurrent.futures import ProcessPoolExecutor
-import contextlib
 import getpass
+import os
 from pathlib import Path, PurePath
 import sys
 from typing import List
@@ -105,6 +105,8 @@ from cylc.uiserver.schema import schema
 from cylc.uiserver.websockets.tornado import TornadoSubscriptionServer
 from cylc.uiserver.workflows_mgr import WorkflowsManager
 
+INFO_FILES_DIR = Path(USER_CONF_ROOT / "info_files")
+
 
 class PathType(TraitType):
     """A pathlib traitlet type which allows string and undefined values."""
@@ -129,10 +131,15 @@ class CylcUIServer(ExtensionApp):
     app_name = 'cylc-gui'
     load_other_extensions = True
     description = '''
-    Cylc - A user interface for monitoring and controlling Cylc workflows.
+    Cylc gui - A user interface for monitoring and controlling Cylc workflows.
     '''  # type: ignore[assignment]
     examples = '''
-        cylc gui [workflow]   # start the cylc GUI [at the workflow page]
+        cylc gui                  # Start the Cylc GUI (At the dashboard page)
+cylc gui [workflow]       # Start the Cylc GUI (at the workflow page),
+cylc gui --new [workflow] # Start the Cylc GUI (at the workflow page), with a
+                            new instance. By default, if there is an existing
+                            gui instance, Cylc will use that.
+
     '''  # type: ignore[assignment]
     config_file_paths = get_conf_dir_hierarchy(
         [
@@ -515,7 +522,9 @@ class CylcUIServer(ExtensionApp):
             # jupyter server isn't expecting to be launched by a Cylc command
             # this patches some internal logic
             argv = sys.argv[2:]
+        os.environ["JUPYTER_RUNTIME_DIR"] = str(INFO_FILES_DIR)
         super().launch_instance(argv=argv, **kwargs)
+        del os.environ["JUPYTER_RUNTIME_DIR"]
 
     async def stop_extension(self):
         # stop the async scan task
@@ -526,12 +535,3 @@ class CylcUIServer(ExtensionApp):
         self.data_store_mgr.executor.shutdown(wait=False)
         # Destroy ZeroMQ context of all sockets
         self.workflows_mgr.context.destroy()
-        self.clean_pid_file()
-
-    @staticmethod
-    def clean_pid_file():
-        """Remove pid file if it exists."""
-        pid_file = Path("~/.cylc/uiserver/pid").expanduser()
-        print(f"Removing Process Id file from {pid_file.parent}.")
-        with contextlib.suppress(FileNotFoundError):
-            pid_file.unlink()
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -58,6 +58,7 @@ import getpass
 import os
 from pathlib import Path, PurePath
 import sys
+from textwrap import dedent
 from typing import List
 
 from pkg_resources import parse_version
@@ -133,14 +134,13 @@ class CylcUIServer(ExtensionApp):
     description = '''
     Cylc gui - A user interface for monitoring and controlling Cylc workflows.
     '''  # type: ignore[assignment]
-    examples = '''
-        cylc gui                  # Start the Cylc GUI (At the dashboard page)
-cylc gui [workflow]       # Start the Cylc GUI (at the workflow page),
-cylc gui --new [workflow] # Start the Cylc GUI (at the workflow page), with a
-                            new instance. By default, if there is an existing
-                            gui instance, Cylc will use that.
+    examples = dedent('''
+    cylc gui                  # Start the Cylc GUI (At the dashboard page)
+    cylc gui [workflow]       # Start the Cylc GUI (at the workflow page)
+    cylc gui --new [workflow] # Start the Cylc GUI (at the workflow page), with
+                                a new instance.
 
-    '''  # type: ignore[assignment]
+    ''')  # type: ignore[assignment]
     config_file_paths = get_conf_dir_hierarchy(
         [
             SITE_CONF_ROOT,  # site configuration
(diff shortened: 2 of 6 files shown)


cylc/uiserver/scripts/gui.py
```diff
@@ -20,8 +20,31 @@
...
+            continue
+        try:
+            # workflow_id, _, _ = parse_id(
+            # arg, constraint='workflows', infer_latest_runs=True)
+
+            # The following line needs to be replaced with the line above in
+            # order to interpret the workflow_id. This, causes problems with
+            # the ServerApp Event loop.
+
+            workflow_id = arg
```
datamel, line 44: !!!
This pr currently only accepts the full workflow id e.g. ` cylc gui workflow-name/run4` rather than `cylc gui workflow-name`

This is due to a problem calling `parse_id`. The async is resulting in the ServerApp error of `There is no current event loop in thread 'MainThread'.`

oliver-sanders (commented):
LGTM, works great

cylc/uiserver/scripts/gui.py
```diff
@@ -20,8 +20,66 @@
...
+            workflow_id, _, _ = task.result()
+            argv = tuple(x for x in argv if x != arg)
+            sys.argv.remove(arg)
+        except InputError:
+            print(f"Workflow '{arg}' does not exist.")
+            break
+    return CylcUIServer.launch_instance(argv or None, workflow_id=workflow_id)
+
+
+def check_pid(pid_file: Path):
```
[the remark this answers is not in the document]
datamel: `--new` option now added. I haven't found a reliable way of auto opening up a gui with an existing instance. This is due to the full url not being stored in the pid, not sure how to get a reliable port etc as this is written before launch.

cylc/uiserver/app.py
```diff
@@ -126,13 +129,12 @@ class CylcUIServer(ExtensionApp):
...
     app_name = 'cylc-gui'
-    default_url = "/cylc"
     load_other_extensions = True
     description = '''
     Cylc - A user interface for monitoring and controlling Cylc workflows.
     '''  # type: ignore[assignment]
     examples = '''
-        cylc gui    # start the cylc GUI
+        cylc gui [workflow]   # start the cylc GUI [at the workflow page]
     '''  # type: ignore[assignment]
```
hjoliver, line 138: Maybe mention what happens if it is already running?

Also, if it is already running `cylc gui --help` doesn't work (I just get the PID message).

Also, the new `--new` option doesn't feature in the help text.

cylc/uiserver/app.py
```diff
@@ -522,3 +528,23 @@ async def stop_extension(self):
...
+                content = json.load(f)
+                del content[str(pid)]
+                json_content = json.dumps(content)
+            with open(pid_file, "w") as f:
+                f.write(json_content)
+        with contextlib.suppress(FileNotFoundError):
+            file_size = os.path.getsize(pid_file)
+            if file_size in [0, 2]:  # delete empty pid file
+                pid_file.unlink()
+                print(f"Removing Process Id file from {pid_file.parent}")
```
hjoliver, line 550:
```suggestion
                print(f"Removed Process ID file from {pid_file.parent}")
```
(Or perhaps change the line order :grin: )

cylc/uiserver/scripts/gui.py
```diff
@@ -20,8 +20,107 @@
...
 
 
 def main(*argv):
-    return CylcUIServer.launch_instance(argv or None)
+    new_gui = False
+    for arg in argv:
+        if arg == "--new":
+            new_gui = True
+            argv = tuple(x for x in argv if x != "--new")
+            sys.argv.remove("--new")
```
hjoliver, line 44: Could be simpler. And, dangerous to change list `argv` during iteration over `argv`?
```suggestion
    new_gui = "--new" in argv
    if new_gui:
        argv = tuple(x for x in argv if x != "--new")
        sys.argv.remove("--new")
```

cylc/uiserver/scripts/gui.py
```diff
@@ -20,8 +20,107 @@
...
+            continue
+        try:
+            loop = asyncio.new_event_loop()
+            task = loop.create_task(
+                parse_id_async(
+                    arg, constraint='workflows'))
+            loop.run_until_complete(task)
+            loop.close()
+            workflow_id, _, _ = task.result()
+            argv = tuple(x for x in argv if x != arg)
```
hjoliver, line 60: Changing iteratee (`argv`) again, during iteration?

hjoliver (commented):
Looks good, tested as working, just a couple of small questions.

wxtim:
I have picked this up and am running with in an attempt to get it over the line.

hjoliver:
Thanks @wxtim - as noted on Element chat, no need to bust a gut if short of time to get this in the release (next release is OK).

> Where is the argument documentation coming from?

I haven't had a chance to look today, sorry.  

> is it OK that `cylc gui --new myflow` works but that `cylc gui myflow --new` doesn't, or is this unexpected?

Personally I think options *should* come before args (otherwise, logically, it sort-of looks like `--new` is an option for the arg rather than the command). However, some (most?) programs are flexible on that, and the cylc CLI is, so ideally this command should do too for consistency.

oliver-sanders:
The CLI docs are coming from Jupyter Server, options are configured "traits", these double as both CLI options and configurations. E.G. the CylcUIServer adds a trait called `--ui-build-dir` which can subsequently be set on the CLI or in the config.

This issue is effectively about extending the Jupyter Server CLI. If the modifications are light we can bodge them in, but if we want greater control over the CLI appearance they we would need to create our own argument parser here, parse the args, then reformat `argv` before calling JupyterServer. Icky but doable.

```python
#!psudopython

def get_option_parser():
    ...
    return parser

@cli_function
def main(..., workflow_id):
    ...
    cylc_opts, jps_opts = parser.parse_known_args()
    if not gui_running or cylc_opts.new:
        argv = ['cylc', 'uiserver', *jps_opts]
        # do Mel's URL magic
        return CylcUIServer.launch_instance(argv)
    else:
        browser.open(get_gui_url())
```

wxtim:
I may or may not come back to this. Probably not a priority for me.

datamel:
Change of implementation:

Originally we were creating and maintaining our own PID file, to establish if an instance of the gui was running. It turns our Jupyter Server creates two files containing the information that we require. These, by default, are housed in `~/.local/share/jupyter/runtime`. The two files created per run are: 
* `jpserver-<pid>.json` and
* `jpserver-<pid>-open.html`

The json file contains info such as url, port number, version of jpserver, token etc...
The html file, when called will open the gui at the original page that it was opened at (adding the workflow_id if cylc gui was called with a wofkflow_id argument.)

In this PR, I have changed where this file is created (by setting an environment variable "JUPYTER_RUNTIME_DIR"). Now it can be found in `.cylc/uiserver/info_files`. Jupyter server cleans these files when shut down correctly but if killed unexpectedly they can remain (I've a bucketload in my dir). I have added a funtion that removes these files if the selected gui process is no longer running - it doesn't loop through all files as I think this could slow down the gui, but if it happens to select one that needs cleaning, it cleans it.

If a gui is started without the `--new` option it will check for these jp server runtime files and adapt the html file to open at the correct page in the gui.

datamel:
I will update the logic to consider no-browser option.

CHANGES.md
```diff
@@ -10,6 +10,14 @@ creating a new release entry be sure to copy & paste the span tag with the
 `actions:bind` attribute, which is used by a regex to find the text to be
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
+
+## __cylc-uiserver-1.2.0 (<span actions:bind='release-date'>Upcoming</span>)__
+
+### Enhancements
+
+[#370](https://github.com/cylc/cylc-uiserver/pull/370) - Gui command adapted:
+`cylc gui workflow_name` is now supported and will open gui at that workflow.
```
oliver-sanders, line 19: Always use workflow ID (name is something special which most people should not use):

```suggestion
`cylc gui workflow_id` is now supported and will open gui at that workflow.
```

cylc/uiserver/scripts/gui.py
```diff
@@ -20,8 +20,165 @@
...
+    new_gui = cylc_opts.new
+    if new_gui:
+        sys.argv.remove(CLI_OPT_NEW)
+    workflow_id = None
+    remove_these_args = []
+    for arg in jp_server_opts:
+        if arg.startswith('-'):
+            continue
+        try:
+            loop = asyncio.new_event_loop()
```
oliver-sanders, line 59: We should only need one loop here. Suggest doing something like this:

```python
def main(*argv):
    ... = asyncio.run(parse_args())

async def parse_args():
    ...
    for arg in jp_server_opts:
        ...
        workflow_id, *_ = await parse_id_async(...)

File: cylc/uiserver/scripts/gui.py [hunk 9]
```diff
@@ -20,8 +20,165 @@
...
+            loop.run_until_complete(task)
+            loop.close()
+            workflow_id, _, _ = task.result()
+            remove_these_args.append(arg)
+        except InputError:
+            remove_these_args.append(arg)
+            print(
+                f"Workflow '{arg}' does not exist. Opening gui at dashboard.",
+                file=sys.stderr)
+        break
```
oliver-sanders, line 72: Might as well error here, other Cylc commands would fail with a no such workflow type error (saves the need for the redirect to dashboard thing).

cylc/uiserver/scripts/gui.py
```diff
@@ -20,8 +20,165 @@
...
+    return CylcUIServer.launch_instance(
+        jp_server_opts or None, workflow_id=workflow_id
+    )
+
+
+def get_arg_parser():
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument(
+        CLI_OPT_NEW,
+        action='store_true',
```
oliver-sanders, line 100: Would be good to separate the arg parsing and jpserver launching bits of the code.

cylc/uiserver/scripts/gui.py
```diff
@@ -20,8 +20,165 @@
...
+    file.write_text(updated_text)
+
+
+def get_valid_existing_gui(gui_files: List[str]):
+    """Finds an existing gui to reopen"""
+    for file in gui_files:
+        pid_match = re.search(r"jpserver-(\d+)-open.html", file)
+        if pid_match is not None:
+            pid = pid_match[1]
+            if pid_exists(int(pid)):
```
oliver-sanders, line 155: Spanner to throw into the works, the process might not be running on this server.

Does Jupyter Server have a mechanism for checking if a server is still running?

datamel: I will have a look and report back, I had forgotten that case.

datamel: So, Jupyter Server does have a command for checking if a server is running. I don't think this is of use to us though 😢 
The command is `jupyter server list`
- this checks for the existence of pid file (remember this doesn't get cleaned up if the server shuts down unexpectedly)
- it then does a `os.kill(pid, 0)` to check the process is running
- it defaults to process running in the case of `PermissionError(1, 'Operation not permitted')`, which I am meeting regularly as I have a load of stale files.
- the result is that it is claiming to have servers running when they are not.

If the process is running on a different server, maybe we can use the info in the pid file to pick the host out?
The info stored in the file is:
```
{
  "base_url": "/",
  "hostname": "localhost",
  "password": false,
  "pid": 11111,
  "port": 8888,
  "root_dir": "/net/home/../username",
  "secure": true,
  "sock": "",
  "token": "blah blah some big long token here",
  "url": "https://localhost:8888/",
  "version": "1.18.1"
} 
```

oliver-sanders (commented):
This doesn't seem to detect a gui launched via JupyterHub, likely my fault, can't remember?

cylc/uiserver/scripts/gui.py
```diff
@@ -20,10 +20,159 @@
...
+def parse_args_opts():
+    """Parse cli args
+
+    Separate JPserver args from Cylc specific ones.
+    Parse workflow id.
+    Raise error if invalid workflow provided
+    """
+    cylc_opts, jp_server_opts = get_arg_parser().parse_known_args()
+    # gui arg is stripped cylc end but need to re-strip after arg parsing here
+    jp_server_opts.remove('gui')
```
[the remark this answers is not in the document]
datamel: I've suppressed that value error. Was getting an index error on trying  `jp_server_opts[0]` and thought maybe this is neater than catching both?

datamel:
> This doesn't seem to detect a gui launched via JupyterHub, likely my fault, can't remember?

Looking into this now.

cylc/uiserver/scripts/gui.py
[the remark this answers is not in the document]
oliver-sanders: Tested running `jupyter-cylc` working now!

oliver-sanders (approved):
Tested, all working as expected.

One outstanding comment here - https://github.com/cylc/cylc-uiserver/pull/370/files/cd9796c448fc626d4eb877abb22ef1542d31656a#r952333283

Agree, Jupyter Hub is a different problem (wasn't thinking straight, that needs to be configured at the site-level).

Cheers!

oliver-sanders:
I tested after rebasing this onto master. Suggest maybe rebase-squashing to see the tests pass with latest UIS code.

cylc/uiserver/scripts/gui.py
```diff
@@ -20,10 +20,163 @@
...
+            update_html_file(gui_file, workflow_id)
+            if '--no-browser' not in sys.argv:
+                webbrowser.open(gui_file, autoraise=True)
+            return
+    return CylcUIServer.launch_instance(
+        jp_server_opts or None, workflow_id=workflow_id
+    )
+
+
+def raise_error(error: str, msg: str):
```
MetRonnie, line 77: I would call this `print_error` seeing as it doesn't actually raise an exception

cylc/uiserver/scripts/gui.py
```diff
@@ -20,10 +20,163 @@
...
+    if new_gui:
+        sys.argv.remove(CLI_OPT_NEW)
+    workflow_id = None
+    workflow_arg = []
+    for arg in jp_server_opts:
+        if arg.startswith('-'):
+            continue
+        else:
+            # argument not an option
+            workflow_arg.append(arg)
```
MetRonnie, line 106:
```suggestion
    workflow_arg = [
        arg for arg in jp_server_opts
        if not arg.startswith('-')
    ]
```

cylc/uiserver/scripts/gui.py
```diff
@@ -20,10 +20,163 @@
...
+
+def get_arg_parser():
+    """Cylc specific options"""
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument(
+        CLI_OPT_NEW,
+        action='store_true',
+        default=False,
+        dest='new'
+    )
```
MetRonnie, line 139: I am not seeing `--new` as a listed option in `cylc gui --help`

MetRonnie (commented):
Partial review

cylc/uiserver/scripts/gui.py
```diff
@@ -20,10 +20,163 @@
...
     init_log()
-    return CylcUIServer.launch_instance(argv or None)
+    new_gui = False
+    new_gui = CLI_OPT_NEW in argv
+    jp_server_opts, new_gui, workflow_id = parse_args_opts()
+    if '--help' not in sys.argv:
+        # get existing jpserver-<pid>-open.html files
+        # assume if this exists then the server is still running
+        # these files are cleaned by jpserver on shutdown
+        existing_guis = glob(os.path.join(INFO_FILES_DIR, "*open.html"))
```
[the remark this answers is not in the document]
datamel: There are currently only two types of files that are stored here. I think someone making a `*.open/html` file is just not going to happen - there are a million ways people can break Cylc by creating dodgy files, I think we are at greater risk of jupyter changing the format of the name if we restrict it so much.

MetRonnie (commented):
Aside from my suggestions, I have tested out and happy it's working

cylc/uiserver/app.py
```diff
@@ -126,13 +129,17 @@ class CylcUIServer(ExtensionApp):
...
     load_other_extensions = True
     description = '''
-    Cylc - A user interface for monitoring and controlling Cylc workflows.
+    Cylc gui - A user interface for monitoring and controlling Cylc workflows.
     '''  # type: ignore[assignment]
     examples = '''
-        cylc gui    # start the cylc GUI
+        cylc gui                  # Start the Cylc GUI (At the dashboard page)
+cylc gui [workflow]       # Start the Cylc GUI (at the workflow page),
+cylc gui --new [workflow] # Start the Cylc GUI (at the workflow page), with a
```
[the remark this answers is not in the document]
MetRonnie: I am seeing it there but not in the list of options. Is there a way to include it?
```
Options
=======
The options below are convenience aliases to configurable class-options,
as listed in the "Equivalent to" description-line of the aliases.
To see all configurable class-options for some <cmd>, use:
    <cmd> --help-all

--debug
    set log level to logging.DEBUG (maximize logging output)
    Equivalent to: [--Application.log_level=10]
--show-config
    Show the application's configuration (human-readable format)
    Equivalent to: [--Application.show_config=True]
--show-config-json
    Show the application's configuration (json format)
    Equivalent to: [--Application.show_config_json=True]
--generate-config
    generate default config file
    Equivalent to: [--JupyterApp.generate_config=True]
-y
    Answer yes to any questions instead of prompting.
    Equivalent to: [--JupyterApp.answer_yes=True]
--log-level=<Enum>
    Set the log level by value or name.
    Choices: any of [0, 10, 20, 30, 40, 50, 'DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL']
    Default: 30
    Equivalent to: [--Application.log_level]
--config=<Unicode>
    Full path of a config file.
    Default: ''
    Equivalent to: [--JupyterApp.config_file]

```

cylc/uiserver/app.py
```diff
@@ -126,13 +129,17 @@ class CylcUIServer(ExtensionApp):
...
+                            new instance. By default, if there is an existing
+                            gui instance, Cylc will use that.
+
     '''  # type: ignore[assignment]
```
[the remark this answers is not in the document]
datamel: I'll change the wording. The formatting is dealt with through traitlets and they indent and dedent it with `yield indent(dedent(self.examples.strip()))`. An additional dedent as suggested is having no effect for me.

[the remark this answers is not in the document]
MetRonnie: From my testing the `dedent` is needed to avoid it being like this:
```
   examples = '''
        cylc gui                  # ...
cylc gui [workflow]       # ...
```

cylc/uiserver/app.py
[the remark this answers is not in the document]
datamel: No, the options are populated by the definition of our config as traitlets. We cannot add --new as a traitlet option as the config is loaded after the --new would be required.

cylc/uiserver/app.py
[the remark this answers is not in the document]
datamel: That is really strange. Two screengrabs from me below with/without dedent:
image

image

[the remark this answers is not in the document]
datamel: Incidentally, the `If there is an existing GUI instance, Cylc will use that.` is required for both the `cylc gui` and the `cylc gui [workflow]` options. I actually think it is neater leaving it as it is, since it conveys the same information without repetition, what do you think?

cylc/uiserver/scripts/gui.py
```diff
@@ -20,10 +20,159 @@
...
+    INFO_FILES_DIR
+)
+
+CLI_OPT_NEW = "--new"
 
 
 def main(*argv):
     init_log()
-    return CylcUIServer.launch_instance(argv or None)
+    new_gui = CLI_OPT_NEW in argv
```
MetRonnie, line 53: This is not used because it is redefined in next line
```suggestion
```

MetRonnie (approved):

PR #380, opened 2022-08-17
Bump cylc-flow version

MetRonnie:
If the next uiserver release is a minor rather than maintenance one, then doesn't it need to have a dependency on the next cylc-flow minor release?

**Requirements check-list**
- [x] Does not need tests 
- [x] No change log entry required

### DIFF
--- a/cylc/uiserver/__init__.py
+++ b/cylc/uiserver/__init__.py
@@ -13,7 +13,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-__version__ = "1.1.1.dev"
+__version__ = "1.2.0.dev"
 
 import os
 from typing import Dict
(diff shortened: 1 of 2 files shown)


hjoliver:
> If the next uiserver release is a minor rather than maintenance one, then doesn't it need to have a dependency on the next cylc-flow minor release?

But I suppose we can't be sure yet that the next UIS release will not be a maintenance release (we might find a bug requiring a quick fix...).

MetRonnie:
I say we branch off the 1.1.0 tag in that case

oliver-sanders:
> **If the next uiserver release is a minor rather than maintenance one, then doesn't it need to have a dependency on the next cylc-flow minor release?**

Yes, this is on the cylc-admin issue template (metadata section) but got missed this time around. The same also applies to cylc-rose.

https://github.com/cylc/cylc-admin/issues/157

oliver-sanders:
(Note that the tests are run against the cylc-flow master branch)

MetRonnie:
Actually as there's currently the dependency on `cylc-flow==8.0.*`, it's currently installing cylc-flow master, then uninstalling and replacing with 8.0.1
https://github.com/cylc/cylc-uiserver/blob/5286d488c4a60afa82deba1d92dff400b8efb97b/.github/workflows/test.yml#L33-L36

hjoliver (approved):

PR #386, opened 2022-09-12
Override a buggy JupyterHub class method.

hjoliver:
Work around a bug in JupyterHub's `launch_instance` method.

These changes close #385 

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Any dependency changes applied to `setup.cfg`.
- [x] Does not need tests (why?).
- [x] Appropriate change log entry included.
- [x] No documentation update required.

PR #387, opened 2022-10-04
Merge 1.1.x into master

MetRonnie:
- #386
- ' #389


cylc/uiserver/__init__.py
```diff
@@ -13,7 +13,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-__version__ = "1.1.0"
+__version__ = "1.1.1.dev"
```
oliver-sanders: 1.2.0.dev?

MetRonnie: That isn't showing up when you choose all commits. It's on 1.2.0.dev

PR #394, opened 2022-10-28
fix unregister double update/delta

### LINKED ISSUE #393: deltas: updated delta issued after pruned delta
When a workflow is removed we would expect a single pruned delta with the ID of the workflow.

* pruned(id)

However, we currently get two deltas:

* updated(id), pruned(id)
* updated(id)

The first delta contains an update for the workflow which is erroneous in this case as nothing should have changed in the workflow data, however this update is probably harmless (as we apply updated *before* pruned).

The second delta is more of a problem as it contains updates for a workflow which no longer exists.

This is a major problem for https://github.com/cylc/cylc-ui/pull/1108 which avoids issues like https://github.com/cylc/cylc-ui/issues/903 by treating added & updated deltas as the same, inserting any missing nodes in the process.

<details>
<summary>Steps to reproduce:</summary>

Open GraphiQL to the following subscription:

```graphql
subscription {
  deltas {
    added {
      workflow {
        id
      }
    }
    updated {
      workflow {
        id
      }
    }
    pruned {
      workflow
    }
  }
}
```

Install a workflow (I'm using `cp` rather than `cylc install` to reduce the number of moving parts and ensure the addition is atomic):

```
cp -r ~/cylc-src/my-workflow ~/cylc-run
```

You should get an added delta for the workflow.

Remove the workflow (I'm using `rm` rather than `cylc clean` to reduce the number of moving parts and ensure the removal is atomic):

```
rm -r ~/cylc-run/my-workflow
```

</details>

### Evidence:

subs

Screenshot 2022-10-27 at 10 53 06

dwsutherland:
These changes close #393 

**Requirements check-list**

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -172,11 +172,21 @@ class DataStoreMgr:
 
         Call this when a workflow has stopped.
         """
-        if update_contact:
+        disconnect_msg = self._get_status_msg(w_id, False)
+        if (
+            update_contact
+            and w_id in self.data
+            and (
+                self.data[w_id][WORKFLOW].status != (
+                    WorkflowStatus.STOPPED.value
+                )
+                or self.data[w_id][WORKFLOW].status_msg != disconnect_msg
+            )
+        ):
             self._update_contact(
                 w_id,
                 status=WorkflowStatus.STOPPED.value,
-                status_msg=self._get_status_msg(w_id, False),
+                status_msg=disconnect_msg,
             )
         if w_id in self.w_subs:
             self.w_subs[w_id].stop()
@@ -254,12 +264,6 @@ class DataStoreMgr:
                 continue
         if topic == 'shutdown':
             self._delta_store_to_queues(w_id, topic, delta)
-            # update the status to stopped and set the status message
-            self._update_contact(
-                w_id,
-                status=WorkflowStatus.STOPPED.value,
-                status_msg=self._get_status_msg(w_id, False),
-            )
             # close connections
             self.disconnect_workflow(w_id)
             return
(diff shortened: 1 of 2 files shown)


oliver-sanders (approved):
LGTM, thanks for the quick fix, this is a big help with the graph view PR.

Re-ran my test again, the second erroneous updated delta is now gone!

The first combined updated/pruned delta makes sense as the contact data is being wiped (correctly). As long as the UI applies updated before pruned (which is does) this is harmless, so all good.

```json
{
  "data": {
    "deltas": {
      "added": {
        "workflow": null
      },
      "updated": {
        "workflow": {
          "id": "~osanders/foo",
          "status": "",
          "host": ""
        }
      },
      "pruned": {
        "workflow": "~osanders/foo"
      }
    }
  }
}
```

oliver-sanders:
@dwsutherland, we've got the dual branch thinggy set up here too. Could you either rebase onto 1.1.x or bump the milestone to 1.2.0.

Happy with either as this issue doesn't appear to cause any side-effects with the current UI data-store (but does with the new one which will ship with UIS 1.2.0).

datamel (approved):
I'm happy with this change and compared with behaviour in the issue.

One small query from me... I'm thinking we can send `update_contact=False` from the call to `disconnect_workflow` at line 264, since we are updating the contact just before the call. 
Happy to approve change, since I don't believe this will cause any harm. Thanks @dwsutherland

dwsutherland:
> @dwsutherland, we've got the dual branch thinggy set up here too. Could you either rebase onto 1.1.x or bump the milestone to 1.2.0.

Rebased and target changed to `1.1.x`

dwsutherland:
> I'm thinking we can send update_contact=False from the call to disconnect_workflow at line 264, since we are updating the contact just before the call.

The same call is made to update the contact within `disconnect_workflow` so I just removed update..
I also added some extra conditions on the contact update (status or msg has to be different), because I realised the workflow manager will also try to disconnect post delta disconnect.

datamel (approved):
I've checked the last commit and I'm happy for this to be merged.

PR #397, opened 2022-12-05
Cat Log Subscription Service

datamel:
This PR provides the data for https://github.com/cylc/cylc-ui/issues/1094.

Sibling PR  https://github.com/cylc/cylc-ui/pull/1158

**Requirements check-list**
- [x] Contains logically grouped changes (willl rebase after review).

### DIFF
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -24,7 +24,6 @@ import socket
 from typing import TYPE_CHECKING, Callable, Union, Dict
 
 from graphene_tornado.tornado_graphql_handler import TornadoGraphQLHandler
-from graphene.types.schema import Schema
 from graphql import get_default_backend
 from graphql_ws.constants import GRAPHQL_WS
 from jupyter_server.base.handlers import JupyterHandler
@@ -35,7 +34,6 @@ from cylc.flow.scripts.cylc import (
     get_version as get_cylc_version,
     list_plugins as list_cylc_plugins,
 )
-from cylc.flow.network.graphql import CylcGraphQLBackend
 
 from cylc.uiserver.authorise import Authorization, AuthorizationMiddleware
 from cylc.uiserver.resolvers import Resolvers
@@ -369,13 +367,10 @@ class UIServerGraphQLHandler(CylcAppHandler, TornadoGraphQLHandler):
 class SubscriptionHandler(CylcAppHandler, websocket.WebSocketHandler):
     """Endpoint for performing GraphQL subscriptions."""
     # No authorization decorators here, auth handled in AuthorizationMiddleware
-    def initialize(self, sub_server, resolvers,
-                   backend=None, schema=None, sub_statuses=None):
+    def initialize(self, sub_server, resolvers, sub_statuses=None):
         self.queue: Queue = Queue(100)
         self.subscription_server: TornadoSubscriptionServer = sub_server
         self.resolvers: Resolvers = resolvers
-        self.backend: CylcGraphQLBackend = backend or get_default_backend()
-        self.schema: Schema = schema
         self.sub_statuses: Dict = sub_statuses
 
     def select_subprotocol(self, subprotocols):
@@ -398,17 +393,15 @@ class SubscriptionHandler(CylcAppHandler, websocket.WebSocketHandler):
         with suppress(ValueError, SyntaxError):
             message_dict = ast.literal_eval(message)
             op_name = ''
+            op_id = message_dict.get("id", None)
             with suppress(KeyError):
                 op_name = message_dict['payload']['operationName']
             if (message_dict['type'] == 'start'
                 and op_name
                     and message_dict['payload']['operationName'] == 'LogData'):
-                op_id = message_dict["id"]
                 self.sub_statuses[op_id] = 'start'
-            if message_dict['type'] == 'stop':
-                op_id = message_dict["id"]
+            if message_dict['type'] == 'stop' and op_id in self.sub_statuses:
                 self.sub_statuses[op_id] = 'stop'
-
         await self.queue.put(message)
 
     async def recv(self):
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -348,16 +348,11 @@ class UIServerGraphQLHandler(CylcAppHandler, TornadoGraphQLHandler):
     @web.authenticated  # type: ignore[arg-type]
     async def execute(self, *args, **kwargs) -> 'ExecutionResult':
         # Use own backend, and TornadoGraphQLHandler already does validation.
-        # subscription = 'subscription { logs }'
-        # result = await self.schema.subscribe(subscription)
-        # async for item in result:
-        #     print(item.data['logs'])
         return await self.schema.execute(
             *args,
             backend=self.backend,
             variable_values=kwargs.get('variables'),
             validate=False,
-            allow_subscriptions=True,
             **kwargs,
         )
 
@@ -386,13 +381,19 @@ class SubscriptionHandler(CylcAppHandler, websocket.WebSocketHandler):
 
     @websockets_authenticated  # noqa: A003
     def open(self, *args, **kwargs):  # noqa: A003
+        print("opening web socket!!!!!!!!!!!!!!!")
         IOLoop.current().spawn_callback(
             self.subscription_server.handle,
             self,
             self.context,
         )
 
-    async def on_message(self, message):     
+    def on_close(self):
+        print("closing web socket!!!!!!!!!!!!!!!")
+
+        print("HEre Is Me in Da Close")
+
+    async def on_message(self, message):
         await self.queue.put(message)
 
     async def recv(self):
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from asyncio import Queue
+from contextlib import suppress
 from functools import wraps
 import json
 import getpass
@@ -364,12 +365,13 @@ class UIServerGraphQLHandler(CylcAppHandler, TornadoGraphQLHandler):
 class SubscriptionHandler(CylcAppHandler, websocket.WebSocketHandler):
     """Endpoint for performing GraphQL subscriptions."""
     # No authorization decorators here, auth handled in AuthorizationMiddleware
-    def initialize(self, sub_server, resolvers, backend=None, schema=None):
+    def initialize(self, sub_server, resolvers, backend=None, schema=None, sub_statuses=None):
         self.queue = Queue(100)
         self.subscription_server = sub_server
         self.resolvers = resolvers
         self.backend = backend or get_default_backend()
         self.schema = schema
+        self.sub_statuses = sub_statuses
 
     def select_subprotocol(self, subprotocols):
         return GRAPHQL_WS
@@ -392,7 +394,24 @@ class SubscriptionHandler(CylcAppHandler, websocket.WebSocketHandler):
         pass
 
     async def on_message(self, message):
-       # print(f"on message web socket!!!!!!!!!!!!!!!")
+        # print(f"on message web socket!!!!!!!{message}!!!!!!!!")
+        with suppress(ValueError):
+            import ast
+            message_dict = ast.literal_eval(message)
+            op_name = ''
+            with suppress(KeyError):
+                op_name = message_dict['operationName']
+            if (message_dict['type'] == 'start' and
+                op_name and
+                    message_dict['operationName'] == 'LogData'):
+                print(f"updated {op_id}....in onmessage..{self.sub_statuses}")
+                op_id = message_dict["id"]
+                self.sub_statuses[op_id] = 'start'
+                print(f"in start and status is now {self.sub_statuses}")
+            if message_dict['type'] == 'stop':
+                op_id = message_dict["id"]
+                self.sub_statuses[op_id] = 'stop'
+                print(f"on message: {self.sub_statuses}")
         await self.queue.put(message)
 
     async def recv(self):
@@ -411,4 +430,5 @@ class SubscriptionHandler(CylcAppHandler, websocket.WebSocketHandler):
                 self.get_current_user()
             ).get('name'),
             'ops_queue': {},
+            'sub_statuses': self.sub_statuses
         }
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -20,9 +20,10 @@ import json
 import getpass
 import os
 import socket
-from typing import TYPE_CHECKING, Callable, Union
+from typing import TYPE_CHECKING, Callable, Union, Dict
 
 from graphene_tornado.tornado_graphql_handler import TornadoGraphQLHandler
+from graphene.types.schema import Schema
 from graphql import get_default_backend
 from graphql_ws.constants import GRAPHQL_WS
 from jupyter_server.base.handlers import JupyterHandler
@@ -33,10 +34,12 @@ from cylc.flow.scripts.cylc import (
     get_version as get_cylc_version,
     list_plugins as list_cylc_plugins,
 )
+from cylc.flow.network.graphql import CylcGraphQLBackend
 
 from cylc.uiserver.authorise import Authorization, AuthorizationMiddleware
+from cylc.uiserver.resolvers import Resolvers
 from cylc.uiserver.websockets import authenticated as websockets_authenticated
-
+from cylc.uiserver.websockets.tornado import TornadoSubscriptionServer
 if TYPE_CHECKING:
     from graphql.execution import ExecutionResult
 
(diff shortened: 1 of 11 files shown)


datamel:
@dwsutherland, I've put you down as a reviewer, hope that is okay. I've made a couple of changes to tornado and I think you might be the best person to check I'm not inadvertently breaking something!

cylc/uiserver/handlers.py
```diff
@@ -386,6 +390,18 @@ def open(self, *args, **kwargs):  # noqa: A003
...
 
     async def on_message(self, message):
+        with suppress(ValueError, SyntaxError):
+            message_dict = ast.literal_eval(message)
+            op_name = ''
+            op_id = message_dict.get("id", None)
+            with suppress(KeyError):
+                op_name = message_dict['payload']['operationName']
+            if (message_dict['type'] == 'start'
+                and op_name
```
oliver-sanders, line 400: Do we need the `op_name`?

Could this be:

```python
self.sub_statuses[op_id] = message_dict.get('type')
```

(no need to check if the operation name is `LogData`?)

cylc/uiserver/handlers.py
```diff
@@ -386,6 +390,18 @@ def open(self, *args, **kwargs):  # noqa: A003
...
+                    and message_dict['payload']['operationName'] == 'LogData'):
+                self.sub_statuses[op_id] = 'start'
+            if message_dict['type'] == 'stop' and op_id in self.sub_statuses:
+                self.sub_statuses[op_id] = 'stop'
```
oliver-sanders, line 404: FYI I think this work should probably go into [graphql-ws](https://github.com/graphql-python/graphql-ws) itself, we should try to do that one day.

cylc/uiserver/resolvers.py
```diff
@@ -16,14 +16,16 @@
...
 from copy import deepcopy
 from subprocess import Popen, PIPE, DEVNULL
 from typing import (
     TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Tuple, Union
 )
-
+from cylc.flow.id import Tokens
 from graphql.language.base import print_ast
 
 from cylc.flow.data_store_mgr import WORKFLOW
```
oliver-sanders, line 31:
```suggestion
from copy import deepcopy
from subprocess import Popen, PIPE, DEVNULL
from typing import (
    TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Tuple, Union
)

from graphql.language.base import print_ast
import psutil

from cylc.flow.data_store_mgr import WORKFLOW
from cylc.flow.id import Tokens
```

cylc/uiserver/resolvers.py
```diff
@@ -57,6 +58,16 @@
...
+LOG_FILE_MAPPINGS = {
+    'job': 'j',
+    'job.out': 'o',
+    'job.err': 'e',
+    'job-activity.log': 'a',
+    'job.status': 's',
+    'job.xtrace': 'x'
+}
+
+
```
oliver-sanders, line 70: Can import this mapping from `cylc.flow.task_job_logs`, this will ensure the mapping is correct for the version of `cylc cat-log` being used.

Note it's back to front in cylc-flow but can be easily reversed:

```python
{
    value: key
    for key, value in x.items()
}
```

cylc/uiserver/resolvers.py
```diff
@@ -321,6 +332,70 @@ async def play(cls, workflows, args, workflows_mgr, log):
...
+        buffer: List[str] = []
+        queue: asyncio.Queue = asyncio.Queue()
+        # Farm out reading from stdout stream to a background task
+        # This is to get around problem where stream is not EOF until
+        # subprocess ends
+        asyncio.create_task(cls.enqueue(proc.stdout, queue))
+        op_id = info.root_value
+        if op_id not in info.context['sub_statuses']:
+            # subscription started in graphiql, add to dict
+            info.context['sub_statuses'][op_id] = 'start'
```
oliver-sanders, line 371: I think this gets done by the earlier code?

cylc/uiserver/resolvers.py
```diff
@@ -321,6 +332,70 @@ async def play(cls, workflows, args, workflows_mgr, log):
...
+                and info.context['sub_statuses'][op_id] == 'stop'
+            ):
+                with suppress(psutil.Error):
+                    cat_log_proc = psutil.Process(proc.pid)
+                    # cat-log process will terminate on tail termination
+                    for tail_proc in cat_log_proc.children():
+                        tail_proc.terminate()
+                # clean the subscription from the status dict
+                info.context['sub_statuses'].pop(op_id)
+                break
```
oliver-sanders, line 384: I think the `enqueue` coroutine will keep running unless explicitly cancelled, it won't shutdown automatically when it goes out of scope. It won't do anything much because it will be `await`ing lines from the `stream`, however, it will clog up the event loop as asyncio will still have to manage this coroutine (unless it bails with traceback because the `stream` isn't readable???).

We should explicitly stop it but cancelling async generators is a right pain with different behaviours at different Python versions annoyingly. I think `task.cancel()` [might do it](https://stackoverflow.com/a/60233813) e.g:

```python
enqueue_task = asyncio.create_task(cls.enqueue(proc.stdout, queue))
...
with suppress(...):  # suppress RuntimeError?
    enqueue_task.cancel()
```

cylc/uiserver/resolvers.py
```diff
@@ -397,10 +472,31 @@ async def service(
...
-        else:
+        elif command == 'play':
             return await Services.play(
                 workflows,
                 kwargs,
                 self.workflows_mgr,
                 log=self.log
             )
+
+        raise Exception()
```
oliver-sanders, line 483: Might be worth using `NotImplementedError` here as it's a little more descriptive than a blank exception and will get excluded from coverage.

```suggestion
        raise NotImplementedError()
```

cylc/uiserver/resolvers.py
```diff
@@ -397,10 +472,31 @@ async def service(
...
+    ):
+        if file:
+            file = LOG_FILE_MAPPINGS.get(file, file)
+        async for ret in Services.cat_log(
+            workflows[0],
+            self.log,
+            info,
+            task,
+            file
+        ):
```
oliver-sanders, line 501: Note: For 8.2.0 it might be worth adapting the interface of `cat-log` to allow it to accept full file names.

cylc/uiserver/schema.py
```diff
@@ -248,14 +249,66 @@ class Arguments:
...
+        command='cat_log',
+        workflow: str,
+        task=None,
+        file=None,
+        **kwargs: Any,
+    ) -> AsyncGenerator[Any, None]:
+        """Cat Log Resolver
+        Expands workflow provided subscription query.
+        """
+        parsed_workflows = [Tokens(workflow)]
```
oliver-sanders, line 271: We may need to handle the error `Tokens` returns for an invalid ID, not sure. Are errors automatically caught and returned if they happen in this resolver?

cylc/uiserver/handlers.py
oliver-sanders, line 404: Because we only add entries to this dictionary (and never remove them) this is a memory leak albeit a very small one.

Is there an opportunity later on in the code, perhaps when the operation is "unsubscribed" later on in the process, to remove this entry from the `dict` to avoid them accumulating?

oliver-sanders (commented):
Looking good.

I think you have filled in for a shortcoming in functionality in graphql-ws, great stuff working that out! We should try to contribute this enhancement back to them at some point.

cylc/uiserver/handlers.py
datamel: Since the only subscriptions we need to track in this dictionary are log data ones, and other subscriptions come in with a start (deltas for example) we need to check for the operation name to ensure we don't add to the dictionary anything that isn't a log data subscription. I think I will rename `sub_statuses`  to `log_sub_statuses` to make this clearer.
Note that if you start a Log subscription in graphiql,depending on how you write the subscription, it may not come in with the `LogData` operation name. In this case we add the op_id to the `sub_statuses` dictionary in `resolvers.py` code to ensure it doesn't get missed.

cylc/uiserver/resolvers.py
datamel: This is just to catch subscriptions added by graphiql interface which come in without the log data operation name.

cylc/uiserver/handlers.py
oliver-sanders: I think it's ok to make this generic rather than targetting it at the LogData operation manually (so long as we remove operations from the dict when appropriate).

(note that subscriptions which aren't called `LogData` will fail)

datamel:
Changes made in response to comments and from discussions over the phone. I think this one is ready for re-review @oliver-sanders.

cylc/uiserver/resolvers.py
```diff
@@ -379,14 +372,19 @@ async def cat_log(cls, workflow: Tokens, log, info, task=None, file=None):
...
                 break
-
             if queue.empty():
                 if buffer:
                     yield list(buffer)
                     buffer.clear()
+                if proc.returncode not in [None, 0]:
+                    (_, stderr) = await proc.communicate()
+                    # pass any error onto ui
+                    yield [stderr.decode()]
```
oliver-sanders, line 386: Should this be:

```python
yield stderr.decode().splitlines()
```

datamel: I'll take a look now.

cylc/uiserver/handlers.py
```diff
@@ -386,6 +390,14 @@ def open(self, *args, **kwargs):  # noqa: A003
         )
 
     async def on_message(self, message):
+        with suppress(ValueError, SyntaxError):
+            message_dict = ast.literal_eval(message)
```
oliver-sanders, line 394: We can use JSON here which should be lighter-weight as it supports fewer types:

```suggestion
            message_dict = json.loads(message)
```


(note will need to fiddle the imports)

cylc/uiserver/resolvers.py
```diff
@@ -321,6 +324,76 @@ async def play(cls, workflows, args, workflows_mgr, log):
         await workflows_mgr.scan()
         return response
 
+    @classmethod
+    async def enqueue(cls, stream, queue):
```
oliver-sanders, line 328: We can get away with a static method here:

```suggestion
    @staticmethod
    async def enqueue(stream, queue):
```

cylc/uiserver/resolvers.py
```diff
@@ -321,6 +324,76 @@ async def play(cls, workflows, args, workflows_mgr, log):
...
+        Used for log subscriptions.
+        """
+        full_workflow = workflow
+        if file and task:
+            full_workflow = Tokens(f"{workflow.workflow_id}//{task}")
+        cmd = ['cylc', 'cat-log']
+        if file:
+            cmd += ['-f', file]
+        else:
+            cmd += ['-m', 't']
```
oliver-sanders, line 345: I think we should always `tail` the logs?

```suggestion
        cmd += ['-m', 't']
```

cylc/uiserver/resolvers.py
```diff
@@ -321,6 +324,76 @@ async def play(cls, workflows, args, workflows_mgr, log):
...
+        op_id = info.root_value
+        if op_id not in info.context['sub_statuses']:
+            # subscription started in graphiql, add to dict
+            info.context['sub_statuses'][op_id] = 'start'
+        while True:
+            if (
+                op_id in info.context['sub_statuses']
+                and info.context['sub_statuses'][op_id] == 'stop'
+            ):
+                with suppress(psutil.Error):
```
oliver-sanders, line 370: I think this can be simplified to:

```suggestion
            if info.context['sub_statuses'][op_id] == 'stop':
                with suppress(psutil.Error):
```

cylc/uiserver/resolvers.py
```diff
@@ -321,6 +324,76 @@ async def play(cls, workflows, args, workflows_mgr, log):
...
+            if queue.empty():
+                if buffer:
+                    yield list(buffer)
+                    buffer.clear()
+                if proc.returncode not in [None, 0]:
+                    (_, stderr) = await proc.communicate()
+                    # pass any error onto ui
+                    yield [stderr.decode()]
+                    break
+                await asyncio.sleep(0.1)
```
oliver-sanders, line 388: This 0.1 second sleep was a placeholder value, I think we will want to increase this for production.

I found this in `man tail`:

```
       -s, --sleep-interval=N
              with  -f,  sleep  for  approximately N seconds (default 1.0) between
              iterations; with inotify and --pid=P, check process P at least  once
              every N seconds
```

Which suggests that new log lines will only come through at one second intervals so we might as well set this sleep to one second.

cylc/uiserver/resolvers.py
```diff
@@ -397,10 +470,34 @@ async def service(
...
+        _command: str,
+        workflows: List[Tokens],
+        task=None,
+        file=None
+    ):
+        if file:
+            # set the correct option for cat log from cylc flow
+            for key, value in JOB_LOG_OPTS.items():
+                if value == file:
+                    file = key
```
oliver-sanders, line 495: From experimentation it looks like you can pass the raw file name to `cylc cat-log` so I think we can get away without this mapping?

```
$ cylc cat-log generic/run2//20200111T0000Z/fool/NN/ -f job.out
```

cylc/uiserver/resolvers.py
```diff
@@ -321,6 +324,76 @@ async def play(cls, workflows, args, workflows_mgr, log):
...
+
+    @classmethod
+    async def cat_log(cls, workflow: Tokens, log, info, task=None, file=None):
+        """Calls `cat log`.
+
+        Used for log subscriptions.
+        """
+        full_workflow = workflow
+        if file and task:
+            full_workflow = Tokens(f"{workflow.workflow_id}//{task}")
```
oliver-sanders, line 340: `Tokens` can raise an exception for invalid `Tokens`, might need to catch this exception and return a more helpful error.

oliver-sanders (commented):
Looking good, need to do some testing...

cylc/uiserver/websockets/tornado.py
```diff
@@ -130,24 +131,36 @@ async def on_start(self, connection_context, op_id, params):
...
+                async for single_result in iterator:
+                    if not connection_context.has_operation(op_id):
+                        break
+                    await self.send_execution_result(
+                        connection_context, op_id, single_result)
+        except Exception as e:
+            await self.send_error(connection_context, op_id, e)
+        await self.send_message(connection_context, op_id, GQL_COMPLETE)
+        await connection_context.unsubscribe(op_id)
+        await self.on_operation_complete(connection_context, op_id)
```
datamel, line 152: This line is not being called, even if moved higher up in the code, or placed in a `finally` block. This results in the dictionary not getting cleaned correctly, which is not ideal.

cylc/uiserver/resolvers.py
```diff
@@ -358,41 +357,36 @@ async def cat_log(cls, workflow: Tokens, log, info, task=None, file=None):
...
+                        await asyncio.sleep(0)
+        finally:
+            with suppress(psutil.Error):
+                cat_log_proc = psutil.Process(proc.pid)
+                # cat-log process will auto-terminate on tail termination
+                for tail_proc in cat_log_proc.children():
+                    tail_proc.terminate()
+                enqueue_task.cancel()
+                with suppress(asyncio.CancelledError):
+                    await enqueue_task
```
oliver-sanders, line 389: At present, if `psutil.Error` gets raised then the `enqueue_task.cancel()` bit won't get run.

To fix, unindent the enqueue_task` bit and it won't get skipped.

```suggestion
            with suppress(psutil.Error):
                cat_log_proc = psutil.Process(proc.pid)
                # cat-log process will auto-terminate on tail termination
                for tail_proc in cat_log_proc.children():
                    tail_proc.terminate()
            enqueue_task.cancel()
            with suppress(asyncio.CancelledError):
                await enqueue_task
```

cylc/uiserver/schema.py
oliver-sanders: I just tried this out, the error gets sent through to the UI in a clear manner:

Screenshot from 2023-01-12 10-16-16

So all good here.

cylc/uiserver/resolvers.py
oliver-sanders: Turns out it already does, sorted!

cylc/uiserver/websockets/tornado.py
oliver-sanders: I can't replicate this issue. Will investigate if we can replicate it, otherwise safe to ignore for now.

cylc/uiserver/resolvers.py
```diff
@@ -321,6 +323,71 @@ async def play(cls, workflows, args, workflows_mgr, log):
...
+        buffer: List[str] = []
+        queue: asyncio.Queue = asyncio.Queue()
+        # Farm out reading from stdout stream to a background task
+        # This is to get around problem where stream is not EOF until
+        # subprocess ends
+        enqueue_task = asyncio.create_task(cls.enqueue(proc.stdout, queue))
+        op_id = info.root_value
+        try:
+            while (op_id in info.context['sub_statuses']
+                    and info.context['sub_statuses'][op_id] != 'stop'):
```
oliver-sanders, line 362: I don't think it's possible for `op_id` not to be in `sub_statuses`?

```suggestion
            while info.context['sub_statuses'][op_id] != 'stop':
```

Or if it is:

```suggestion
            while info.context['sub_statuses'].get(op_id) != 'stop':
```

cylc/uiserver/tests/conftest.py
```diff
@@ -20,8 +20,10 @@
 import logging
 from pathlib import Path
 from shutil import rmtree
+import shutil
```
oliver-sanders, line 23: `rmtree` is already imported:

```suggestion
from shutil import rmtree
```

cylc/uiserver/tests/conftest.py
```diff
@@ -386,3 +388,16 @@ def _caplog(caplog, uiserver, level=logging.INFO):
...
+def workflow_run_dir(request):
+    "Set up workflow run dir, with log_dirs"
+    flow_name = f'cylctb-uiserver-{str(uuid4())[:8]}'
+    run_dir = Path('~/cylc-run' / Path(flow_name)).expanduser().resolve()
+    run_dir.mkdir(parents=True, exist_ok=True)
+    log_dir = Path(run_dir / 'log' / 'scheduler')
+    log_dir.mkdir(parents=True, exist_ok=True)
+    yield flow_name, log_dir
+    if not request.session.testsfailed:
+        shutil.rmtree(run_dir)
```
oliver-sanders, line 403:
```suggestion
        rmtree(run_dir)
```

oliver-sanders (commented):
This is pretty much ready to roll :rocket:.

Just one comment which I unhelpfully put on the cylc-ui end of this PR about leaving room in the schema to add other fields to the log e.g. to better communicate "file not found" type errors or MIME type information.

https://github.com/cylc/cylc-ui/pull/1158#discussion_r1066846780

And the last couple of minor comments...

cylc/uiserver/schema.py
```diff
@@ -282,21 +284,28 @@ async def resolve_logs(
             task,
             file
         ):
-            yield item
+            obj = UISSubscriptions.Logs()
+            obj.lines = item
+            yield obj
```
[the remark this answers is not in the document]
datamel: Ah I'll try that

oliver-sanders (approved):
:+1:

Kudos on the subscription stopping logic, that looked hard to work out!

PR #406, opened 2023-01-12
cat-log: handle large files

### LINKED ISSUE #403: cat-log: handle large files
Some badly written tasks may produce extremely large log files.

These file may be too large to view in the UI so we will need to protect against this.

The easiest option will be for the `cat-log` service to count the number of lines it has received and truncate the file once it has surpassed that limit. It would then need to yield a couple of extra lines explaining that the file has been truncated due to length, kill the `cat-log` process and return.

It's possible but awkward to implement this in `cylc cat-log` itself so is probably best to do this here. We may want to consider extending this to a paginated solution long term.

**Pull requests welcome!**

oliver-sanders:
> Built on #397

Came up with an ultra simple solution which should be good enough to hold up for now.

* Truncate files longer than 5000 lines.
* Closes #403

**Requirements check-list**

### DIFF
--- a/.github/workflows/2_auto_publish_release.yml
+++ b/.github/workflows/2_auto_publish_release.yml
@@ -65,3 +65,10 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
         release-url: ${{ steps.create-release.outputs.html_url }}
+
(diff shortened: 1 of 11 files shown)


datamel:
Closed as cherry-picked onto https://github.com/cylc/cylc-uiserver/pull/397

PR #414, opened 2023-01-25
gui: fix browser open on macos

oliver-sanders:
Fixes a minor bug discovered on macos but likely impacting other systems.

Add the `file://` prefix in front of the file path to ensure the browser opens it as a file rather than a URL.

Note the behaviour of a file path without the file prefix is going to be OS/browser specific.

This is fine providing that `gui_file` is guaranteed to be an absolute path which will require a quick check.

**Requirements check-list**
- [x] No changelog (un released bug)

### DIFF
--- a/cylc/uiserver/scripts/gui.py
+++ b/cylc/uiserver/scripts/gui.py
@@ -65,7 +65,7 @@ def main(*argv):
             )
             update_html_file(gui_file, workflow_id)
             if '--no-browser' not in sys.argv:
-                webbrowser.open(gui_file, autoraise=True)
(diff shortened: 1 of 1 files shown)


datamel (approved):
Tested on met office systems, not on macos (leaving that to you @oliver-sanders). Test failure unrelated.

hjoliver (approved):
Looks legit to me.

oliver-sanders:
Merging with two approvals.

Note that master is currently targetting the 1.2.1 milestone. We'll only switch to the two-branch approach if we need to develop features/bugfixes in tandem.

Issue #416, opened 2023-01-31
Add Clean Logic for Info-Files


datamel:
Jupyter Server creates two files when starting up the uiserver:
- `jpserver-<pid>-open.html`
- `jpserver-<pid>.json`

These files are currently housed in `~/.local/share/jupyter/runtime` for files generated by `cylc hub` and in `~/.cylc/uiserver/info_files` for files generated by `cylc gui`.

We currently have no cleaning logic in place for these files since jupyter code on clean shutdown will remove these.

Once we have selected a file for reuse, we should add a check that the process is running (and cleaning/reselecting) before pressing ahead with using for opening the gui as it appears these files can get left behind more often than anticipated.

dwsutherland:
This is kind of related to:
https://github.com/cylc/cylc-uiserver/issues/392

datamel:
Update on this, after discussion in UK Team meeting, we should check the selected url is there before opening. It may not be there for two reasons:
- ui-server was opened on another host
- ui-server may no longer be up (e.g. crash, or killed with `kill -9`)

In either case, we can prompt user to clean files and try with another running uiserver until list is exhausted.
This issue should also change the selection of file from random.choice to selecting the newest uiserver file.

dwsutherland:
>  we should check the selected url is there before opening

For `cylc gui` or CLI via UIS? or both?
Do performance implications matter for the latter?

datamel:
> > we should check the selected url is there before opening
> 
> For `cylc gui` or CLI via UIS? or both? Do performance implications matter for the latter?
I think both.

The immediate concern is the `cylc gui` command, since this re-uses the file which could be for an unreachable server - this results in an error page at launch which will prevent the user from seeing the gui. Users may encounter this at present if they have had a server crash, so it is a bug which needs fixing for the next release. At present the solution to this is for them to manually delete the `~/.cylc/uiserver/info_files` folder or to start the gui with `--new` option. 

With regards to the CLI via UIS, I think again this will need cleaning, since rogue files from crashes or `kill -9`s will possibly result in sending of port and host info of servers that are not running?

This is complicated by not knowing the actual host where it was launched. This continues to be a problem of sorts. I did have a look at the Jupyter docs again to see if there was a way to solve this problem but I didn't see anything obvious that could help us here.

PR #431, opened 2023-03-01
Clean gui info files

datamel:
closes: https://github.com/cylc/cylc-uiserver/issues/416

This fixes a bug whereby, if an existing ui-server is killed with a `kill -9` or with a system crash, jupyter server info files (which serve as a contact file) can be left behind. This is problematic as our logic (added in https://github.com/cylc/cylc-uiserver/pull/370) can select the broken server for re-use which results in the gui opening at an error page, rather than the cylc gui.

**How to test**
For reviewers to test (on master to reproduce the bug)...
- clean your info files dir (~/.cylc/uiserver/info-files)
- open a uiserver with `cylc gui`
- look at info files dir, you should have two files (one .json and a html open file), take the pid from this filename
- kill the process (kill -9 <pid>)
- run another gui `cylc gui` (don't use the `--new` option)
- you should get an error page instead of the cylc gui

Test again on this branch and the bug should be fixed, with a prompt asking you if you wish to remove the files (either response should result in a gui being opened correctly)
 
Note that this branch also changes the logic - the html open file is now no longer written to in this process, so functional testing of various gui openings (e.g. with/without `--new` and `workflow_id`) would be prudent.

- [x] If this is a bug fix, PR should be raised against the relevant branch.


cylc/uiserver/scripts/gui.py
```diff
@@ -26,10 +26,11 @@
 from contextlib import suppress
 from glob import glob
 import os
-from pathlib import Path
-import random
 import re
+from requests.exceptions import RequestException
+import requests
```
datamel, line 31: `requests` is currently coming in to cylc environment as it is required by jupyterhub, Sphinx and metomi-rose.
Is this sufficient or should I be adding it as a dependency of cylc-uiserver for completeness? Not sure if this could cause conflicts with pinning for the other packages so I haven't added it for now. Reviewer thoughts?

datamel: As suggested at team meeting I have added this as a loose dependency with no pinning to ensure no conflict with other package requirements.

cylc/uiserver/scripts/gui.py
```diff
@@ -53,25 +54,78 @@ def main(*argv):
...
+    existing_guis.sort(key=os.path.getmtime, reverse=True)
+    for gui_file in existing_guis:
+        url = fish_url_from_file(gui_file)
+        if url and is_active_gui(url):
+            return url
+        check_remove_file(gui_file)
+    return None
+
+
+def fish_url_from_file(gui_file):
```
wxtim, line 90: `get_url_from_file` more comprehensible for non native speakers - fish as a verb is a bit wierd IMO.

cylc/uiserver/scripts/gui.py
```diff
@@ -53,25 +54,78 @@ def main(*argv):
...
+    try:
+        os.unlink(gui_file)
+        os.unlink(json_file)
+    except Exception:
+        pass
+
+
+def check_remove_file(gui_file) -> None:
+    """Ask user if they want to remove the file."""
+    print("The following file cannot be used to open the Cylc GUI:"
```
wxtim, line 120: IMO this should be amber to draw attention to the fact that the program wants input in the terminal - this isn't very common for the UIS.

wxtim (commented):
Needs a changelog entry. Otherwise ready to roll. :)

- [x] Read Change, looks sane.
- [x] Manually tested and found to fix bug.

wxtim (approved):

cylc/uiserver/tests/test_gui.py
```diff
@@ -13,59 +13,153 @@
...
+    """Testing functionally the gui selection and cleaning process"""
+    # set up file structure
+    info_files_dir = Path(tmp_path/'.cylc'/'uiserver'/'info_files')
+    info_files_dir.mkdir(parents=True, exist_ok=True)
+    for i in range(1, 5):
+        pid = randint(1000, 100000)
+        html_file = (info_files_dir / f"jpserver-{pid}-open.html")
+        # the json file is unused but created empty to ensure the html is the
+        # file used for selection
+        _json_file = (info_files_dir / f"jpserver-{pid}.json")
```
MetRonnie, line 112: This is not actually being created in the test
```suggestion
        json_file = (info_files_dir / f"jpserver-{pid}.json")
        json_file.touch()
```

MetRonnie (commented):
Tested. @wxtim do you want to have a look at my comments and approve if you're ok with them being applied?

cylc/uiserver/scripts/gui.py
```diff
@@ -53,25 +54,79 @@ def main(*argv):
...
+        if url and is_active_gui(url):
+            return url
+        check_remove_file(gui_file)
+    return None
+
+
+def get_url_from_file(gui_file):
+    with open(gui_file, "r") as f:
+        file_content = f.read()
+    url_extract_regex = re.compile('url=(.*?)\"')
```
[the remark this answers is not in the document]
wxtim: 🤣

cylc/uiserver/scripts/gui.py
```diff
@@ -53,25 +54,79 @@ def main(*argv):
...
+    match = url_extract_regex.search(file_content)
+    return match.group(1) if match else None
+
+
+def is_active_gui(url):
+    """Returns true if return code is 200 from server"""
+    try:
+        req = requests.get(url)
+        if req.status_code == 200:
+            return True
```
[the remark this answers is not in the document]
wxtim: Why not go the whole hog and 

```suggestion
def is_active_gui(url: str) -> bool:
    """Returns true if return code is 200 from server"""
    try:
        return requests.get(url).status_code == 200
```

MetRonnie (approved):

PR #436, opened 2023-04-03
setup: 1.3.0 branch

oliver-sanders:
* Pushed master:1.2.x
* Bumped the version to 1.3.0 on master (this PR).

PR #437, opened 2023-04-03
Install coreutils

oliver-sanders:
> Built on: #436

We are using `tail --follow=name` which isn't posix. This is ok because we state GNU `coreutils` as a dependency of Cylc.

For Mac OS / BSD the GNU coreutils need to be installed alongside the BSD variants as documented here: https://cylc.github.io/cylc-doc/stable/html/installation.html#installing-on-mac-os

This PR uses a modified version of the `brew install` script used in the cylc-flow functional tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,6 +24,25 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Brew Install
+        if: startsWith(matrix.os, 'macos')
+        run: |
+          # install system deps
+          brew update
+          brew install bash coreutils
+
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
MetRonnie, line 1: While you're here, do you want to add `if: startsWith(matrix.os, 'ubuntu')` to the type checking step? It's currently taking 2 mins to run on MacOS and shouldn't be needed as type checking should be platform independent

MetRonnie (approved):

.github/workflows/test.yml
oliver-sanders: That sounds sensible.

PR #457, opened 2023-05-26
Logs subscription: yield error msg if cat-log exits due to deleted file

MetRonnie:
Relies on https://github.com/cylc/cylc-flow/pull/5559

**Check List**

- [x] No dependency changes 
- [x] No tests included as small change
- [x] No changelog entry as small change
- [x] No docs needed
- [x] Master branch


wxtim (approved):
- [x] Read
- [x] Checked that it allows https://github.com/cylc/cylc-ui/pull/1306 to work correctly.

cylc/uiserver/resolvers.py
```diff
@@ -379,10 +379,14 @@ async def cat_log(cls, id_: Tokens, log, info, file=None):
...
-                        (_, stderr) = await proc.communicate()
-                        # pass any error onto ui
-                        yield {'error': stderr.decode()}
+                    if proc.returncode is not None:
+                        if proc.returncode:
+                            (_, stderr) = await proc.communicate()
+                            # pass any error onto ui
+                            yield {'error': stderr.decode()}
+                        else:
+                            # cat-log exited 0 meaning file was removed
```
MetRonnie, line 388: Btw something I found out with the functional tests for https://github.com/cylc/cylc-flow/pull/5559 is that on the Ubuntu (GH Actions) version of `tail`, it exits 1 when the file is deleted. Whereas on our version of `tail` it exits 0. Anyway, both cases are handled here.

oliver-sanders: Does tail exiting non-zero cause cat-log to exit non-zero?

MetRonnie: Oh, just checked and yes it does. `cat-log` exits with whatever `tail` exited with

oliver-sanders: Looks like the exit code is filesystem dependent for the file removed case, possibly due to the presence/absence of inotify.

cylc/uiserver/resolvers.py
```diff
@@ -379,10 +379,13 @@ async def cat_log(cls, id_: Tokens, log, info, file=None):
                     if buffer:
                         yield {'lines': list(buffer)}
                         buffer.clear()
-                    if proc.returncode not in [None, 0]:
+                    if proc.returncode is not None:
```
oliver-sanders, line 382: The best solution I can think of to capture the "file deleted" case is stderr scraping :vomiting_face:

I.E. change the `'error'` message if the string:

```
no files remaining
```

Is in stderr. Will check with BSD tail to see what string it uses.

MetRonnie: So what/when should an error message be yielded? Is there anything wrong with the current approach?

oliver-sanders: I think we were trying to make the "log file deleted" scenario more intuitive.

I think this would result in the message "no files remaining" appearing in the GUI (without the "cylc cat-log exited 0" bit) which " is reasonable, we could swap it out for "file removed" if we cared to scrape stderr which is yuck enough to be a valid reason not to.

oliver-sanders: (latest commit changes this output to something must more end-user readable)

oliver-sanders (approved):
Nice.

wxtim (approved):
🦛  to approve. 😄

PR #481, opened 2023-08-09
Use towncrier for changelog

MetRonnie:


oliver-sanders:
Is this necessary for cylc-uiserver?

MetRonnie:
The big advantage I see is we no longer have to worry about whether a changelog entry is in the right section after PRs get bumped around to different milestones

oliver-sanders:
I get the intention, I'm just not sure the problem warrants the solution here.

MetRonnie:
Any downside to using towncrier?

oliver-sanders:
Extra action/integration to maintain, no obvious advantage?

MetRonnie:
IMO never having to faff about with `CHANGES.md` is worth any minor maintenance overhead (no particular need to touch this config ever again).

.github/workflows/update_ui.yml
```diff
@@ -55,6 +55,14 @@ jobs:
             git commit -m "ui: install ${version}"
           done
 
+      - name: Add to changelog
```
wxtim, line 58: Nice.

wxtim (approved):
Looks good

MetRonnie:
Merging with 1 review on the grounds it's working successfully on cylc-flow and the fact I had to fix another duplicate heading in `CHANGES.md`

Issue #377, opened 2022-08-11
scan endpoint


oliver-sanders:
We have an internal interface for requesting a workflow scan:

https://github.com/cylc/cylc-uiserver/blob/c3f002f89b3dfeb3333b411842c928cb96293b95/cylc/uiserver/workflows_mgr.py#L413-L416

We should expose this, probably as a REST endpoint.

**Pull requests welcome!**

oliver-sanders:
Needed for https://github.com/cylc/cylc-ui/issues/1092

PR #493, opened 2023-09-05
Feature add scan mutation

markgrahamdawson:
Closes https://github.com/cylc/cylc-uiserver/issues/377

This was going to bee added as a rest endpoint but after discussion with @oliver-sanders it was decided it would be better to add another mutation to uiserver

**Check List**

- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present). - **NA**
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.- **NA**
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch. - **NA**


cylc/uiserver/resolvers.py
```diff
@@ -534,22 +543,27 @@ async def service(
         workflows: Iterable['Tokens'],
         kwargs: Dict[str, Any]
     ) -> List[Union[bool, str]]:
```
markgrahamdawson, line 545: refactored multiple if else into dictionary (because flake8 told me to)

cylc/uiserver/resolvers.py
```diff
@@ -534,22 +543,27 @@ async def service(
...
                 self.workflows_mgr,
-                log=self.log
-            )
+                log=self.log,
+            ),
+            'scan': Services.scan(
+                kwargs,
+                self.workflows_mgr,
+            ),
+        }
```
oliver-sanders, line 565: We can move this dictionary out of the function to avoid re-declaring it every time the function is run e.g:

```python
SERVICES = { ... }

async def service(...): ...
```

cylc/uiserver/resolvers.py
```diff
@@ -279,6 +279,15 @@ async def clean(
...
         return cls._return("Workflow(s) cleaned")
 
+    @classmethod
+    async def scan(
+        cls,
+        args: dict,
+        workflows_mgr: 'WorkflowsManager',
+    ):
+        await workflows_mgr.scan()
+        return cls._return("Workflow(s) scanned")
```
oliver-sanders, line 289: It's more like "scan requested" at this point.

oliver-sanders (commented):
LGTM

CHANGES.md
```diff
@@ -12,6 +12,10 @@ $ towncrier create <PR-number>.<break|feat|fix>.md --content "Short description"
 
 <!-- towncrier release notes start -->
 
+### Enhancements
+
+[#493](https://github.com/cylc/cylc-uiserver/pull/493) - Added mutation for scanning filesystem.
+
```
MetRonnie, line 18:
```suggestion
```
We have switched to a new way of generating the changelog (see the comment above). But actually this change does not need a changelog entry as no-one is going to use the mutation directly (not to say it isn't an important change though!)

cylc/uiserver/resolvers.py
```diff
@@ -529,31 +539,42 @@ async def mutator(
...
-        info: 'ResolveInfo',
-        command: str,
-        workflows: Iterable['Tokens'],
-        kwargs: Dict[str, Any]
-    ) -> List[Union[bool, str]]:
+    async def get_service(self, command, kwargs, workflows):
+        method = getattr(Services, command)
         if command == 'clean':
-            return await Services.clean(
+            mutation_response = partial(
```
MetRonnie, line 545: @oliver-sanders I don't think we need `functools.partial` for this kind of pattern?

MetRonnie: Just saw the comment above

> refactored multiple if else into dictionary (because flake8 told me to)

In this case I don't see a problem with multiple `else if`; I'm happy with a `# noqa: <err code>` comment being added to the offending line

cylc/uiserver/resolvers.py
markgrahamdawson: This logic has been moved out of the 'service' method and moved into a 'get_service' method.

Initially Flake8 was unhappy with multiple if-else statements and suggested using a dictionary instead, however I struggled to make that work. I also tried using getattr and partial functions to get around the multiple if else - also couldnt make that work. Have now reverted back to the if-else option (which flake8 isn't complaining about now).

getattr and partial are still being used but are not necessary. The get_service method could just be...

`
    async def get_service(self, command, kwargs, workflows):
        if command == 'clean':
            mutation_response = Services.clean(
                workflows,
                kwargs,
                self.workflows_mgr,
                log=self.log,
                executor=self.executor
            )
        elif command == 'play':
            mutation_response = Services.play(
                workflows,
                kwargs,
                self.workflows_mgr,
                log=self.log
            )
        elif command == 'scan':
            mutation_response = Services.scan(
                kwargs,
                self.workflows_mgr
            )
        return await mutation_response()
`

cylc/uiserver/resolvers.py
markgrahamdawson: More details here:
https://github.com/cylc/cylc-uiserver/pull/493#discussion_r1322843810
But yea - I think I probably shouldnt have touched what was there in the first place

cylc/uiserver/resolvers.py
```diff
@@ -529,31 +539,61 @@ async def mutator(
...
         self,
         info: 'ResolveInfo',
         command: str,
         workflows: Iterable['Tokens'],
-        kwargs: Dict[str, Any]
+        kwargs: Dict[str, Any],
     ) -> List[Union[bool, str]]:
-        if command == 'clean':
+
+        if command == 'clean': # noqa: <err code>
```
MetRonnie, line 577: I meant the specific error code that flake8 gave for this line

cylc/uiserver/resolvers.py
```diff
@@ -529,31 +539,61 @@ async def mutator(
...
-
         elif command == 'play':
             return await Services.play(
                 workflows,
                 kwargs,
                 self.workflows_mgr,
                 log=self.log
             )
-
-        raise NotImplementedError()
```
MetRonnie, line 556: This should be kept in order to noisily fail if we mess something up

cylc/uiserver/resolvers.py
```diff
@@ -529,31 +539,61 @@ async def mutator(
...
+                log=self.log
+            )
+        elif command == 'scan':
+            mutation_response = partial(
+                method,
+                kwargs,
+                self.workflows_mgr
+            )
+        return await mutation_response()
+
```
MetRonnie, line 568: Can remove this now the same thing is implemented below

cylc/uiserver/resolvers.py
```diff
@@ -534,24 +543,29 @@ async def service(
         info: 'ResolveInfo',
         command: str,
         workflows: Iterable['Tokens'],
-        kwargs: Dict[str, Any]
+        kwargs: Dict[str, Any],
     ) -> List[Union[bool, str]]:
-        if command == 'clean':
+
+        if command == 'clean': # noqa: <SIM116 Use a dictionary lookup instead of 3+ if/elif-statements>
```
MetRonnie, line 549: Just the error code is fine 🙂
```suggestion
        if command == 'clean': # noqa: SIM116
```
(Otherwise you'll hit the "line too long" error code!)

oliver-sanders (commented):
LGTM

MetRonnie (approved):

MetRonnie:
@oliver-sanders Is that an approval or are you still yet to test etc?

hjoliver (approved):
LGTM too 👍

PR #496, opened 2023-09-08
Allow setting user conf path through environment variables

### LINKED ISSUE #489: Configurable USER_CONF_ROOT path
### Problem

The `USER_CONF_ROOT` path is used in a variety of places for storing logs, secrets and the Jupyter database. While most uses are configurable in `jupyter_config.py`, the output path of `RotatingUISFileHandler` is not. We'd like to place this log in a different location with append-only permissions.

### Proposed Solution

* Allow `USER_CONF_ROOT` to be configured with an environment variable, e.g. `$CYLC_CONF_PATH` in the same way as [cylc-flow](https://cylc.github.io/cylc-doc/stable/html/reference/config/global.html#envvar-CYLC_CONF_PATH).

and/or

* Add a new configuration item e.g. `c.CylcUIServer.server_log_path` that sets an output path for `RotatingUISFileHandler` (defaulting to `USER_CONF_ROOT/log` as it currently does).

I can see a complication that `RotatingUISFileHandler` is initialised very early in the startup, potentially before the configuration is loaded.

ScottWales:
Teach the uiserver about two environment variables:

`$CYLC_CONF_PATH` (default `~/.cylc`) sets a path to read the uiserver configuration from, and also allows `USER_CONF_ROOT` in the server configuration to be updated by users. This variable matches the behaviour of `cylc-flow`.

`$CYLC_UISERVER_LOG_PATH` (default `~/.cylc/uiserver`) sets the output path for Cylc uiserver logs. These logs are created before the jupyter configuration is loaded so are unable to be configured there.

Closes #489 

**Check List**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,6 +52,7 @@ below.
  - David Matthews
  - Mel Hall
  - Christopher Bennett
+ - Scott Wales
  <!-- end-shortlog -->
 
 (All contributors are identifiable with email addresses in the git version
--- a/README.md
+++ b/README.md
@@ -141,7 +141,8 @@ The Cylc Hub will load the following files in order:
 
 3) User Config
 
-   This file
+   This file configures the Hub/UIS for the current user. The default path can
+   be changed by the ``CYLC_CONF_PATH`` environment variable.
 
    (`~/.cylc/uiserver/jupyter_config.py`)
 
@@ -174,7 +175,8 @@ are run use the `ServerApp.jpserver_extensions` configuration, see the
 [Jupyter Server configuration documentation](https://jupyter-server.readthedocs.io/en/latest/other/full-config.html#other-full-config).
 
 By default the Cylc part of the UI Server log is written to
-`~/.cylc/uiserver/uiserver.log`.
+`~/.cylc/uiserver/log/log`. This can be changed with the environment
+variable ``CYLC_UISERVER_LOG_PATH``.
 
 <!--
 TODO: Link to Jupyter Server logging_config docs when published
--- a/cylc/uiserver/config_util.py
+++ b/cylc/uiserver/config_util.py
@@ -38,7 +38,7 @@ SITE_CONF_ROOT: Path = Path(
     or GlobalConfig.DEFAULT_SITE_CONF_PATH,
     UISERVER_DIR
 )
-USER_CONF_ROOT = Path.home() / '.cylc' / UISERVER_DIR
+USER_CONF_ROOT = Path(os.getenv('CYLC_CONF_PATH') or Path.home() / '.cylc', UISERVER_DIR)
 
 
 def get_conf_dir_hierarchy(
--- a/cylc/uiserver/logging_util.py
+++ b/cylc/uiserver/logging_util.py
@@ -30,7 +30,7 @@ class RotatingUISFileHandler(logging.handlers.RotatingFileHandler):
     LOG_NAME_EXTENSION = "-uiserver.log"
 
     def __init__(self):
-        self.file_path = Path(USER_CONF_ROOT / "log").expanduser()
+        self.file_path = Path(os.getenv('CYLC_UISERVER_LOG_PATH') or USER_CONF_ROOT, "log").expanduser()
 
     def on_start(self):
         """Set up logging"""
--- a/cylc/uiserver/tests/test_config.py
+++ b/cylc/uiserver/tests/test_config.py
@@ -84,6 +84,20 @@ def test_cylc_site_conf_path(clear_env, capload, monkeypatch):
     ]
 
 
+def test_cylc_user_conf_path(clear_env, capload, monkeypatch):
+    """The user config should change to $CYLC_CONF_PATH if set."""
+    monkeypatch.setenv('CYLC_CONF_PATH', 'elephant')
+    monkeypatch.setattr(config_util, '__version__', '0')
+    load()
+    assert capload == [
+        SYS_CONF,
+        (SITE_CONF / 'jupyter_config.py'),
+        (SITE_CONF / '0/jupyter_config.py'),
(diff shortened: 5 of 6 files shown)


oliver-sanders (commented):
There are a couple of differences between this PR and the [cylc-flow use of `CYLC_CONF_PATH`](
https://cylc.github.io/cylc-doc/stable/html/reference/config/global.html):

1. In cylc-flow, `CYLC_CONF_PATH` overrides both the user and site configuration paths, not just the user one.
2. In cylc-flow, `CYLC_CONF_PATH` is flat, i.e. it's `$CYLC_CONF_PATH/global.cylc` not `$CYLC_CONF_PATH/flow/global.cylc`.

If the aim is to configure Cylc UI Server defaults for users, then you'll want to use the site-configuration to do this, (e.g. `c.CylcUIServer.logging_config` or something similar) `CYLC_CONF_PATH` won't help if it turns off the site-config (note the site-config path is already configurable via `CYLC_SITE_CONF_PATH`).

ScottWales:
Thanks @oliver-sanders, I didn't notice `CYLC_CONF_PATH` did a full override.

Rather than user configuration I'm wanting to relocate all the other files that the uiserver creates in the home directory under `USER_CONF_ROOT`, as that's going to be read-only in our production ui server. Some paths can be modified by the configuration but others cannot, for instance the [log file path](https://github.com/cylc/cylc-uiserver/blob/master/cylc/uiserver/logging_util.py#L33) and [INFO_FILES_DIR](https://github.com/cylc/cylc-uiserver/blob/master/cylc/uiserver/app.py#L109).

What do you think of the same functionality under a different environment variable name, say `CYLC_UISERVER_RUNTIME_DIR`?

ScottWales:
It's an imposed security constraint from our systems people, the web server VM running Cylc hub has /home mounted read-only to limit the impact of a compromised web server. Interactions with the Cylc server VM have to happen over the api.

hjoliver:
> , the web server VM running Cylc hub has /home mounted read-only

@ScottWales you mention the hub here, but from your description above, the reason for this PR is to relocate the Cylc UI Server logs? The UI Server does not have to (in fact, is not expected to) run where the (privileged) hub runs. 

The UIS needs to run (as the user) where it has access to user-owned scheduler DBs and log files, so that it can serve critical offline data.

Note also the UI Server log is not where workflow interventions are logged with user attribution (if other authenticated users have been authorized by the owner to do that) - that goes to the scheduler logs.  The UI Server log is  technical ase-level Jupyter server thing.

Does this change anything, or have I misunderstood your intent here?

ScottWales:
Ok, that's interesting - we have the uiserver being spawned on the same host as hub using `sudospawner-singleuser`. How is your configuration spawning uiserver?

hjoliver:
> Ok, that's interesting - we have the uiserver being spawned on the same host as hub using sudospawner-singleuser. How is your configuration spawning uiserver?

We have thus far not deployed Cylc 8 with Jupyterhub.  Same at MO I believe.  For the moment we have told users to run `cylc gui` at the terminal, which starts a UI Server and prints a token-protected URL for the user. "Single user mode" is good for everything except cross-user visibility and the authorization layer.

Hub deployment is coming soon but the team has still had higher-priority issues to knock off first. It's not a concern, Jupyterhub is proven tech. It is used to spawn notebook servers on other hosts all the time. There's are a bunch of hub spawners available, to support different ways of starting servers on other hosts. On the team, we have tested a few of these in the past. All that needs to be done, I think, is to see what spawners are available, choose one that is appropriate for your environment, and test it out. Worst case scenario we might need to modify an existing spawner or perhaps even make a custom one, but that should not be a big job.

oliver-sanders:
> It's an imposed security constraint from our systems people, the web server VM running Cylc hub has /home mounted read-only to limit the impact of a compromised web server. Interactions with the Cylc server VM have to happen over the api.

This makes sense for the Hub (i.e. jupyter hub) which should run under a special user account, but this does not make sense for the UI Server which runs as the user.

* The Hub
  * A Jupyter Hub instance.
  * Spawns servers on behalf of users.
  * Runs a reverse-proxy allowing users to connect to these workflows using a friendly URL.
  * Does NOT manage workflows, meddle with user accounts, etc, it just spawns servers.
  * Recommend running this under a special user account, and spawning user's servers using `sudo` ([guide](https://jupyterhub.readthedocs.io/en/stable/howto/configuration/config-sudo.html)). Otherwise can be run under `root`.
* The UI Server
  * A Jupyter Server instance with the Cylc extension loaded.
  * (Note Jupyter Lab is also a Jupyter Server extension like the Cylc UI Server so both can be run on the same server if desired).
  * Runs as the user.
  * May monitor/control/run that user's workflows.

The `cylc hub` command is a wrapper for the `jupyterhub` command, the wrapper loads [this configuration file](https://github.com/cylc/cylc-uiserver/blob/master/cylc/uiserver/jupyter_config.py) which helps with Cylc integration. You don't need to use the `cylc hub` command, you could use `jupyterhub` and configure it yourself, the wrapper is only there for convenience.

Jupyter Hub is not a Cylc product, so I'm not in a good place to advise with this, they do have a [very active forum](https://discourse.jupyter.org/) where you could ask about this sort of read-only configuration. Like most systems, Jupyter Hub does need to be able to create files (database and possibly log files, etc) so you won't be able to go completely read-only.

> We have thus far not deployed Cylc 8 with Jupyterhub. Same at MO I believe

We haven't deployed Jupyter Hub ourselves yet, but don't let that stop you, it's ready to go, we just haven't had the time yet. Note deploying Jupyter Hub + Cylc UI Server is no different to deploying Jupyter Hub + Jupyter Lab.

I would strongly recommend waiting for cylc-uiserver 1.4.0 which we will release later today with a spot of luck. This upgrades the Cylc UI Server to work with the latest versions of Jupyter components. Watch out for an announcement here: https://cylc.discourse.group/t/cylc-8-2-0-released/702/3

Issue #469, opened 2023-07-17
authorisation failures result in traceback


oliver-sanders:
API requests which fail authorisation result in traceback.

E.G. if user_a only has READ permissions, but they try to stop a workflow, the request will be denied, but traceback will be printed to stderr and will appear in the logs.

We should try to make this appear more nicely.

markgrahamdawson:
- After looking into this it was found that the source of the trace-back is the authorization middleware for graphQL.

- The trace-back is coming from the [graphql-python ](https://github.com/graphql-python) library which we dont have much control over. 

- Also worth noting that in graphene3 the implementation is a little different so any fix might not be necessary/ might not work when the libraries we are using get updated from graphene2.

1. Deciding to leave issue for the time being

oliver-sanders:
We should re-attempt after upgrading to graphql-core/graphene v3, currently blocked by #333

PR #508, opened 2023-09-18
used log_message in HttpError

markgrahamdawson:
Closes https://github.com/cylc/cylc-uiserver/issues/469

Previous error message
`tornado.web.HTTPError: HTTP 403: Forbidden`
New error message
`tornado.web.HTTPError: HTTP 403: Authorization failed for mdawson:requested to read.`

**Check List**

### DIFF
--- a/cylc/uiserver/authorise.py
+++ b/cylc/uiserver/authorise.py
@@ -523,7 +523,7 @@ class AuthorizationMiddleware:
                        f":requested to {op_name}.")
         if message:
             log_message = log_message + " " + message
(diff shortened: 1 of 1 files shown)


markgrahamdawson:
After looking into this it was found that the source of the trace-back is the authorization middleware for graphQL.
The trace-back is coming from the [graphql-python ](https://github.com/graphql-python) library which we dont have much control over. 
Also worth noting that in graphene3 the implementation is a little different so any fix might not be necessary/ might not work when the libraries we are using get updated from graphene2.
Deciding to leave issue for the time being

PR #509, opened 2023-09-21
Allow local logs for cat-log

### LINKED ISSUE #506: Allow configuration to tell cat-log to fetch logs from (local) disk rather than remote host
For jobs running via PBS, job logs are not available until the job has completed (yes, even if the job runs for 4 hours and generates GBs of logs). I don't know why this is the case, but it is the case. PBS spools log messages somewhere and then moves them into the correct locations when the job is done. As such, if remote log fetching is enabled, then pulling the job logs from the job host means additional SSH connections, network traffic and load on the login hosts for zero value. This is even more the case when those connections are left open by users who navigate to another tab or whatever.

It would be ideal if there was a configuration item that told cat-log to only serve files from local disk, if that is what the user wanted.

ScottWales:
Add a configuration item `c.CylcUIServer.force_remote_logs` defaulting to `True` (matching existing behaviour)

When `True`, `cat-log` will use the `--force-remote` option to always fetch from the remote Cylc server

When `False` logs will only be fetched if they are not present on the local server

Fixes #506 

**Check List**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,6 +59,7 @@ below.
  - Mel Hall
  - Christopher Bennett
  - Mark Dawson
+ - Scott Wales
  <!-- end-shortlog -->
 
 (All contributors are identifiable with email addresses in the git version
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -65,6 +65,7 @@ from pkg_resources import parse_version
 from tornado import ioloop
 from tornado.web import RedirectHandler
 from traitlets import (
+    Bool,
     Dict,
     Float,
     Int,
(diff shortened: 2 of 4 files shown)


oliver-sanders:
As mentioned in #506 there shouldn't be a need to make this configurable, someone needs to go through the cat-log code and make any required adjudgements to ensure that the command uses the local files when available and file listings remain correct. This isn't something I've had the time to do so far.

oliver-sanders:
> someone needs to go through the cat-log code and make any required adjudgements to ensure that the command uses the local files when available and file listings remain correct. This isn't something I've had the time to do so far.

I have made time for this and written up the conclusions on #506.

I have opened issues to address the caveats of removing `--force-remote` whilst ensuring that log files are always accessible (target 8.5.0). Once these are in, we should change the UIS default to remove the `--force-remote` option (https://github.com/cylc/cylc-uiserver/issues/667).

I'm going to close this PR as the new default behaviour should satisfy all. Please follow up if otherwise.

Note that with or without the `--force-remote` option, the `cylc cat-log` command will go to the remote server in many cases to list or retrieve logs. If there are any remaining concerns over this, please raise a cylc-flow issue as it will have to be addressed in the `cat-log` command itself.

PR #512, opened 2023-10-13
schema: add source workflows

oliver-sanders:
Add source workflows to the schema.

Unblocks https://github.com/cylc/cylc-ui/issues/548, though we will probably want to extend this to subscriptions for that.

**Check List**

### DIFF
--- a/cylc/uiserver/schema.py
+++ b/cylc/uiserver/schema.py
@@ -25,27 +25,36 @@ from typing import TYPE_CHECKING, Any, List, Optional
 import graphene
 from graphene.types.generic import GenericScalar
 
+from cylc.flow.data_store_mgr import (
+    DELTA_ADDED,
+)
 from cylc.flow.id import Tokens
 from cylc.flow.network.schema import (
     CyclePoint,
     GenericResponse,
     ID,
-    SortArgs,
-    Task,
     Mutations,
     Queries,
-    process_resolver_info,
     STRIP_NULL_DEFAULT,
+    SortArgs,
     Subscriptions,
+    Task,
+    Workflow,
     WorkflowID,
     _mut_field,
+    get_nodes_all,
+    get_workflows,
+    process_resolver_info,
     sstrip,
-    get_nodes_all
 )
 from cylc.uiserver.resolvers import (
     Resolvers,
     list_log_files,
     stream_log,
+)
+from cylc.uiserver.services.source_workflows import (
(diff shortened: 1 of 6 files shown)


dwsutherland:
Looks fine, tests passing.
Probably would need to create this on the `cylc-flow` end to add to subscriptions (which might make more sense, as field(s) in a data-store).

hjoliver:
Is that a formal review/approval @dwsutherland ?

oliver-sanders:
Getting this into subscriptions would make sense, making it easier for cylc-ui views to access this information.

Would we have to add these fields in cylc-flow or could we shim them into cylc-uiserver similar to how we extend the GraphQL schema?

dwsutherland:
> Would we have to add these fields in cylc-flow or could we shim them into cylc-uiserver similar to how we extend the GraphQL schema?

Well, if we add this to the `cylc-flow` schema then it will be available both ends, and we may not need to figure it out if the Scheduler already knows and it's a workflow field of the data-store.

oliver-sanders:
Because this is UIS specific functionality that the Scheduler should never know about it would be better to try and keep this logic centralised in cylc-uiserver. Is is possible to extend the protobuf schema in a similar way to how we extend the GraphQL schema?

dwsutherland:
Ah, so it's not about knowing the source of installed/running workflows, it's about knowing `cylc-src` or it's aliases...

Yeah, better to keep it at UIS then, and we shouldn't need to deal with protobuf .. can just extend the subscriptions at the UIS

oliver-sanders:
Ok, I'll look into doing that. Would it be possible to extend the store to hold this or would I need to add a parallel store?

oliver-sanders:
(may as well draft this until the fields are available in subscriptions)

oliver-sanders:
(rebased)

oliver-sanders:
> (may as well draft this until the fields are available in subscriptions)

I've now got a use case for this which doesn't "require" the data to be available via subscriptions - opening source workflow directories in Jupyter Lab.

Jupyter Lab appears only to work with relative file paths, not absolute ones, so I have added the path to the source workflow relative to the server's CWD.

dwsutherland (approved):
Nice:
image

image


Extending to subscriptions wouldn't be too hard, but like this I suppose you'd need to import more subscription machinery.. Which we may end up doing more and more, the way we're heading.

PR #514, opened 2023-10-25
GH Actions: use branch sync workflow

MetRonnie:
Also turn off codecov comments. I've requested the Codecov app be added to this repo so we get it as CI checks instead, but it might need @hjoliver to approve that


wxtim (approved):

MetRonnie:
Codecoc app integration has been added

PR #525, opened 2023-11-20
Fix `cylc play` service

### LINKED ISSUE #524: Unable to set CYLC_VERSION in web UI for Play
### Description

When I try to Play a workflow via the web UI, and try to set a CYLC_VERSION for it via the Play>Edit interface, I get an error and it does not work.

### Reproducible Example

1. Launch a workflow (stopped)
2. Do an edit play
image
3. Set a cylcVersion string appropriately and scroll down to click on Submit
image
4. See the error toast message appear
image

### Expected Behaviour

No error. Workflow should start.

MetRonnie:
- cylc version arg was not working
- could not play multiple workflows

Closes #524 

Should we backport this to 1.3.x?

**Check List**

- [x] Tests are included 
- [x] No docs needed

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -95,26 +95,6 @@ def snake_to_kebab(snake):
     raise TypeError(type(snake))
 
 
-def check_cylc_version(version):
-    """Check the provided Cylc version is available on the CLI.
-
-    Sets CYLC_VERSION=version and tests the result of cylc --version
-    to make sure the requested version is installed and selectable via
-    the CYLC_VERSION environment variable.
-    """
-    proc = Popen(
-        ['cylc', '--version'],
-        env={**os.environ, 'CYLC_VERSION': version},
-        stdin=DEVNULL,
-        stdout=PIPE,
-        stderr=PIPE,
-        text=True
-    )
-    ret = proc.wait(timeout=5)
-    out, err = proc.communicate()
-    return ret or out.strip() == version
-
-
 def _build_cmd(cmd: List, args: Dict) -> List:
     """Add args to command.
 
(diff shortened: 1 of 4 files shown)


cylc/uiserver/resolvers.py
```diff
@@ -346,22 +319,16 @@ async def play(cls, workflows, args, workflows_mgr, log):
...
 
             except Exception as exc:
                 # oh noes, something went wrong, send back confirmation
                 return cls._error(exc)
 
-            else:
-                # send a success message
-                return cls._return(
-                    'Workflow started'
-                )
```
oliver-sanders, line 361: Looks like the original intention was to return a list with the status of each workflow which makes sense.

https://github.com/cylc/cylc-uiserver/commit/2f3a80a8eaafe4cb981332d8894c51391e901b5a#diff-6cea667f14824a3338879a9b66beae2f50139ec3d0ba7e0670110ba9debf13d7R173-R190

MetRonnie: The UI only accepts a response of the format `[bool, message]` where `bool` indicates the success/failure or just `message` which implies success

oliver-sanders: So something like this:

```
[
    False,
    [
        'abc: started',
        'bcd: started',
        'cde: error, workflow was previously running with Cylc 8.0.1, use --upgrade to restart',
    ]
]
```

MetRonnie: Done

cylc/uiserver/resolvers.py
```diff
@@ -289,79 +269,75 @@ async def scan(
...
+        if failed:
+            return cls._error(
+                "\n".join(f"{wflow}: {msg}" for wflow, msg in results.items())
+            )
```
oliver-sanders, line 335:
```suggestion
        if failed:
            return cls._error(
                "\n\n" + "\n\n".join(
                    f"{wflow}: {msg}" for wflow, msg in results.items()
                )
            )
```

Results in:

Screenshot from 2023-11-22 11-08-17

* The first `\n\n` is to separate the "Command failed: play -` message from the first result.
* The second `\n\n` is to put a blank line between each result as the stderr can be multiline.

oliver-sanders (approved):
:+1:

wxtim (approved):

wxtim:
Will merge after doing a little due diligence on failing test.

Issue #547, opened 2024-01-04
efficiency: investigate bottleneck


oliver-sanders:
This workflow has proven to be remarkably difficult for the UIS & UI to handle:

```ini
#!Jinja2

{% set members = 10 %}
{% set hours = 100 %}

[scheduler]
    allow implicit tasks = True

[task parameters]
    member = 0..{{members}}
    fcsthr = 0..{{hours}}
  [[templates]]
    member = member%(member)03d
    fcsthr = _fcsthr%(fcsthr)03d

[scheduling]
  initial cycle point = 2000
  runahead limit = P3
  [[xtriggers]]
    start = wall_clock(offset=PT7H15M)
  [[graph]]
    T00,T06,T12,T18 = """
        @start & prune[-PT6H]:finish => prune & purge
        @start => sniffer:ready_<member,fcsthr> => <member,fcsthr>_process? => finish
        <member,fcsthr>_process:fail? => fault
      """

[runtime]
    [[sniffer]]
        [[[outputs]]]
{% for member in range(0, members + 1) %}
    {% for hour in range(0, hours + 1) %}
            ready_member{{ member | pad(3, 0) }}_fcsthr{{ hour | pad(3, 0) }} = {{ member }}{{ hour }}
    {% endfor %}
{% endfor %}
```

For more information see: https://cylc.discourse.group/t/slow-load-of-cylc-workflows-disconnects/823/19

Investigation so far has confirmed:
* The scheduler is not a source of delay.
* The UIS chokes on the update for several seconds.
* The UI chokes on the deltas of several seconds.
* The browser takes a couple of seconds to update.

This issue focuses on the UIS side of things.

oliver-sanders:
Early investigations using https://github.com/cylc/cylc-uiserver/pull/546:

* As expected, there is no single source of the problem :(
* Resolvers
  * ~10% of CPU time for my example was used by [`cylc.flow.network.graphql.IgnoreFieldMiddleware.resolve`](https://github.com/cylc/cylc-flow/blob/master/cylc/flow/network/graphql.py#L342) and it's callchain.
  * This is the cost of actually computing the response, but it's only ~10% of the CPU time?!
  * `iscoroutinefunction` called here took a reasonable amount of time, it's possible that upgrading our dependencies may remove the requirement to check for synchronous functions?
* Asyncio
  * Asycio is processing a very large number of tasks and taking quite a while doing it.
[...]

Issue #333, opened 2022-03-23
dependency sustainability


oliver-sanders:
Cylc UI Server has two dependencies from the graphql-python org that are starting to look worrying:

* graphene-tornado
  * [![GitHub last commit](https://img.shields.io/github/last-commit/graphql-python/graphene-tornado)](https://github.com/graphql-python/graphene-tornado)
  * Adds tornado support to Graphene.
  * Bruno maintains the conda-forge feedstock.
  * A v3 beta was started but never released.
  * One bug we have spotted may be related - https://github.com/cylc/cylc-uiserver/issues/328
* graphql-ws
  * [![GitHub last commit](https://img.shields.io/github/last-commit/graphql-python/graphql-ws) ](https://github.com/graphql-python/graphql-ws)
  * Adds websocket support to GraphQL).
  * Hillary & Bruno maintain the conda-forge feedstock.
  * The project was forked for Graphene 3, but then abandoned.

These projects are fairly small and live in the graphql-python organisation, however, are in need of a little TLC. They are currently stuck on Graphene 2, Graphene has since moved onto version 3 (Nov 2021).

Unfortunately it is relatively unusual to write GraphQL *servers* in Python (clients less unusual), GraphQL came out the JS ecosystem where the most popular server lives. Consequently there isn't that much interest from the wider community in some of these more niche integrations in the Python ecosystem.

To safeguard the future of these projects they need to make it to Graphene 3, unfortunately this requires an understanding of Graphene that we don't have which makes it difficult for us to help.

What can/should we do and when can we afford the time to do it?

oliver-sanders:
If the Graphene route becomes difficult there are two other Python GraphQL server implementations which may or may not offer an easier route:

* Ariadne
* Tartiflette

oliver-sanders:
update: 2023

graphene-tornado is now graphene >=3 compatible, but pending release, pending a maintainer to make the release.

Because graphql-ws is still graphene < 3 this doesn't help us so #388 is blocked with or without this release.

oliver-sanders:
Moving to Ariadne or Tartiflette is completely possible, however, this would involve swapping our GraphQL "engine" for theirs which could introduce migration pain.

Both Ariadne and Tartiflette are schema-first which means we would need to serialise our Graphene schema, then load it back in again. This should work but it might require a little extra work hooking in our resolvers, etc.

Furthermore, although both support websockets, neither support Tornado so we would still require an integration there bringing us back to the original problem.

[...]

PR #548, opened 2024-01-10
graphql-ws: optimise the "resolve" routine

oliver-sanders:
* Partially address
* The resolve routine is implmented recursively in the graphql-ws library.
* Because the function is async this results in a large number of async tasks being created when the library is used with large, deeply nested schema.
* Async tasks have an overhead, above that of regular function calls.
* For the example in https://github.com/cylc/cylc-uiserver/issues/547 this resulted in over 10 seconds of overheads.

Results **96% saving!**:
* Before (`base_async.resolve`):
  * cumtime: 11.24s
  * calls: 268'053
* After: (`resolve.resolve`):
  * cumtime - 0.3965s
  * calls: 24

Suggest adding this routine to the cylc-uiserver library for now as the graphql-ws library is not currently maintained (see #333).

The original implementation can be found here: https://github.com/graphql-python/graphql-ws/blob/7ef25ecfaf5390bf1a0cb4023272d8cb074368c2/graphql_ws/base_async.py#L37-L62

TODO:
* [x] Reviewers, please replicate profiling results.
* [x] Changelog entry.
* [x] License legal stuff.

**Check List**

### DIFF
--- a/cylc/uiserver/websockets/resolve.py
+++ b/cylc/uiserver/websockets/resolve.py
@@ -27,7 +27,8 @@ found in the graphql-ws library with the above license.
 This is temporary code until the change makes its way upstream.
 """
 
-# NOTE: transient dependency from graphql-ws not
+# NOTE: transient dependency from graphql-ws purposefully not
+# reflected in cylc-uiserver dependencies 
 from promise import Promise
 
 from graphql_ws.base_async import is_awaitable
@@ -39,7 +40,7 @@ async def resolve(
     _key=None,
 ):
     """
-    Recursively wait on any awaitable children of a data element and resolve
+    Wait on any awaitable children of a data element and resolve
     any Promises.
     """
     stack = [(data, _container, _key)]
--- a/cylc/uiserver/websockets/resolve.py
+++ b/cylc/uiserver/websockets/resolve.py
@@ -28,7 +28,7 @@ This is temporary code until the change makes its way upstream.
 """
 
 # NOTE: transient dependency from graphql-ws purposefully not
-# reflected in cylc-uiserver dependencies 
+# reflected in cylc-uiserver dependencies
 from promise import Promise
 
 from graphql_ws.base_async import is_awaitable
--- a/cylc/uiserver/websockets/resolve.py
+++ b/cylc/uiserver/websockets/resolve.py
@@ -1,0 +1,66 @@
+# MIT License
+#
+# Copyright (c) 2017, Syrus Akbary
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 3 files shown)


cylc/uiserver/websockets/resolve.py
```diff
@@ -0,0 +1,66 @@
...
+# SOFTWARE.
+
+"""
+This file contains an implementation of "resolve" derived from the one
+found in the graphql-ws library with the above license.
+
+This is temporary code until the change makes its way upstream.
+"""
+
+# NOTE: transient dependency from graphql-ws not
```
wxtim, line 30: Not what?

cylc/uiserver/websockets/resolve.py
```diff
@@ -0,0 +1,66 @@
...
+from graphql_ws.base_async import is_awaitable
+
+
+async def resolve(
+    data,
+    _container=None,
+    _key=None,
+):
+    """
+    Recursively wait on any awaitable children of a data element and resolve
```
wxtim, line 42: I thought that de-recursivizing was the whole point (tm)

wxtim:
IMO Adding it and modifying it : Would be nice if these were separate commits to keep that info in the repo history?

wxtim:
foo2

cylc/uiserver/websockets/resolve.py
```diff
@@ -0,0 +1,66 @@
...
+            items = _data.items()
+        elif isinstance(_data, list):
+            items = enumerate(_data)
+        else:
+            items = None
+        if items is not None:
+            stack.extend([
+                (child, _data, key)
+                for key, child in items
+            ])
```
wxtim, line 66: - Spaces make it clear that there are three different if loops.
- English notes.
- `items` => `children`

```suggestion
        # Await actual data if _data is awaitable:
        if is_awaitable(_data):
            _data = await _data
            if isinstance(_data, Promise):
                _data = _data.value
            if _container is not None:
                _container[_key] = _data

        # Populate child tasks from data:
        if isinstance(_data, dict):
            children = _data.items()
        elif isinstance(_data, list):
            children = enumerate(_data)
        else:
            children = None

        # If there are child tasks, put them into the stack:
        if items is not None:
            stack.extend([
                (child, _data, key)
                for key, child in items
            ])
```

wxtim (commented):
- [x] Read code. I have views on readability, but I'm happy that the change replicates the functioning of the original. Might be nice to unit test it?
- [x] Tried it out. Looks like there is a considerable speed gain for larger workflows.

Not pressed the approve button because it's still draft, presumably indicating futher work to come?

Might we unit test this function?

cylc/uiserver/websockets/resolve.py
oliver-sanders: Agreed, but, defo out of scope, but moreover this kinda isn't our code...

This file is a temporary measure pending the merge of a PR in an upstream repo which is held up by #333, so we should try to keep changes here minimal (and push changes to the upstream repo when it's ready).

cylc/uiserver/websockets/resolve.py
oliver-sanders:
```suggestion
# NOTE: transient dependency from graphql-ws purposefully not
# reflected in cylc-uiserver dependencies 
```

cylc/uiserver/websockets/resolve.py
oliver-sanders:
```suggestion
    Wait on any awaitable children of a data element and resolve
```

cylc/uiserver/websockets/resolve.py
```diff
@@ -0,0 +1,67 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""
+This file contains an implementation of "resolve" derived from the one
+found in the graphql-ws library with the above license.
+
+This is temporary code until the change makes its way upstream.
+"""
```
oliver-sanders, line 28: @dpmatthews, can you check you're happy with this.

cylc/uiserver/websockets/resolve.py
```diff
@@ -0,0 +1,67 @@
...
+
+# NOTE: transient dependency from graphql-ws purposefully not
+# reflected in cylc-uiserver dependencies
```
wxtim, line 31:
```suggestion
# reflected in cylc-uiserver dependencies
```

wxtim (approved):
Took the liberty of removing the trailing whitespace causeing the style check to fail.

Do you want to appoint a second review?

oliver-sanders:
The actual diff is trivial and we're choked up so might just slide this through on 1.

PR #558, opened 2024-02-16
authorisation: allow jupyter lab to work in standalone mode

### LINKED ISSUE #556: jupyterlab: enable access in standalone mode
Jupyter Lab is currently blocked in standalone (i.e. token auth) mode.

Context: In standalone mode, Jupyter server authorisation is open by default, i.e. the bearer of the token has full permissions. This makes some sense, but doesn't play nicely for multi-user access to a single server when running behind Jupyter Hub (because it grants full perms to any authenticated user).

So to protect against this, we define a Cylc "authorizer" which reduces the default authorzation to the server's owner (as opposed to any authenticated user).

https://github.com/cylc/cylc-uiserver/blob/c821ecc977fd6b4e439546dc75a3864c8ef6cdd4/cylc/uiserver/jupyter_config.py#L99-L103

This works fine for Jupyter Hub use cases. Unfortunately, this authorizer is also configured for standalone use cases which it subsequently breaks due to the way token auth works.

Suggest modifying the authorizer to give full perms if the user is token authenticated.

oliver-sanders:
Closes #556 

Jupyter Lab authorisation is open by default, a curious choice, safe for regular usage, but it becomes a security hole when deploying via Jupyter Hub and allowing users to see each other's servers.

Since this is key to Cylc Hub deployment, and we don't want our users accidentally opening security holes (Cylc must have safe, sane and secure defaults), we override the default security policy to only authorizing the server owner to access Jupyter Lab.

Unfortunately, I forgot to add a necessary backdoor to this policy for token authenticated sessions (i.e. sessions that are not run through Jupyter Hub) where the bearer of the token has full permission over the server.

This means that at present, Jupyter Lab is blocked by default when the Cylc UI Server is launched as a token authenticated (i.e. standalone) server. This PR reinstates access for this situation and this situation only.

* [x] Requires through security testing both in standalone mode and multi-user mode.

**Check List**

- [ ] Tests are included (manual security testing required).

### DIFF
--- a/cylc/uiserver/utils.py
+++ b/cylc/uiserver/utils.py
@@ -14,6 +14,35 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
+from typing import TYPE_CHECKING
+
+from jupyter_server.auth.identity import PasswordIdentityProvider
+
+if TYPE_CHECKING:
+    from cylc.uiserver.handlers import CylcAppHandler
+    from jupyter_server.auth.identity import (
+        IdentityProvider as JPSIdentityProvider,
+    )
+
+
+def is_bearer_token_authenticated(handler: 'CylcAppHandler') -> bool:
+    """Returns True if this request is bearer token authenticated.
+
+    Bearer tokens, e.g. tokens (?token=1234) and passwords, are short pieces of
+    text that are used for authentication. These can be used in single-user
+    mode (i.e. "cylc gui"). In these cases the bearer of the token is awarded
+    full privileges.
+
+    In multi-user mode, we have more advanced authentication based on an
+    external service which allows us to implement fine-grained authorisation.
+    """
+    identity_provider: 'JPSIdentityProvider' = (
+        handler.serverapp.identity_provider  # type: ignore[union-attr]
+    )
+    return identity_provider.__class__ == PasswordIdentityProvider
+    # NOTE: not using isinstance to narrow this down to just the one class
+
+
 def _repr(value):
     if isinstance(value, dict):
         return '<dict>'
--- a/changes.d/558.fix.md
+++ b/changes.d/558.fix.md
@@ -1,0 +1,1 @@
+Permit Jupyter Lab to be run in the same Jupyter Server instance as the Cylc UI Server in standalone mode (i.e. via `cylc gui`), note it was already possible to do this in multi-user mode (i.e. via `cylc hub`).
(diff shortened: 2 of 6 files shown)


cylc/uiserver/utils.py
```diff
@@ -14,6 +14,31 @@
...
+    """Returns True if this request is bearer token authenticated.
+
+    E.G. The default single-user token-based authenticated.
+
+    In these cases the bearer of the token is awarded full privileges.
+    """
+    identity_provider: 'JPSIdentityProvider' = (
+        handler.serverapp.identity_provider  # type: ignore[union-attr]
+    )
+    return identity_provider.__class__ == PasswordIdentityProvider
```
MetRonnie, line 38: Why is it `True` if using `PasswordIdentityProvider`? I'm not familiar with the jupyter_server API but `PasswordIdentityProvider` suggests not using a token.

Also I have spotted `jupyter_server.auth.identity.IdentityProver` has its own `is_token_authenticated` method, is there a reason we are not using this? - https://github.com/jupyter-server/jupyter_server/blob/da948f050455a58d7afe07665ac2d777ee5bb2ff/jupyter_server/auth/identity.py#L489

oliver-sanders: Passwords and tokens are both equivalent i.e. there is bearer token (or password) which grants full access to the server.

I'll look into the `is_token_authenticated` method. Because of the strange way Jupyter Hub wraps Jupyter Server I'm not sure whether it's safe to rely on it or not.

oliver-sanders: @MetRonnie 

tldr; No, the jupyter_server `is_token_authenticated` method does not return the right results for this usage.

```
>>> PasswordIdentityProvider().is_token_authenticated(SimpleNamespace(current_user='me'))
False
```

This cylc-uiserver implementation is focused on determining whether the session is "bearer token" (more general than just tokens per-se) authenticated (passwords are just another form of bearer token) or whether authentication is being provided by an external service.

If authentication is bearer token based, then the bearer of the token should have full permissions to the server, that is to say, fine grained permissions do not apply as there is only one level of user. The safest way to implement this (that I could come up with) is to whitelist only the one authentication provider class that we want this rule to apply to.

Long term, it may be worth raising this with Jupyter Server as I think multi-user server access may be off their radar.

MetRonnie: Perhaps rename our method to `is_bearer_token_authenticated` or `is_bearer_authenticated` to avoid confusion?

oliver-sanders: Changed and added more info into the docstring.

MetRonnie (commented):
LGTM, will approve once we have tested this

markgrahamdawson (approved):

MetRonnie (approved):

PR #574, opened 2024-03-28
data store: remove threading

### LINKED ISSUE #194: Revisit use of multiple subscription sockets/threads
> One socket may have many outgoing and many incoming connections.

Found here:
https://zguide.zeromq.org/docs/chapter2/#Plugging-Sockets-into-the-Topology

Can we reduce the need for multi-threaded UIS by using one subscription socket with multiple connections? (to every publisher/Scheduler)

oliver-sanders:
Closes #194

Run subscribers via asyncio rather than the ThreadPoolExecutor.

I think the only non-blocking operations in the code being called were:
- time.sleep  (has an async variant)
- asyncio.sleep (async)
- self.socket.recv_multipart (async)

If so, I don't think we need to use threading here so I've refactored the code so that the underlying async functions could be called via asyncio.

If so, this cuts out the need for the `ThreadPoolExecutor` removing the workflow limit.

**Check List**

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -32,11 +32,11 @@ Subscriptions are currently run in a different thread (via ThreadPoolExecutor).
 """
 
 import asyncio
-from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
+from functools import wraps
 from pathlib import Path
 import time
-from typing import Dict, Optional, Set
+from typing import Dict, Optional, Set, NamedTuple
 
 from cylc.flow.exceptions import WorkflowStopped
 from cylc.flow.id import Tokens
@@ -72,6 +72,34 @@ def log_call(fcn):
     return _inner
 
 
+def call_to_tuple(fcn):
+    """Turns function calls into an (args, kwargs) tuple.
+
+    Examples:
+        >>> xargs(list)(1, 2, a=3, b=4)
+        [(1, 2), {'a': 3, 'b': 4}]
+
+    """
+    @wraps(fcn)
+    def _inner(*args, **kwargs):
+        nonlocal fcn
+        return fcn((args, kwargs))
+
+    return _inner
+
+
+class Subscriber(NamedTuple):
+    """Represents an active subscription.
+
+    Args:
+        subscriber: The subscriber client object.
+        task: The asyncio task running the client.
+
+    """
+    subscriber: WorkflowSubscriber
+    task: asyncio.Task
+
+
 class DataStoreMgr:
     """Manage the local data-store acquisition/updates for all workflows.
 
@@ -80,14 +108,6 @@ class DataStoreMgr:
             Service that scans for workflows.
         log:
             Application logger.
-        max_threads:
-            Max number of threads to use for subscriptions.
-
-            Note, this determines the maximum number of active workflows that
-            can be updated.
-
-            This should be overridden for real use in the UIS app. The
-            default is here for test purposes.
 
     """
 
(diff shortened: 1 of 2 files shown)


cylc/uiserver/data_store_mgr.py
```diff
@@ -351,7 +407,6 @@ def _reconcile_update(self, topic, delta, w_id):
                         'pb_data_elements',
```
oliver-sanders, line 407: Do we need to use run_coroutine_threadsafe` any more (line 404)?

hjoliver:
Nice, hope we can get this working 🚀

oliver-sanders:
Ping @dwsutherland, I may well have overlooked something, does this make sense to you?

(note, tests are unhappy for async reasons)

cylc/uiserver/data_store_mgr.py
```diff
@@ -96,15 +116,34 @@ class DataStoreMgr:
...
+    def shutdown(self):
+        """Stop the data store manager.
+
+        This will stop any active subscribers.
+
+        Note: It will not wait for pending messages to be processed.
+        """
+        for w_id in self.subscribers:
+            self._stop_subscription(w_id)
+            self.message_processor_task.cancel()
```
oliver-sanders, line 146:
```suggestion
        self.message_processor_task.cancel()
```

dwsutherland:
> Ping @dwsutherland, I may well have overlooked something, does this make sense to you?
> 
> (note, tests are unhappy for async reasons, but it _should_ work for real usage)

Will have a look, I think I wanted to do this when I was first building, but didn't manage to figure out how to have non-blocking subscription receive loops or something

cylc/uiserver/data_store_mgr.py
dwsutherland: No, I would think not.

dwsutherland:
The changes look quite simple .. had a play, and I did notice something I couldn't reproduce on master.
When stopping a workflow in the WUI (latest master `cylc-ui`):
image
```
.
.
.
[I 2024-04-09T21:37:34.955 JupyterHub log:191] 200 GET /hub/error/503?url=%2Fuser%2Fsutherlander%2Fcylc%2Fsubscriptions (@127.0.0.1) 1.33ms
21:37:37.405 [ConfigProxy] error: 503 GET /user/sutherlander/cylc/subscriptions connect ECONNREFUSED 127.0.0.1:33077
[I 2024-04-09T21:37:37.407 JupyterHub log:191] 200 GET /hub/error/503?url=%2Fuser%2Fsutherlander%2Fcylc%2Fsubscriptions (@127.0.0.1) 1.35ms
[W 2024-04-09T21:37:39.146 JupyterHub base:1154] User sutherlander server stopped, with exit code: 0
[I 2024-04-09T21:37:39.146 JupyterHub proxy:356] Removing user sutherlander from proxy (/user/sutherlander/)
```
Not sure why.. could be something unrelated? maybe this branch needs rebased..

dwsutherland:
Ah, think I might know why, maybe .. the subscriber resolver (`cylc-flow`) might use that `w_subs` variable

cylc/uiserver/data_store_mgr.py
```diff
@@ -152,37 +191,32 @@ async def connect_workflow(self, w_id, contact_data):
...
+                contact_data[CFF.HOST],
+                contact_data[CFF.PUBLISH_PORT],
+            )
+            # make a one-off request to provide the initial data
+            await self._entire_workflow_update(w_id)
+        except WorkflowStopped:
+            self.disconnect_workflow(w_id)
+        except Exception as exc:
+            self.log.error(f'Failed to connect to {w_id}: {exc}')
             self.disconnect_workflow(w_id)
```
oliver-sanders, line 214: Moved error handling up a level from the `_start_subscription` and `_entire_workflow_update` methods, no more threads to interfere with error reporting.

cylc/uiserver/data_store_mgr.py
oliver-sanders: Added comment but haven't removed right now as this would require a little more async wrangling.

cylc/uiserver/data_store_mgr.py
```diff
@@ -72,6 +72,34 @@ def _inner(*args, **kwargs):  # works for serial & async calls
     return _inner
 
 
+def _call_to_tuple(fcn):
```
wxtim, line 75: @oliver-sanders : I've read the docs and think I might need you to explain this in person.

What are the limitations on the sort of functions I can use this on?

It certainly doesn't work on 

```python
def multiply(a, b):
    return a * b

def cube(a):
    return a ^ 3
```

If it's meant to be like this can we have some description in the docstring of what we can and can't input?

wxtim (approved):

dwsutherland:
I'm still getting the same thing happening, when stopping a workflow (from UI of CLI):
image

When using hubless, the UIS just disconnects and process stops:
```
.
.
.
[I 2024-05-06 19:14:36.301 CylcUIServer] $ cylc play --color=never --mode live five/run1
[I 2024-05-06 19:14:36.834 CylcUIServer] [data-store] connect_workflow('~sutherlander/five/run1', <dict>)
[I 2024-05-06 19:14:49.404 CylcUIServer] [data-store] disconnect_workflow('~sutherlander/five/run1')
(uiserver) sutherlander@cortex-vbox:cylc-uiserver$ 
```

dwsutherland (changes requested):
I'm still getting the same thing happening, when stopping a workflow (from UI of CLI):
image

When using hubless, the UIS just disconnects and process stops:
```
.
.
.
[I 2024-05-06 19:14:36.301 CylcUIServer] $ cylc play --color=never --mode live five/run1
[I 2024-05-06 19:14:36.834 CylcUIServer] [data-store] connect_workflow('~sutherlander/five/run1', <dict>)
[I 2024-05-06 19:14:49.404 CylcUIServer] [data-store] disconnect_workflow('~sutherlander/five/run1')
(uiserver) sutherlander@cortex-vbox:cylc-uiserver$ 
```

So it appears the UIS stops or crashes when a workflow is stopped/disconnected.

dwsutherland:
Does it make sense to deal with this one:
https://github.com/cylc/cylc-uiserver/issues/584
at the same time?

oliver-sanders:
Can move away from Tornado async interfaces in due course (these are already translated into asyncio calls by the tornado library), Jupyter Server hasn't moved over yet so we may need to wait a while before changing ourselves.

PR #581, opened 2024-04-05
GH Actions: fix sync PRs not happening on push?

MetRonnie:
Maybe there is a bug in using multiple `*`? There are no occurrences of the sync workflow being triggered by push in the history: https://github.com/cylc/cylc-uiserver/actions/workflows/branch_sync.yml


MetRonnie:
Ah nevermind, we don't have that workflow on 1.4.x so I think that's why it's not being triggered

PR #582, opened 2024-04-05
GH Actions: add branch sync workflow from master

MetRonnie:
There are no occurrences of the sync workflow being triggered by push in the history: https://github.com/cylc/cylc-uiserver/actions/workflows/branch_sync.yml

I think this is because the workflow did not exist on the 1.4.x branch


oliver-sanders (approved):
We're probably done with this branch now, but might as well.

PR #592, opened 2024-05-03
Don't protect static assets from XSRF

minrk:
JupyterHub 4.1 applies XSRF checks to authenticated GET requests, which is not necessary for static assets. It would be a valid alternative to not authenticate these requests.

This solves the static asset request, described in https://github.com/jupyterhub/jupyterhub/issues/4800

The userprofile request must be addressed in https://github.com/cylc/cylc-ui

**Check List**

- [x] No dependency changes 
- [x] Tests are not needed for this
- [x] cylc-doc PR not needed

### DIFF
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -184,6 +184,10 @@ class CylcStaticHandler(CylcAppHandler, web.StaticFileHandler):
     def initialize(self, *args, **kwargs):
         return web.StaticFileHandler.initialize(self, *args, **kwargs)
 
+    def check_xsrf_cookie(self):
+        # don't need XSRF protections on static assets
+        return
+
     @web.authenticated
     def get(self, path):
         # authenticate the static handler
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -192,10 +192,15 @@ class CylcStaticHandler(CylcAppHandler, web.StaticFileHandler):
     def get(self, path):
         # authenticate the static handler
         # this provides us with login redirection and token caching
-        # accessing xsrf_token ensures xsrf cookie is set if it needs to be,
-        # e.g. setting it during request for /index.html
-        # to be available for next request to /userprofile
-        self.xsrf_token  # noqa
+        if not path:
+            # Request for /index.html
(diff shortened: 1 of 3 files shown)


MetRonnie:
@oliver-sanders I think this should be on 1.4.5 milestone?

oliver-sanders:
(we will probs skip 1.4.5 since we are near ready for 1.5.0 release but we can merge it through as normal)

minrk:
I added one more commit based on https://github.com/jupyterhub/jupyterhub/issues/4800#issuecomment-2092945342 to ensure the xsrf cookie is set on `GET /cylc/`

cylc/uiserver/handlers.py
```diff
@@ -192,6 +192,10 @@ def check_xsrf_cookie(self):
     def get(self, path):
         # authenticate the static handler
         # this provides us with login redirection and token caching
+        # accessing xsrf_token ensures xsrf cookie is set if it needs to be,
+        # e.g. setting it during request for /index.html
+        # to be available for next request to /userprofile
+        self.xsrf_token  # noqa
```
MetRonnie, line 198: Ah, brilliant!

MetRonnie: > That's _almost_ got it, however I think when the `index.html` is cached by the browser, the GET never goes through our static handler and so the XSRF cookie is not set 😢. This seems more prevalent on Firefox than Chrome

I think setting [`Cache-Control: no-cache`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#response_directives) does the trick 🤞

```suggestion
        if not path:
            # Request for /index.html
            # Accessing xsrf_token ensures xsrf cookie is set
            # to be available for next request to /userprofile
            self.xsrf_token
            # Ensure request goes through this method even when cached so
            # that the xsrf cookie is set on new browser sessions
            # (doesn't prevent browser storing the response):
            self.set_header('Cache-Control', 'no-cache')
```

MetRonnie (approved):
🚀

oliver-sanders (approved):
Makes sense, happy removing xsrf checks for static resources, tested with JupyterHub 4.1.5, couldn't force a page load error. Many thanks!

PR #594, opened 2024-05-09
Avoid requiring `jupyter_config.py` in to be in `uiserver` subdirectory

### LINKED ISSUE #565: UI Server Configuration does not seem to honour $CYLC_SITE_CONF_PATH environment variable
### Description

The [documentation](https://cylc.github.io/cylc-doc/latest/html/reference/config/ui-server.html#ui-server-config) suggests that the `$CYLC_SITE_CONF_PATH` environment variable can be set to advise Cylc UI Servers where to find the `jupyter_config.py` file. When running the Cylc Hub with `sudospawner`, setting the `$CYLC_SITE_CONF_PATH` in `sudospawner-singleuser`:

```
#!/bin/bash -l
export CYLC_SITE_CONF_PATH="/data/jupyterhub/config/"
# Delegate to the Cylc hub server
exec "$(dirname "$0")/cylc" "hubapp" $@
```

does not result in the `jupyter_config.py` in that directory being used.

Further, if the  `jupyter_config.py` in the Cylc Hub's pre-packaged default location `<CONDA_ENV>/site-packages/cylc/uiserver/jupyter_config.py` is not readable by the workflow owner, no permissions are provided but no warning is raised either. This makes debugging site_authorization issues a little harder.

### Reproducible Example

Spawn a UI server, after exporting `CYLC_SITE_CONF_PATH` to be a different directory, with an all-user readable  `jupyter_config.py` that grants access to other users.

Make sure that the UI server's pre-packaged default location for `jupyter_config.py` is unreadable to non-owner users/and or does not grant any access to other users.

Attempt to connect to that UI server as another user. Note that access is not granted.

### Expected Behaviour

Exporting `CYLC_SITE_CONF_PATH` before calling `cylc hubapp` would result in the `jupyter_config.py` file under that directory to be used.

If no `jupyter_config.py` files are found/readable, a warning is printed in addition to stating that users have no access.

wxtim:
Closes #565 

Previously it was assumed to be in `CYLC_SITE_CONF_PATH/uiserver`.

Ensure that user is warned if file at `CYLC_SITE_CONF_PATH` is
* Unreadable
* Non-existent

**Check List**

### DIFF
--- a/changes.d/594.fix
+++ b/changes.d/594.fix
@@ -1,0 +1,1 @@
+Avoid requiring jupyter_config.py in to be in uiserver subdirectory when using CYLC_SITE_CONF_PATH
--- a/cylc/uiserver/config_util.py
+++ b/cylc/uiserver/config_util.py
@@ -30,7 +30,8 @@ from cylc.uiserver import (
 )
 
 # base configuration - always used
-DEFAULT_CONF_PATH: Path = Path(uis_pkg).parent / 'jupyter_config.py'
+CONF_FILE_NAME = 'jupyter_config.py'
+DEFAULT_CONF_PATH: Path = Path(uis_pkg).parent / CONF_FILE_NAME
 UISERVER_DIR = 'uiserver'
 # UIS configuration dirs
 SITE_CONF_ROOT: Path = Path(
--- a/cylc/uiserver/jupyterhub_config.py
+++ b/cylc/uiserver/jupyterhub_config.py
@@ -25,8 +25,8 @@ import os
 from pathlib import Path
 
 from cylc.uiserver.config_util import (
+    CONF_FILE_NAME,
     DEFAULT_CONF_PATH,
-    UISERVER_DIR,
     USER_CONF_ROOT,
     SITE_CONF_ROOT,
     get_conf_dir_hierarchy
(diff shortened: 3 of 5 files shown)


oliver-sanders (changes requested):
> Previously it was assumed to be in CYLC_SITE_CONF_PATH/uiserver.

This was correct and consistent with cylc-flow which looks in `$CYLC_SITE_CONF_PATH/flow`.

The `CYLC_SITE_CONF_PATH` directory is supposed to look like this:

```
.
|-- flow
|   `-- global.cylc
`-- uiserver
    `-- jupyter_config.py
```

The intention is that it can be used to configure all of the Cylc components.

wxtim:
> As of the current versions setting CYLC_SITE_CONF_PATH to the root of the above structure works fine for me.

So this can be closed?

oliver-sanders:
Check the issue.

oliver-sanders:
Had a read through the issue, I think these are the main points:

1. The config file didn't get picked up in the location the docs suggested.
  * Check the current behaviour (i.e. `$CYLC_SITE_CONF_DIR/uiserver/jupyter_config.py`) makes sense and is consistent with cylc-flow.
  * If so, amend the docs as needed.
2. If the config file is not readable, there is no error.
  * This is likely Jupyter Server behaviour (try reproducing this using `jupyter server`).
  * If this is special Cylc behaviour, **not** Jupyter Server behaviour, then it's a bug.
  * Otherwise, we could consider pushing an upstream change to Jupyter Server.
3. If no config files are found / readable, they would like a warning stating that users have no access.
  * I'm not sure I get this bit.

wxtim:
> Had a read through the issue, I think these are the main points:

Sorry, so have I, just not been high enough on my todo-list. It _is_ on my radar.

wxtim:
> The config file didn't get picked up in the location the docs suggested.
> 
>     * Check the current behaviour (i.e. `$CYLC_SITE_CONF_DIR/uiserver/jupyter_config.py`) makes sense and is consistent with cylc-flow.
> 
>     * If so, amend the docs as needed.

This works as documented [here](https://cylc.github.io/cylc-doc/stable/html/reference/config/global.html#envvar-CYLC_SITE_CONF_PATH), although there might be a case for an extra note on the uiserver app docs.

wxtim:
@oliver-sanders  -  I'm not convinced that this PR is necessary.

hjoliver:
Can we close this then?

Issue #585, opened 2024-04-22
data_store_mgr: clear data from workflows that no longer have active subscriptions


oliver-sanders:
We need the top-level workflow data from every workflow in a user's `~/cylc-run` directory.

E.G:
* Workflow ID
* Host
* Port
* etc

Additionally, we require any other objects that the user has requested, e.g. Tasks, Jobs, etc.

When a user navigates away from a workflow, we no longer need to preserve these Tasks, Jobs, etc in the store and should purge it in due course to avoid stopped workflow data accumulating.

One approach would be to set a timer when the last subscription for a workflow is closed [1]. Unless another subscription on the workflow [1] is opened when this timer expires, then we purge the data.

This issue is similar to #568 which is about reducing the number of Protobuf "topics" we subscribe to as this issue also requires the monitoring of active subscriptions. This issue differs in that it is about clearing the data in the store that we have already acquired but is no longer of use rather than reducing the amount we request.

[1] Excluding top-level only subscriptions (see above).

Issue #568, opened 2024-03-26
protobuf: dynamic subscriptions


oliver-sanders:
> **Background:**
>
> We use GraphQL as an API, but we do not use it as a "store". The data does into a Protobuf serialisation, we then "resolve" GraphQL requests off of this serialisation, re-serialising it into JSON format for transfer to the UI.
>
> So GraphQL is the external API for UI-UIS comms, Protobuf is both the store and the internal API for UIS-Flow comms.

Presently, the UIS subscribes to all topics for all active workflows (running, paused, stopping). This is vastly surplus to requirements as (at present) users can only look at one workflow at a time in the UI. This may cause excess load on the UIS and, potentially, high memory usage.

In order to field requests we only need to subscribe to the subset of topics that are required to satisfy active GraphQL subscriptions. E.G, to satisfy the requirements of the UI when looking at the tree view for workflow "a" we would need:

1. A subscription to the `WORKFLOW` topic for all active workflows. This is required for workflow listing (for the GScan view).
2. A subscription to the `TASK_PROXIES`, `FAMILY_PROXIES`, `TASKS` and `JOBS` topics for workflow "a" only (for the tree view).

To get started, it might be easier to consider two subscription levels:
* `WORKFLOW` - for all active workflows.
* `ALL` - for any workflows that match active GraphQL subscriptions/queries.

Getting the data store to handle two different subscription levels is straight forward:

<details>
<summary>Subscription Level Patch</summary>

```diff
diff --git a/cylc/uiserver/data_store_mgr.py b/cylc/uiserver/data_store_mgr.py
index 101fc36..578dcac 100644
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -36,7 +36,7 @@ from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
 from pathlib import Path
 import time
-from typing import Dict, Optional, Set
+from typing import Dict, Optional, Set, List
 
 from cylc.flow.exceptions import WorkflowStopped
 from cylc.flow.id import Tokens
@@ -80,16 +80,31 @@ class DataStoreMgr:
     RECONCILE_TIMEOUT = 5.  # seconds
     PENDING_DELTA_CHECK_INTERVAL = 0.5
 
+    SUBSCRIPTION_LEVELS = {
+        'all': (
+            'pb_entire_workflow',
+            {ALL_DELTAS.encode('utf-8'), b'shutdown'},
[...]

PR #597, opened 2024-05-29
Sync minimal flow data

dwsutherland:
addresses https://github.com/cylc/cylc-uiserver/issues/585
addresses https://github.com/cylc/cylc-uiserver/issues/568

Sibling https://github.com/cylc/cylc-flow/pull/6113

This change aims to:
- [x] interrogate the graphql subscription for data requirements.
- [x] Sync only the data required.
- [x] Reduce the protobuf subscription topics and all data dump size.
- [x] Handle data requirements for queries.

This first attempt is functional, reads the fragments requested by the UI with
```{'AddedDelta', 'WorkflowData', 'UpdatedDelta'}```
being the workflow summary information of the dashboard subscription.

~~The UI doesn't seem to reduce the subscription if you go from tree view to dashboard, however, closing the UI will reduce it back down.~~

- [x] One thing we need to be able to handle is general queries hitting the UIS, as we'll want CLI via UIS in at some point and to be able to handle random queries.. For this I think we can use that timer concept, keeping a full sync of workflows queried for some amount of time.. **Done**

**Check List**

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -103,7 +103,22 @@ class DataStoreMgr:
         self.log = log
         self.data = {}
         self.w_subs: Dict[str, WorkflowSubscriber] = {}
-        self.topics = {ALL_DELTAS.encode('utf-8'), b'shutdown'}
+        self.topics = {
+            ALL_DELTAS.encode('utf-8'),
+            WORKFLOW.encode('utf-8'),
+            b'shutdown'
+        }
+        # If fragments in graphql sub for minimal sync
+        self.min_sync_fragments = {
+            'AddedDelta',
+            'WorkflowData',
+            'UpdatedDelta'
+        }
+        # set of workflows to sync all data
+        self.full_sync_workflows = set()
+        self.full_sync_gql_subs = set()
+        # dict of workflow full sync subscriber IDs
+        self.full_sync_workflow_gql_subs = {}
         self.loop = None
         self.executor = ThreadPoolExecutor(max_threads)
         self.delta_queues = {}
@@ -163,6 +178,9 @@ class DataStoreMgr:
 
         self.delta_queues[w_id] = {}
 
+        # setup sync subscriber set
+        self.full_sync_workflow_gql_subs[w_id] = set()
+
         # Might be options other than threads to achieve
         # non-blocking subscriptions, but this works.
         self.executor.submit(
@@ -172,7 +190,25 @@ class DataStoreMgr:
             contact_data[CFF.HOST],
             contact_data[CFF.PUBLISH_PORT]
         )
-        successful_updates = await self._entire_workflow_update(ids=[w_id])
+
+        result = await self.workflow_data_update(w_id, minimal=True)
+
+        if result:
+            # don't update the contact data until we have successfully updated
+            self._update_contact(w_id, contact_data)
+
+    @log_call
+    async def workflow_data_update(
+        self,
+        w_id: str,
+        minimal: Optional[bool] = None
+    ):
+        if minimal is None:
+            minimal = w_id in self.full_sync_workflows
+        successful_updates = await self._entire_workflow_update(
+            ids=[w_id],
+            minimal=minimal
+        )
 
         if w_id not in successful_updates:
             # something went wrong, undo any changes to allow for subsequent
@@ -180,6 +216,7 @@ class DataStoreMgr:
             self.log.info(f'failed to connect to {w_id}')
             self.disconnect_workflow(w_id)
             return False
+        return True
 
     @log_call
     def disconnect_workflow(self, w_id, update_contact=True):
@@ -206,6 +243,9 @@ class DataStoreMgr:
         if w_id in self.w_subs:
             self.w_subs[w_id].stop()
             del self.w_subs[w_id]
+        if w_id in self.full_sync_workflow_gql_subs:
+            del self.full_sync_workflow_gql_subs[w_id]
+            self.full_sync_workflows.discard(w_id)
 
     def get_workflows(self):
         """Return all workflows the data store is currently tracking.
@@ -283,8 +323,18 @@ class DataStoreMgr:
             # close connections
             self.disconnect_workflow(w_id)
             return
-        self._apply_all_delta(w_id, delta)
-        self._delta_store_to_queues(w_id, topic, delta)
+        elif topic == WORKFLOW:
+            if w_id in self.full_sync_workflows:
+                return
+            self._apply_delta(w_id, WORKFLOW, delta)
+            # might seem clunky, but as with contact update, making it look
+            # like an ALL_DELTA avoids changing the resolver in cylc-flow
+            all_deltas = DELTAS_MAP[ALL_DELTAS]()
+            all_deltas.workflow.CopyFrom(delta)
+            self._delta_store_to_queues(w_id, ALL_DELTAS, all_deltas)
+        elif w_id in self.full_sync_workflows:
+            self._apply_all_delta(w_id, delta)
+            self._delta_store_to_queues(w_id, topic, delta)
 
     def _clear_data_field(self, w_id, field_name):
         if field_name == WORKFLOW:
@@ -294,24 +344,28 @@ class DataStoreMgr:
 
     def _apply_all_delta(self, w_id, delta):
         """Apply the AllDeltas delta."""
-        delta_times = self.data[w_id]['delta_times']
         for field, sub_delta in delta.ListFields():
-            delta_time = getattr(sub_delta, 'time', 0.0)
-            # If the workflow has reloaded clear the data before
-            # delta application.
-            if sub_delta.reloaded:
-                self._clear_data_field(w_id, field.name)
-                delta_times[field.name] = 0.0
-            # hard to catch errors in a threaded async app, so use try-except.
-            try:
-                # Apply the delta if newer than the previously applied.
-                if delta_time >= delta_times.get(field.name, 0.0):
-                    apply_delta(field.name, sub_delta, self.data[w_id])
-                    delta_times[field.name] = delta_time
-                    if not sub_delta.reloaded:
-                        self._reconcile_update(field.name, sub_delta, w_id)
-            except Exception as exc:
-                self.log.exception(exc)
+            self._apply_delta(w_id, field.name, sub_delta)
+
+    def _apply_delta(self, w_id, name, delta):
+        """Apply delta."""
+        delta_times = self.data[w_id]['delta_times']
+        delta_time = getattr(delta, 'time', 0.0)
+        # If the workflow has reloaded clear the data before
+        # delta application.
+        if delta.reloaded:
+            self._clear_data_field(w_id, name)
+            delta_times[name] = 0.0
+        # hard to catch errors in a threaded async app, so use try-except.
+        try:
+            # Apply the delta if newer than the previously applied.
+            if delta_time >= delta_times.get(name, 0.0):
+                apply_delta(name, delta, self.data[w_id])
+                delta_times[name] = delta_time
+                if not delta.reloaded:
+                    self._reconcile_update(name, delta, w_id)
+        except Exception as exc:
+            self.log.exception(exc)
 
     def _delta_store_to_queues(self, w_id, topic, delta):
         # Queue delta for graphql subscription resolving
@@ -370,7 +424,9 @@ class DataStoreMgr:
                 self.log.exception(exc)
 
     async def _entire_workflow_update(
-        self, ids: Optional[list] = None
+        self,
+        ids: Optional[list] = None,
+        minimal: Optional[bool] = False
     ) -> Set[str]:
         """Update entire local data-store of workflow(s).
 
@@ -418,6 +474,8 @@ class DataStoreMgr:
                         key: value.last_updated
                         for key in DATA_TEMPLATE
                     }
+                    continue
+                elif minimal:
                     continue
                 new_data[field.name] = {n.id: n for n in value}
             self.data[w_id] = new_data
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by dwsutherland** [comment]
more updates inc..

**Comment by dwsutherland** [comment]
Ok I've added more features and taken some ideas from #568 .. 
Turns out you can subscribe/unsubscribe from topics without restarting the connection:
https://github.com/cylc/cylc-uiserver/issues/568

If you print the `topic` near the top of the `data_store_mgr.py` `_update_workflow_data` method, you can see it changing when you open and close the graph view:
image


It's even more noticeable with multiple workflows running, and switching between them in the UI. 
image

**Comment by dwsutherland** [comment]
Not sure the reason for this `mypy` failure:
image

`new_data[field.name].CopyFrom(value)` has always been there... and passes locally

Also, I get a lot of this:
```
ERROR cylc/uiserver/tests/test_workflows_mgr_data_store.py::test_data_store_changes[active-active-actions8] - AttributeError: 'CylcUIServer' object has no attribute 'profiler'. Did you mean: 'profile'?

ERROR cylc/uiserver/utils.py::uiserver.utils.fmt_call - AttributeError: 'CylcUIServer' object has no attribute 'profiler'. Did you mean: 'profile'?
```
from `pytest` locally.. I assume I'm missing some sort of `profiler`

**Review by oliver-sanders — COMMENTED** [review]
@dwsutherland, I don't think we are going to be able to get this ready for 1.5.0, however, provided we don't need to change much of the cylc-flow code we can make a UIS release for this change at any point so we don't necessarily need to tether this work to cylc-flow milestones.

There are two technical barriers to cross with this work:
1. Determining topics we must subscribe to in order to satisfy a query/subscription.
2. Determining when we no longer require to subscribe to a topic (as a result of query completion / subscription completion).

I think we need to really pin these two mechanisms down before proceeding as they are quite involved and very fundamental to UIS functionality.

File: cylc/uiserver/data_store_mgr.py [hunk 1]
```diff
@@ -502,3 +567,64 @@ def _get_status_msg(self, w_id: str, is_active: bool) -> str:
...
+    def graphql_sub_interrogate(self, sub_id, info):
+        """Scope data requirements."""
+        fragments = set(info.fragments.keys())
+        minimal = (
+            (
+                fragments
+                <= SUBSCRIPTION_LEVELS[MIN_LEVEL]['criteria']['fragments']
+            )
+            and bool(fragments)
+        )
```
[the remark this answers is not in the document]
dwsutherland: They already do, once the subscription ends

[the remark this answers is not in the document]
dwsutherland: > I don't think that we can reliably determine the topic(s) we need to subscribe to by looking at the GraphQL fragments

We can, and it works, but yes not reliable.. if we could pick up the fragments from the associated UI code then maybe..
Parsing for every field requiring resolvers into sets could be another way..

dwsutherland:
> provided we don't need to change much of the cylc-flow code

It can go in now, and not break anything.. So if we can at least do that for 8.3.0, then we can move forward with this later.

> I don't think that we can reliably determine the topic(s) we need to subscribe to by looking at the GraphQL fragments

We can, and it works, but yes not reliable (if UI graphql changes, it could break it).. if we could pick up the fragments from the associated UI code then maybe..  For any subscription that doesn't contain those fragments (i.e. a bespoke subscription) it will default to max.
Parsing for every field requiring resolvers into sets could be another way..
Should only need to happen at the start of a new subscription, so we can probably get away with something more intensive then.

cylc/uiserver/data_store_mgr.py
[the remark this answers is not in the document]
dwsutherland: https://github.com/dwsutherland/cylc-uiserver/blob/2605026328a6168664c6a32d27131c1652bc12af/cylc/uiserver/data_store_mgr.py#L624-L630

dwsutherland:
Have a solution for queries, hopefully post it up tomorrow..

dwsutherland:
Query sync level change and expire implemented, and functional:
image

The expiry is set to 60sec, with a periodic callback every 30sec to check and demote. (~30sec expiry in the above example)

The expiry will take precedence over subscription start/stops, so the level will remain.
Otherwise the levels promote/demote immediately in accordance with subscription start/stop.

cylc/uiserver/data_store_mgr.py
```diff
@@ -56,6 +56,27 @@
...
+MAX_LEVEL = 'max'
+SUBSCRIPTION_LEVELS = {
+    MIN_LEVEL: {
+        'topics': {WORKFLOW.encode('utf-8'), b'shutdown'},
+        'criteria': {
+            'fragments': {
+                'AddedDelta',
+                'WorkflowData',
+                'UpdatedDelta'
+            },
```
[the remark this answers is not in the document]
oliver-sanders: @dwsutherland, I had a poke at the `info` object and found that it contains the complete parsed query.

I had a quick go at using this to extract the requested field types. It's crude, but it seems to be good enough to determine whether a query/subscription is requesting top-level Cylc types e.g. `tasks`, `taskProxies`, `edges` which is all we need to know for this purpose.

```python
from graphql import language


TOP_LEVEL_THINGGIES = {
    # TODO: populate from somewhere else?
    'tasks',
    'taskProxies',
    'families',
    'familyProxies',
    'edges',
}


def iter_top_level_types(info):
    """Loop through top-level Cylc types requested in a query/subscription."""
    ret = set()
    for selection_set in iter_selection_sets(info):
        for selection in selection_set.selections:
            # TODO: check this handles query nesting correctly, it appears to
            if isinstance(selection, language.ast.Field):
                field_name = selection.name.value
                if field_name in TOP_LEVEL_THINGGIES:
                    if field_name not in ret:
                        ret.add(field_name)
                        yield field_name


def iter_selection_sets(info):
    """Loop through top-level selection sets in the query definitions."""
    if hasattr(info, 'operation'):
        yield info.operation.selection_set
    if hasattr(info, 'fragments'):
        for fragment in info.fragments.values():
            yield fragment.selection_set
```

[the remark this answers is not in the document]
dwsutherland: I agree we can do this... 
IMO - this can just be a refinement change, the bones are ready in the PR, and we can just build on it by changing the criteria from fragments to field sets.

But we're running out of time for 8.3.0.. so going to have to bump the cylc-flow end to 8.4

dwsutherland:
> > provided we don't need to change much of the cylc-flow code
> 
> It can go in now, and not break anything.. So if we can at least do that for 8.3.0, then we can move forward with this later.
> 

I was wrong on this, I think the current UIS will break with the cylc-flow end in first (because the data-store manager on the UIS doesn't contain the methods that will be called by the resolvers)

So they both need to go in .. which is a shame .. because I assume the next 8.x release will be some way off..

dwsutherland:
I think I may have found an easier way to hook into the query/(subscriptions) before execution .. Actually, at other points of the execution too:
https://github.com/graphql-python/graphene-tornado/blob/master/graphene_tornado/graphql_extension.py

So, probably a good thing it doesn't go in yet

cylc/uiserver/data_store_mgr.py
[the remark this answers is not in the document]
oliver-sanders: So long as this doesn't require changes to cylc-flow (which is likely necessary for back-compat), then this can be put into a cylc-uiserver release at any point (no need to wait for any particular cylc-flow release).

oliver-sanders:
> because I assume the next 8.x release will be some way off..

8.3.0 was a monster, we're hoping to resume a more regular release cycle once this is out the way, 8.4.0 could be ~1 month away.

However, back compat is important, otherwise the GUI won't be able to connect to older workflows, so a cylc-uiserver side approach is definitely preferable. Note, if we do need to cut off back-compat, the way we do this is to increment the API version number in cylc-flow (currently 5). There is an API number filter in the scan mechanism in cylc-uiserver which will filter out older or newer workflows avoiding runtime problems.

oliver-sanders:
Reminder (largely for myself): This work requires an interface for tracking subscriptions so we can tell what things users are actively subscribed to. This should work by parsing the `workflows` GraphQL argument or similar (we can't *rely* on subscription query format, variable name conventions, etc).

We should be good to resume this work post the graphql-core v3 changes.

dwsutherland:
Have rebased both ends, and still works:
image
(switching between dashboard and tree view switches the data sync)


Will work on moving away from Fragments (and then un-draft)

dwsutherland:
I've made the refinement changes, appears to work well:
[dynamic-sync-level.webm](https://github.com/user-attachments/assets/261bb63a-f8ea-4dff-adfb-d8ce292363ef)

I did, however, add a workaround to bind task and family proxies together.. Because they would sometimes arrive at the UI out of order, which would cause the families to not be pruned if the task was not pruned first.. (so really the nature of how the UI works)

cylc/uiserver/tests/test_data_store_mgr.py
oliver-sanders, line 1: Should add some tests for the GraphQL field extraction and subscription topic matching functions.

cylc/uiserver/data_store_mgr.py
```diff
@@ -507,3 +686,94 @@ def _get_status_msg(self, w_id: str, is_active: bool) -> str:
...
+        extract_ast_fields(info.operation, fields)
+        for frag in getattr(info, 'fragments', {}).values():
+            extract_ast_fields(frag, fields)
+        self.sync_level_graphql_subs[sub_id] = {'shutdown'}
+        sync_level_subs = self.sync_level_graphql_subs[sub_id]
+        for category in DATA_TEMPLATE:
+            if fields.intersection(
+                SUBSCRIPTION_CATALOGUE[category]['criteria']
+            ):
+                sync_level_subs.add(category)
```
oliver-sanders, line 746: Unused?

cylc/uiserver/data_store_mgr.py
```diff
@@ -507,3 +686,94 @@ def _get_status_msg(self, w_id: str, is_active: bool) -> str:
...
+
+    async def graphql_sub_data_match(self, w_id, sub_id):
+        """Match store data level to requested graphql subscription."""
+        sync_level_wsubs = self.workflow_sync_graphql_subs[w_id]
+        sync_level_wsubs.add(sub_id)
+
+        # set new sync level
+        await self._update_subscription_level(w_id)
```
oliver-sanders, line 754: Unused?

cylc/uiserver/data_store_mgr.py
```diff
@@ -507,3 +686,94 @@ def _get_status_msg(self, w_id: str, is_active: bool) -> str:
...
+
+    async def graphql_sub_discard(self, sub_id):
+        """Discard graphql subscription references."""
+        if sub_id in self.sync_level_graphql_subs:
+            del self.sync_level_graphql_subs[sub_id]
+        for w_id in self.workflow_sync_graphql_subs:
+            if sub_id not in self.workflow_sync_graphql_subs[w_id]:
+                continue
+            self.workflow_sync_graphql_subs[w_id].discard(sub_id)
+            await self._update_subscription_level(w_id)
```
oliver-sanders, line 764: Unused?

cylc/uiserver/data_store_mgr.py
```diff
@@ -507,3 +686,94 @@ def _get_status_msg(self, w_id: str, is_active: bool) -> str:
...
+        if expire_delay is None:
+            expire_delay = QUERY_SYNC_EXPIRY
+        expire_time = time.time() + expire_delay
+        for w_id in w_ids:
+            if self.workflow_sync_level[w_id] == level:
+                # Already required level
+                continue
+            self.workflow_query_sync_timers[w_id]['expiry'] = expire_time
+            self.workflow_query_sync_timers[w_id]['level'].update(level)
+            await self._update_subscription_level(w_id)
```
oliver-sanders, line 784: Unused?

oliver-sanders (commented):
> I've made the refinement changes, appears to work well:

:tada:

<br />

> I did, however, add a workaround to bind task and family proxies together.. Because they would sometimes arrive at the UI out of order, which would cause the families to not be pruned if the task was not pruned first.. (so really the nature of how the UI works)

For cylc-ui the main thing is to ensure that we don't subscribe to *everything* just to power the workflows sidebar. So lumping tasks and families together is no biggie.

cylc/uiserver/tests/test_data_store_mgr.py
dwsutherland: There are tests on the `cylc-flow` end testing subscriptions, but yes the level extraction is a dummy method on that end... Just trying to get the coverage up there at the moment...

cylc/uiserver/data_store_mgr.py
dwsutherland: Called by the `cylc-flow` defined resolver.

cylc/uiserver/data_store_mgr.py
dwsutherland: Called by the `cylc-flow` defined resolver.

cylc/uiserver/data_store_mgr.py
dwsutherland: Called by the `cylc-flow` defined resolver.

cylc/uiserver/data_store_mgr.py
dwsutherland: Called by the `cylc-flow` defined resolvers.

cylc/uiserver/tests/test_data_store_mgr.py
dwsutherland: Want the ws tests in first:
https://github.com/cylc/cylc-uiserver/pull/768

Then I can just add to the subscription tests, to find the level of syncs match what's expected with the document.

oliver-sanders:
> The UI doesn't seem to reduce the subscription if you go from tree view to dashboard, however, closing the UI will reduce it back down.

When you navigate back to the Dashboard, the UI doesn't *reduce* the subscription, it *stops* it.

There are two subscriptions:
* *global:*
  * The "global" subscription powers the dashboard and the workflows sidebar.
  * It subscribes to all workflows indiscriminately.
  * It lives for the lifetime of the UI.
* *workflow:*
  * The "workflow" subscription powers the main workflow view (`#/workspace`) as well as the standalone views (`#/tree`, `#/graph`, etc).
  * It only ever subscribes to a single workflow.
  * When you navigate to a workflow (e.g, `#/workspace`), the subscription is created.
    * If you navigate back (e.g. to the Dashboard), the subscription is cancelled.
    * If you navigate to another workflow, the subscription is cancelled, then re-created (with the workflow ID changed).
  * In the `#/workspace` view:
    * When a new tab is opened, the subscription will be reissued if the data requirements of the new tab are not already covered by the existing subscription.
    * When a tab is closed, nothing happens - we don't bother re-issueing the subscription at present.

At the moment, this PR does not detect the subscription stop event, so the subscription level will only reduce if the UI is closed.

The other thing we need to handle is an unplanned subscription termination, e.g, due to network disconnect, client sleep, browser tab crash, etc.

dwsutherland:
> When you navigate back to the Dashboard, the UI doesn't reduce the subscription, it stops it.

Works for me:
works_for_me

> At the moment, this PR does not detect the subscription stop event, so the subscription level will only reduce if the UI is closed.

graphiql subscription shows this too:
graphiql_works_too


So it detects the subscription stop for me at least

dwsutherland:
Have added back compat:
image
both ways
image

oliver-sanders:
> The UI doesn't seem to reduce the subscription if you go from tree view to dashboard, however, closing the UI will reduce it back down.
> -- @dwsutherland 
>
>> Works for me:
>> So it's not true for me at least
>> -- @dwsutherland 

Sorry, I'm confused, which is it?

dwsutherland:
> > The UI doesn't seem to reduce the subscription if you go from tree view to dashboard, however, closing the UI will reduce it back down.
> > -- @dwsutherland

Apologies, It seems like I hadn't updated the description/that-line since I first posted this (last year)

> > > Works for me:
> > > So it's not true for me at least
> > > -- @dwsutherland
> 
> Sorry, I'm confused, which is it?

Sorry:

> So it's not true for me at least

It detects the subscription stop for me (as shown).

cylc/uiserver/tests/test_data_store_mgr.py
oliver-sanders: https://github.com/cylc/cylc-uiserver/pull/768 merged

PR #606, opened 2024-06-19
adapt to cylc-flow code refactor

oliver-sanders:
Goes with https://github.com/cylc/cylc-flow/pull/6136


oliver-sanders:
(Draft due to a hitch with the upstream PR)

PR #607, opened 2024-06-24
Log any cat-log failure when listing log files

MetRonnie:
When trying to set up `cylc hub`, Dave encountered cat-log problems which were swallowed by the UIS. This PR ensures any such errors are logged.

**Check List**

- [x] Tests are not needed
- [x] No changelog entry needed as minor
- [x] No docs needed

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -16,14 +16,19 @@
 """GraphQL resolvers for use in data accessing and mutation of workflows."""
 
 import asyncio
-from contextlib import suppress
 import errno
-from getpass import getuser
 import os
-from copy import deepcopy
 import signal
-from subprocess import Popen, PIPE, DEVNULL
+from contextlib import suppress
+from copy import deepcopy
+from getpass import getuser
+from subprocess import (
+    DEVNULL,
+    PIPE,
+    Popen,
+)
 from typing import (
+    TYPE_CHECKING,
     Any,
     AsyncGenerator,
     Callable,
@@ -31,31 +36,27 @@ from typing import (
     Iterable,
     List,
     Optional,
-    TYPE_CHECKING,
     Tuple,
     Union,
 )
 
-from graphql.language.base import print_ast
 import psutil
-
 from cylc.flow.data_store_mgr import WORKFLOW
-from cylc.flow.exceptions import (
-    ServiceFileError,
-    WorkflowFilesError,
-)
+from cylc.flow.exceptions import ServiceFileError, WorkflowFilesError
 from cylc.flow.id import Tokens
 from cylc.flow.network.resolvers import BaseResolvers
-from cylc.flow.scripts.clean import CleanOptions
-from cylc.flow.scripts.clean import run
+from cylc.flow.scripts.clean import CleanOptions, run
+from graphql.language.base import print_ast
 
 if TYPE_CHECKING:
     from concurrent.futures import Executor
     from logging import Logger
     from optparse import Values
-    from graphql import ResolveInfo
+
     from cylc.flow.data_store_mgr import DataStoreMgr
     from cylc.flow.option_parsers import Options
+    from graphql import ResolveInfo
+
     from cylc.uiserver.workflows_mgr import WorkflowsManager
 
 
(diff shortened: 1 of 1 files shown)


MetRonnie:
I'm inclined to say no automated tests are needed for this

hjoliver (approved):

cylc/uiserver/resolvers.py
```diff
@@ -438,22 +438,26 @@ async def cat_log_files(cls, id_: Tokens):
...
+        ret_code = await proc_job.wait()
 
-        # MOTD returned in stderr, no use in returning
-        out_job, _ = await proc_job.communicate()
-        if out_job:
+        out, err = await proc_job.communicate()
+        if ret_code:
+            raise Exception(
+                f"Command failed ({ret_code}): {' '.join(cmd)}\n{err.decode()}"
+            )
```
oliver-sanders, line 454: I don't think raising an exception is the right way to go here. This leaves traceback (that we don't want) in the UIS stderr, however, there is no record of this in the UIS log (exceptions don't go to the log by default).

Suggest logging this with `log.error`.

hjoliver: OK, I agree with that.

MetRonnie: Good spot on the exception not being printed in the log file.

However I think raising an exception is the way to go. In order to return this error to the UI so the user can get the feedback that something has gone wrong (via an error snackbar - cylc/cylc-ui#1842), the only other choice is to add a field on the schema (e.g. called `error`), which seems like an abuse of graphQL?

https://github.com/cylc/cylc-uiserver/blob/3395e8baad095eb98400d41c12f877c298961d20/cylc/uiserver/schema.py#L555-L565

Unfortunately Graphene seems to lack any documentation on error handling.

In terms of preventing `GraphQLLocatedError` traceback from appearing in stderr, I think this would be a separate issue as it happens any time an uncaught exception occurs

oliver-sanders: > the only other choice is to add a field on the schema (e.g. called error), which seems like an abuse of graphQL?

Using the error field is the correct way to send errors back to the client (it is not an abuse). We already do this for other functionality, e.g:

https://github.com/cylc/cylc-uiserver/blob/f6a05fc5212d7f3bdbd6d525f5660135fa198f75/cylc/uiserver/resolvers.py#L396

Note, `data` and `error` are standard GraphQL return fields, check out the GraphQL docs rather than Graphene.

MetRonnie: > `data` and `error` are standard GraphQL return fields

`data` and/or **`errors`** are indeed the GraphQL protocol for the JSON response. But these should not be confused with fields _within_ the `data` response.

There is no apparent way (that I am aware of) of responding with `errors` without causing a `GraphQLLocatedError` traceback. If you capture any exceptions and return them in the resolver function, it still leads to the same outcome.

What you have linked to in the log file contents resolver is returned within the `data` response; it is included as a field called `error` in the schema (but could be called anything we like)

https://github.com/cylc/cylc-uiserver/blob/f6a05fc5212d7f3bdbd6d525f5660135fa198f75/cylc/uiserver/schema.py#L618-L622

```json
{
  "data": {
    "logs": {
      "lines": [
        "2024-07-08T12:39:47+01:00 INFO - [20240709T0500Z/p/01:running] => succeeded\n",
        "2024-07-08T12:39:47+01:00 INFO - [20240709T0500Z/q/01:running] => succeeded\n"
      ],
      "connected": null,
      "path": null,
      "error": null,
      "__typename": "Logs"
    }
  }
}
```
I think it makes sense to do it this way, because `error` is mainly used as a _result_ for when the log has been deleted, which we expect to occur frequently.

I guess we could do a similar thing for listing the log files, even if we don't expect this to fail in the normal course of things. Luckily we wouldn't have to worry about backward compatibility in the schema because cylc-flow is not involved here.

MetRonnie:
I realised that there are a few ways `cat-log` will exit non-zero when navigating in the UI (e.g. navigating to `/log/non-exist`), so have settled for just logging the stderr and returning an empty list as before

oliver-sanders (approved):

PR #612, opened 2024-07-02
More gracefully override JPS `ExtensionApp.config_file_paths`

MetRonnie:
This allows patching in the tests so the tests don't pick up our user/site configs. Currently the tests fail locally for me because they pick up my config.

### DIFF
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -59,7 +59,7 @@ import os
 from pathlib import Path, PurePath
 import sys
 from textwrap import dedent
-from typing import List
+from typing import List, Optional
 
 from pkg_resources import parse_version
 from tornado import ioloop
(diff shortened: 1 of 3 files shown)


cylc/uiserver/app.py
```diff
@@ -421,6 +413,21 @@ def __init__(self, *args, **kwargs):
...
+                [
+                    SITE_CONF_ROOT,  # site configuration
+                    USER_CONF_ROOT,  # user configuration
+                ], filename=False
+            )
+            # Next include currently needed for directory making
+            ret.insert(0, str(Path(uis_pkg).parent))  # packaged config
+            ret.reverse()
+            self._config_file_paths = ret
+        return self._config_file_paths
```
oliver-sanders, line 429: This change does not appear to be necessary, the only bit that's needed is the `autouse`?

MetRonnie: Couldn't get it to work without this change. Not sure why, but pyright is not happy with overriding a property with a class attribute anyway (see https://github.com/microsoft/pyright/issues/3646#issuecomment-1172558289)

oliver-sanders: Monkeypatching seems to work fine for me (with your config)?

I found the `config_file_paths` definition in Jupyter_Core and it is indeed a property so this makes sense :+1:.

oliver-sanders (approved):

PR #619, opened 2024-08-07
shutdown: terminate subprocesses

oliver-sanders:
* We use a subprocess pool for running backgorund commands (e.g. `cylc clean`).
* This creates subprocesses to run our commands in, however, it does not kill the subprocess once the command has completed, it keeps the subprocess open for future reuse (more efficient than creating and destroying them every time).
* On shutdown we need to close the pool to mop up these subprocesses (this doesn't happen automatically).

From experimentation, it would appear that these subprocess do die naturally when run in single-user mode (`cylc gui`), however, become orphan processes when run in muilti-user mode (`cylc hub`, tested with `sudospawner`).

Unfortunately, this does not clean up the dangling subprocesses in multi-user mode due to https://github.com/cylc/cylc-uiserver/issues/392

**Check List**

### DIFF
--- a/changes.d/619.fix.md
+++ b/changes.d/619.fix.md
@@ -1,0 +1,1 @@
+Ensure that subprocesses created by Cylc UI Server are cleaned up correctly when the server shuts down.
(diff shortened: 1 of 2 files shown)


oliver-sanders:
CI issues fixed in https://github.com/cylc/cylc-uiserver/pull/620

MetRonnie:
Kicking tests

MetRonnie (approved):
LGTM, haven't tested

wxtim (approved):
Tried it out, seems OK

PR #620, opened 2024-08-08
setup: types-pkg-resources -> types-setuptools

oliver-sanders:
Dev installation is currently broken due to a "yanked" project ?!

Follow the instructions on this page:

https://pypi.org/project/types-pkg-resources/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -103,7 +103,7 @@ tests =
     # https://github.com/pytest-dev/pytest/issues/12263
     pytest>=6,<8.2
     towncrier>=23
-    types-pkg_resources>=0.1.2
+    types-setuptools
     types-requests>2
 all =
     %(hub)s


MetRonnie (approved):
Actually I'm not sure we need this dependency at all (mypy passes without it)

setup.cfg
```diff
@@ -103,7 +103,7 @@ tests =
     # https://github.com/pytest-dev/pytest/issues/12263
     pytest>=6,<8.2
     towncrier>=23
-    types-pkg_resources>=0.1.2
+    types-setuptools
```
oliver-sanders, line 106: > mypy passes without it

Not sure if that means removing the dep removes (some) type checking?

```suggestion
```

?

MetRonnie: Not sure, I'll just merge this as-is

Issue #643, opened 2024-11-07
cat-log: clean up processes on websocket disconnect


oliver-sanders:
I suspect that `cylc cat-log` processes created to satisfy the `LogData` subscription, are not killed if the websocket is closed.

> (investigation) Requires confirmation.

Ideally, we would detect the websocket close event, and kill the subprocess.

If this is not possible, a bandage might be to apply a (generous) timeout to the `cylc cat-log` process to ensure these processes are cleaned up eventually. This might actually be a good idea in general provided the default timeout is long enough not to cause user freustration.

PR #645, opened 2024-11-11
cat-log: add timeout to processes

oliver-sanders:
* Partially addresses #643
* Add a timeout to `cylc cat-log` processes:
  * Ensures we don't accumulate them by accident.
  * Improves efficiency by closing long-lived log streams (the user probably isn't looking at the log file any more and the log file has probably stopped growing anyway).
* At present, `cylc cat-log` processes may be left behind if the websocket is closed abruptly. This has been observed as the result of a proxy timeout.

**Check List**

### DIFF
--- a/changes.d/645.fix.md
+++ b/changes.d/645.fix.md
@@ -1,0 +1,2 @@
+Add a default timeout for the `cylc cat-log` command which is used to provide access to log files in the cylc-ui.
+This timeout can be adjusted with the `log_timeout` option.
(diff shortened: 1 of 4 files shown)


hjoliver:
(gah, the trouble orphaned cat-log processes have caused over the years!)

wxtim (approved):
Reasonable. Thank you for all the extra annotation. Some of this code is NOT self documenting IMO.

One Spelling error

PR #647, opened 2024-11-28
GH Actions: downgrade MacOS runner to 12 to retain Python 3.9 testing

MetRonnie:
https://github.com/actions/runner-images/issues/10812

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
         os: ['ubuntu-latest']
         python: ['3.8', '3.9']
         include:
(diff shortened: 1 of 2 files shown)


oliver-sanders (approved):

MetRonnie:
Agh, we will have to switch to using `setup-miniconda` anyway: https://github.com/actions/runner-images/issues/10721

> GitHub Actions is starting the deprecation process for `macOS 12`. While the image is being deprecated, You may experience longer queue times during peak usage hours. Deprecation will begin on 10/7/24 and the image will be fully unsupported by **12/3/24**

PR #650, opened 2024-12-03
GH Actions: upgrade MacOS runner back to latest

MetRonnie:
Reverts https://github.com/cylc/cylc-uiserver/pull/647

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
         os: ['ubuntu-latest']
         python: ['3.8', '3.9']
         include:
-          - os: 'macos-12'
+          - os: 'macos-latest'
             python: '3.9'
     env:
       PYTEST_ADDOPTS: --cov --color=yes
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,6 +34,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Patch DNS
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -34,6 +34,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Patch DNS
+        uses: cylc/release-actions/patch-dns@v1
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python }}
```
MetRonnie, line 43: This was not actually failing because Python 3.9 was removed; it would install it anyway. But the installation was failing due to
```
chmod: changing permissions of 'python3.9': Operation not permitted
```
([example](https://github.com/cylc/cylc-uiserver/actions/runs/12067460876/job/33650461146#step:5:26))

For some reason this is fixed by moving to before the brew install step

oliver-sanders: > For some reason this is fixed by moving to before the brew install step

That may suggest that the Brew Python is replacing the system Python?

Might be worth testing `python --version` at some point after this step to make sure we're actually getting the version of Python we are configuring?

MetRonnie: Pytest is reporting 3.9: https://github.com/cylc/cylc-uiserver/actions/runs/12146399589/job/33870231037?pr=650#step:11:14

oliver-sanders (approved):

PR #657, opened 2025-01-09
offline data: enable querying failed jobs

oliver-sanders:
Partially addresses https://github.com/cylc/cylc-uiserver/issues/440

* Allow job queries to be performed on jobs which have failed or have not run.
* Implemented the `ids`, `exids`, `states`, `exstates` fields for offline job queries (partially addresses #440).
* Fix the `state` field which previously returned `submit_status`.

Example query:

```graphql
query {
  jobs(live: false, workflows: ["*"], ids: ["*/foo"]) {
    id
    state
  }
}
```

Notes:
* [ ] The analysis view will need to be updated to use the `states` argument to filter for succeeded tasks (previously this was implied).
* [ ] If there are any queries that use the `state` field, these will need to be updated now that it returns the job state rather than the run_status.

**Check List**

### DIFF
--- a/cylc/uiserver/schema.py
+++ b/cylc/uiserver/schema.py
@@ -20,7 +20,7 @@ extra functionality specific to the UIS.
 """
 
 from functools import partial
-from typing import TYPE_CHECKING, Any, List, Optional
+from typing import TYPE_CHECKING, Any, List, Optional, Tuple
 
 import graphene
 from graphene.types.generic import GenericScalar
@@ -47,7 +47,16 @@ from cylc.flow.network.schema import (
     _mut_field,
     get_nodes_all
 )
+from cylc.flow.task_state import (
+    TASK_STATUS_FAILED,
+    TASK_STATUS_RUNNING,
+    TASK_STATUS_SUBMITTED,
+    TASK_STATUS_SUBMIT_FAILED,
+    TASK_STATUS_SUCCEEDED,
+    TASK_STATUS_WAITING,
+)
 from cylc.flow.util import sstrip
+
 from cylc.uiserver.resolvers import (
     Resolvers,
     list_log_files,
@@ -264,7 +273,7 @@ class Scan(graphene.Mutation):
     result = GenericScalar()
 
 
-async def get_elements(root, info, **kwargs):
+async def get_elements(query_type, root, info, **kwargs):
     if kwargs['live']:
         return await get_nodes_all(root, info, **kwargs)
 
@@ -287,10 +296,10 @@ async def get_elements(root, info, **kwargs):
     kwargs['exworkflows'] = [
         Tokens(w_id) for w_id in kwargs['exworkflows']]
 
-    return await list_elements(kwargs)
+    return await list_elements(query_type, kwargs)
 
 
-async def list_elements(args):
+async def list_elements(query_type, args):
     if not args['workflows']:
         raise Exception('At least one workflow must be provided.')
     elements = []
@@ -302,9 +311,18 @@ async def list_elements(args):
         )
         with CylcWorkflowDAO(db_file, is_public=True) as dao:
             conn = dao.connect()
-            if 'tasks' in args:
+            if query_type == 'jobs':
                 elements.extend(
-                    run_jobs_query(conn, workflow, args.get('tasks')))
+                    run_jobs_query(
+                        conn,
+                        workflow,
+                        ids=args.get('ids'),
+                        exids=args.get('exids'),
+                        states=args.get('states'),
+                        exstates=args.get('exstates'),
+                        tasks=args.get('tasks'),
+                    )
+                )
             else:
                 elements.extend(run_task_query(conn, workflow))
     return elements
@@ -321,6 +339,7 @@ SELECT
     cycle,
     submit_num,
     submit_status,
+    run_status,
     time_run,
     time_run_exit,
     job_id,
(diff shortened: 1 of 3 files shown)


oliver-sanders:
(sorting out some SQL issues and getting some tests arranged)

cylc/uiserver/schema.py
```diff
@@ -396,78 +415,299 @@ def run_task_query(conn, workflow):
...
+            (note this is interpreted as a bool, the exact value is ignored).
+
+    Returns:
+        status: The job status, e.g "submitted" or "running".
+
+    Examples:
+        >>> _state_to_status(0, None, True)
+        'running'
+        >>> _state_to_status(None, None, None)
+        'waiting'
```
wxtim, line 525: IMO You should test all possibilites: This logic gives me a headache (Submit status only makes sense when it's an exit code, but 1 is truthy in python). I think that full tests are good for documentation:

```suggestion
        >>> _state_to_status(None, None, None)
        'waiting'
        >>> _state_to_status(1, None, None)
        'submit-failed'
        >>> _state_to_status(0, None, True)
        'running'
        >>> _state_to_status(0, 0, True)
        'succeeded'
        >>> _state_to_status(0, 1, True)
        'failed'
```

cylc/uiserver/schema.py
```diff
@@ -396,78 +415,299 @@ def run_task_query(conn, workflow):
...
+            status = TASK_STATUS_FAILED
+    elif time_run is not None:
+        status = TASK_STATUS_RUNNING
+    elif submit_status is not None:
+        if submit_status == 0:
+            status = TASK_STATUS_SUBMITTED
+        else:
+            status = TASK_STATUS_SUBMIT_FAILED
+    else:
+        status = TASK_STATUS_WAITING
```
wxtim, line 544:
```suggestion
    for status, state in _JOB_STATUS_TO_STATE.items():
        if state == (submit_status, run_status, time_run):
            return status
    else:
        # I'm not sure about this, but it's a possibility.
        raise ValueError(
            f"Invalid state: {submit_status}, {run_status}, {time_run}"
        )
```
The else is a matter of opinion. But this seems like less data-as-logic.

cylc/uiserver/schema.py
oliver-sanders: > IMO You should test all possibilites

If I did that I would just be replicating the dictionary above, essentially:

```
x  = {
   'a': 1,
    'b': 2,
   'c': 3
}

"""
>>> x['a']
1
>>> x['b']
2
>>> x['c']
3
"""
```

Which would only really be testing that the dictionary matches the test, not that the code is actually doing the right thing.

cylc/uiserver/schema.py
oliver-sanders: I'm not sure this logic (pre-existing) is 100% the reverse of the dictionary I've added (you can pass improbable combinations of options into this function).

cylc/uiserver/schema.py
```diff
@@ -396,78 +415,299 @@ def run_task_query(conn, workflow):
...
+                    if token == 'job':
+                        value = int(value)
+                    item.append(rf'{column} GLOB ?')
+                    where_args.append(value)
+            items.append(r'(' + ' AND '.join(item) + r')')
+
+        if items:
+            where_stmts.append(
+                r'(' + ' OR '.join(items) + ')'
+            )
```
wxtim, line 598: Possibly simpler?
```suggestion
    # filter by cycle/task/job ID
    id_items = []
    for id_ in ids or []:
        item = []
        for token, column in (
            ('cycle', 'cycle'),
            ('task', 'name'),
            ('job', 'submit_num'),
        ):
            value = id_[token]
            if value:
                if token == 'job':
                    value = int(value)
                item.append(rf'{column} GLOB ?')
                where_args.append(value)
        id_items.append(r'(' + ' AND '.join(item) + r')')
    else:
        if id_items:
            where_stmts.append(r'(' + ' OR '.join(id_items) + r')')
```

cylc/uiserver/schema.py
```diff
@@ -396,78 +415,299 @@ def run_task_query(conn, workflow):
...
+                ('task', 'name'),
+                ('job', 'submit_num'),
+            ):
+                value = id_[token]
+                if value:
+                    if token == 'job':
+                        value = int(value)
+                    item.append(rf'{column} GLOB ?')
+                    where_args.append(value)
+            items.append(r'(' + ' AND '.join(item) + r')')
```
wxtim, line 593: F-string might be tidier?

```suggestion
        id_items.append(rf'({" AND ".join(item)})')
```

cylc/uiserver/schema.py
```diff
@@ -396,78 +415,299 @@ def run_task_query(conn, workflow):
...
-            'finished_time': row[5],
-            'job_ID': row[6],
-            'platform': row[7],
-            'submitted_time': row[8],
+            'state': _state_to_status(row[3], row[4], row[5]),
+            'started_time': row[5],
+            'finished_time': row[6],
+            'job_ID': row[7],
+            'platform': row[8],
+            'submitted_time': row[9],
```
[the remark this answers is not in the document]
oliver-sanders: Reasonable, wasn't really planning on refactoring.

cylc/uiserver/schema.py
oliver-sanders: I'd prefer not to declare the `id_items` list in situations where we don't need it (not filtering is more common than filtering), but don't really care.

cylc/uiserver/schema.py
```diff
@@ -396,78 +415,299 @@ def run_task_query(conn, workflow):
...
+
+    # filter by job state
+    if states:
+        items = []
+        for status in states:
+            submit_status, run_status, time_run = _status_to_state(status)
+            if submit_status is None:
+                # hasn't yet submitted (i.e. there is no job)
+                continue
+            item = [r'IFNULL(submit_status,999) = ?']
```
oliver-sanders, line 626: FYI: SQL skips comparisons for Null values, so we need to use `IFNULL` to force the comparison to be respected.

cylc/uiserver/schema.py
wxtim: No:
* `submit_status` and `run_status` can be `None`, `0` or any positive integer - `>0` is probably the correct test.
* `time_run` is a string(?) or None.

wxtim: Happy to leave pre-existing function alone.

cylc/uiserver/schema.py
wxtim: Not a big deal. Don't _really_ care that much.

wxtim (approved):
Looks good - made a reasonable effort to break it.

cylc/uiserver/schema.py
```diff
@@ -396,78 +415,299 @@ def run_task_query(conn, workflow):
...
+                ('task', 'name'),
+                ('job', 'submit_num'),
+            ):
+                value = id_[token]
+                if value:
+                    if token == 'job':
+                        value = int(value)
+                    item.append(rf'{column} GLOB ?')
+                    where_args.append(value)
+            id_items.append(rf'({" AND ".join(item)})')
```
markgrahamdawson, line 593: the tests are failing because of this... I cant see it anywhere else in the schema.py ?

cylc/uiserver/schema.py
MetRonnie: > submit_status and can be 0 or any positive integer

No it won't be. These are loaded from the DB and will always be `1 | 0 | None`

cylc/uiserver/schema.py
oliver-sanders: Ach, I applied a code suggestion that didn't work.

markgrahamdawson (approved):

cylc/uiserver/schema.py
MetRonnie: This is duplicating https://github.com/cylc/cylc-flow/blob/335216488793401ce3f7f90ebf07c11d3187b63c/cylc/flow/data_store_mgr.py#L1655-L1668

can we centralise this logic?

oliver-sanders: > can we centralise this logic?

Sure, but not in a single PR.

wxtim: @oliver-sanders Please can you make a new issue based on this thread.

MetRonnie (approved):

PR #613, opened 2024-07-04
Auth tidy

oliver-sanders:
(low priority)

Looking into the feasibility of making a change to the auth code.

Before changing anything around, I wanted to get the coverage up.

Overview:
* Blacken code.
* Simplify code.
* Fill some test holes.

Full details in the individual commit messages (easier to review changes one by one).

As always with auth code changes (even ones like this which shouldn't change any functionality), please extensively test this (note you can use exvcylc17 for this)!

Plz squash merge.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #660, opened 2025-01-13
Fix AttributeError: 'Authorization' object has no attribute 'log'

MetRonnie:
Edit: fixes a bug introduced in #613 

**Check List**

### DIFF
--- a/cylc/uiserver/authorise.py
+++ b/cylc/uiserver/authorise.py
@@ -165,12 +165,12 @@ class Authorization:
         log,
     ):
         self.owner_user_name: str = owner_user_name
-        self.owner_user_groups: List[str] = self._get_groups(
-            self.owner_user_name
-        )
         self.log = log
         self.owner_auth_conf: dict = owner_auth_conf
         self.site_auth_config: dict = site_auth_conf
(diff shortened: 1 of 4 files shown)


wxtim:
Please can you paste what you did to demonstrate the bug?

cylc/uiserver/authorise.py
```diff
@@ -165,12 +165,12 @@ def __init__(
...
         self.owner_user_name: str = owner_user_name
-        self.owner_user_groups: List[str] = self._get_groups(
-            self.owner_user_name
-        )
         self.log = log
         self.owner_auth_conf: dict = owner_auth_conf
         self.site_auth_config: dict = site_auth_conf
+        self.owner_user_groups: List[str] = self._get_groups(
+            self.owner_user_name
+        )
```
MetRonnie, line 173: I don't know how to reproduce it, but the cause should be fairly self-evident from looking at `self._get_groups()`:
https://github.com/cylc/cylc-uiserver/blob/ef4a4d1ec613dec6aba6e6446f530fbf285659d4/cylc/uiserver/authorise.py#L487-L492

Can't use `self.log` before it's been defined

wxtim (approved):
👍🏼

PR #672, opened 2025-03-10
graphql-core v3 upgrade

dwsutherland:
partially addresses https://github.com/cylc/cylc-uiserver/issues/333

Is a sibling to:
https://github.com/cylc/cylc-flow/pull/6478

(cylc-flow/cylc-uiserver PR needed to be merged at the same time)

Dependencies on:

`graphene-tornado==2.6.*`
`graphql-ws==0.4.4`

have been removed and/or rewritten as a Cylc port of active projects.

graphql-core-v3

**Check List**


cylc/uiserver/graphql/subscribe.py
```diff
@@ -0,0 +1,251 @@
+# Taken from:
+# https://github.com/graphql-python/graphql-core/blob/v3.2.6/src/graphql
+# /execution/subscribe.py
+# Modified to include `execution_context_class` and `middleware` in
+# the execution of GraphQL subscriptions.
+# This should not be necessary with some unspecified future releases as
+# the head of graphql-core has these included.
```
oliver-sanders, line 7: What release should this arrive in?

cylc/uiserver/graphql/tornado.py
```diff
@@ -0,0 +1,529 @@
+# This is a Cylc port of the graphene-django graphene integration:
+# https://github.com/graphql-python/graphene-django
```
oliver-sanders, line 2: TODO: Clear licencing implications.

cylc/uiserver/graphql/tornado_ws.py
```diff
@@ -0,0 +1,381 @@
+# This Cylc solution for graphql subscriptions over websockets is essentially
+# a condense from of the stagnant graphql-ws project:
+# https://github.com/graphql-python/graphql-ws
+# and our implementation of the tornado integration PR that never got merged:
+# https://github.com/graphql-python/graphql-ws/pull/25/files
```
oliver-sanders, line 5: TODO: Clear licencing implications.

cylc/uiserver/graphql/tornado_ws.py
```diff
@@ -0,0 +1,381 @@
...
+    instantiate_middleware,
+)
+
+from cylc.uiserver.authorise import AuthorizationMiddleware
+from cylc.uiserver.graphql.subscribe import subscribe
+
+
+NO_MSG_DELAY = 1.0
+
+GRAPHQL_WS = "graphql-ws"
```
oliver-sanders, line 35: @oliver-sanders please check the websocket subprotocol.

cylc/uiserver/handlers.py
```diff
@@ -314,32 +313,39 @@ class UIServerGraphQLHandler(CylcAppHandler, TornadoGraphQLHandler):
...
+    def initialize(
+        self, schema=None, middleware=None, root_value=None, graphiql=False,
+        pretty=False, batch=False, execution_context_class=None,
+        validation_rules=None, auth=None, subscription_path=None, **kwargs
+    ):
         super(TornadoGraphQLHandler, self).initialize()
-        self.auth = auth
-        self.schema = schema
 
+        self.schema = schema or self.schema
```
oliver-sanders, line 323: `or self.schema`?

(also seen in other variables below)

Is this to handle `initialize` being called on the same handler multiple times with different arguments?

oliver-sanders (commented):
:rocket:

cylc/uiserver/graphql/__init__.py
oliver-sanders, line 1: Note to all, this work involved porting the authorisation system to the new GraphQL framework.

We need to perform a security review to ensure this is watertight with the changes.

cylc/uiserver/app.py
```diff
@@ -507,11 +507,12 @@ def initialize_handlers(self):
                 {
                     'schema': schema,
                     'resolvers': self.resolvers,
-                    'backend': CylcGraphQLBackend(),
                     'middleware': [
                         AuthorizationMiddleware,
                         IgnoreFieldMiddleware
                     ],
+                    'execution_context_class': CylcExecutionContext,
+                    'subscription_path': 'cylc/subscriptions',
```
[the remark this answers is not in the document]
dwsutherland: We're not serving up a separate GraphiQL from the UIS.. So I can remove this.. Will comment below.

cylc/uiserver/graphql/render_graphiql.py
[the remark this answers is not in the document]
dwsutherland: Actually, we could get rid of this if we like.. Because I noticed this too..

But as I was doing a port of `graphene-django` I thought I would port a version of `graphql-server`s template also..

Thoughts? should we only port what we need? probably a good idea, it would mean less to maintain..

cylc/uiserver/graphql/subscribe.py
dwsutherland: Well `graphql-core` HEAD has the removed this module and included subscriptions as part of the execution/execute.. Which also includes using middleware/execution context which we need:
https://github.com/graphql-python/graphql-core/blob/416247c1d511350445c23096f9491fbef424b69b/src/graphql/execution/execute.py#L2114

So we need a release of `graphql-core` which includes this, and a corresponding release of `graphene` compatible with this.
Will create and issue

cylc/uiserver/handlers.py
dwsutherland: I think, if not set, it will pick up the defaults set by initialized `TornadoGraphQLHandler` parent..

Tornado spins up a new handler for each request I think.. But I think the options are set in the app on UIS startup..

cylc/uiserver/graphql/render_graphiql.py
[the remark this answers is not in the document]
dwsutherland: Actually I think I might remove ..
I feel like this option would be only useful if you developed the an in isolation of a UI version.. Then you could just spin up a GraphiQL end point to test.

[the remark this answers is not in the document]
dwsutherland: Removed

cylc/uiserver/app.py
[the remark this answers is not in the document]
dwsutherland: GraphiQL removed..

cylc/uiserver/graphql/subscribe.py
```diff
@@ -0,0 +1,251 @@
+# Taken from:
+# https://github.com/graphql-python/graphql-core/blob/v3.2.6/src/graphql
```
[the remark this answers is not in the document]
dwsutherland: I might actually put this on cylc-flow end.. In case we have graphql subs on the `cylc-flow` end before a new `graphql-core`/`graphene` is released....

cylc/uiserver/graphql/render_graphiql.py
```diff
@@ -0,0 +1,310 @@
+# Taken from (graphql-server)[https://github.com/graphql-python/graphql-server/blob/master/graphql_server/render_graphiql.py]
```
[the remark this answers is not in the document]
oliver-sanders: Note to self: Make sure this code is rebase-squashed out of the repo to save going through legal wrangling.

cylc/uiserver/graphql/subscribe.py
oliver-sanders: > What release should this arrive in?

Taking a look at the repo, I think this will arrive in v 3.3.0. They have already published alpha releases of this.

> Well graphql-core HEAD has the removed this subscribe module and included subscriptions as part of the execution/execute

It looks like that change [was made in 2022](https://github.com/graphql-python/graphql-core/commit/689a70d052a58277d20cc6b6348b0eb31ad3a113), but hasn't made it into a stable release yet?!

These v3.3 alpha releasees go back three years! Yikes!

> So we need a release of graphql-core which includes this, and a corresponding release of graphene compatible with this.

Taking a look at graphene, they have pinned to `graphql-core>=3.1,<3.3`, so it looks like they will indeed need a new release of Graphene to support a new release of graphql-core.

Unfortunately, this is a problem for us as long as we need to support Python 3.7 (as any new release will not support this).

oliver-sanders: I think:

* We need to host this code in cylc-uiserver (as you have done).
* We need to pin to graphql-core v3.2.
* We can look at migrating once they have made v3.3.0 release and we no longer require Python 3.7 support.

cylc/uiserver/graphql/render_graphiql.py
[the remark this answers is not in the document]
dwsutherland: rebase squash done.

cylc/uiserver/graphql/subscribe.py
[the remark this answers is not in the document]
dwsutherland: removed, and rebase squashed..

cylc/uiserver/handlers.py
oliver-sanders: > Tornado spins up a new handler for each request I think

Yes. We rely on this for security reasons which is why this caught my interest.

Will have a quick check to make sure.

cylc/uiserver/graphql/subscribe.py
dwsutherland: moved to `cylc-flow`  (because we may need to use it there)

cylc/uiserver/graphql/tornado_ws.py
dwsutherland: The modifications are reasonably quite large, but yes, still need to clear..

oliver-sanders:
@dwsutherland, please could you rebase this onto master (pick up recent GH actions changes) and edit this line:

https://github.com/cylc/cylc-uiserver/blob/4e7c9a6454a201190437953449da0f9ed72b2669/.github/workflows/test.yml#L30

To add `3` to the list to unlock testing with the latest Python versions (which these dependency changes will allow!).

MetRonnie (commented):
https://github.com/dwsutherland/cylc-uiserver/pull/4

cylc/uiserver/graphql/tornado_ws.py
MetRonnie: What is needed to be done in order to switch to the `graphql-transport-ws` subprotocol? (which I think is described here: https://github.com/graphql/graphql-over-http/blob/main/rfcs/GraphQLOverWebSocket.md)

Switching to `graphql-transport-ws` at some point is needed to address https://github.com/cylc/cylc-ui/issues/1028

MetRonnie:
> partially addresses #333

I think it closes it?

cylc/uiserver/graphql/tornado_ws.py
```diff
@@ -0,0 +1,383 @@
...
+from cylc.uiserver.authorise import AuthorizationMiddleware
+from cylc.uiserver.schema import SUB_RESOLVER_MAPPING
+
+
+NO_MSG_DELAY = 1.0
+
+GRAPHQL_WS = "graphql-ws"
+WS_PROTOCOL = GRAPHQL_WS
+GQL_DATA = "data"  # Server -> Client
+GQL_COMPLETE = "complete"  # Server -> Client
```
MetRonnie, line 39: Duplicated below
```suggestion
```

cylc/uiserver/graphql/tornado_ws.py
MetRonnie: Actually, I was curious and had a little dig into it. Seems like this module can be lightly modified to support the `graphql-transport-ws` subprotocol, although ideally we would go with a pre-rolled solution if one is available?

dwsutherland:
> > partially addresses #333
> 
> I think it closes it?

Well, this in conjunction with it's `cylc-flow`/`cylc-ui` siblings

cylc/uiserver/graphql/tornado_ws.py
dwsutherland: Well, the communication structure already fits the definition (aside from close events maybe?), with channels and ids etc..

oliver-sanders: > ideally we would go with a pre-rolled solution if one is available?

There is no off-the-shelf solution for websocket support in Tornado, hence this module.

The diff required to support the graphql-transport-ws subprotocol is fairly minimal. There's an (unmerged) example in the abandoned graphql-ws library: https://github.com/graphql-python/graphql-ws/pull/65

I would suggest sticking with the current subprotocol (as David has implemented) for this graphql-core v3 upgrade. We can tackle the subprotocol change later, no need to do it now.

cylc/uiserver/graphql/tornado.py
```diff
@@ -0,0 +1,449 @@
...
+# with reference to:
+# https://github.com/graphql-python/graphene-tornado
+# https://github.com/graphql-python/graphql-server
+#
+# Excludes GraphiQL
```
oliver-sanders, line 7: Comply with MIT licence requirements:

```suggestion
# The MIT License (MIT)
#
# Copyright (c) 2016-Present Syrus Akbary
#
# Permission is hereby granted, free of charge, to any person obtaining a copy
# of this software and associated documentation files (the "Software"), to deal
# in the Software without restriction, including without limitation the rights
# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
# copies of the Software, and to permit persons to whom the Software is
# furnished to do so, subject to the following conditions:
#
# The above copyright notice and this permission notice shall be included in all
# copies or substantial portions of the Software.
#
# ----------------------------------------------------------------------------
#
# This is a Cylc port of the graphene-django graphene integration:
# https://github.com/graphql-python/graphene-django
# with reference to:
# https://github.com/graphql-python/graphene-tornado
# https://github.com/graphql-python/graphql-server
#
# Excludes GraphiQL
```

cylc/uiserver/graphql/tornado.py
oliver-sanders, line 1: Please add "foreign" code (that we plan to remove in due course) to:

* The "ignore" list in .codecov.yml.
* The "exclude" list in tox.ini.

This will stop CI from complaining about these files.

cylc/uiserver/graphql/tornado_ws.py
oliver-sanders, line 1: Please add "foreign" code (that we plan to remove in due course) to:

* The "ignore" list in .codecov.yml.
* The "exclude" list in tox.ini.

This will stop CI from complaining about these files.

oliver-sanders:
FYI: You accidentally [force-pushed](https://github.com/cylc/cylc-uiserver/compare/5ea2845516fb665f2ac098ec014425879612795d..ac0589ea899d89efb1bc31ec8523c1c2ad9f2370) over some suggestions so will need to re-apply them.

oliver-sanders:
I've been getting this traceback shortly after UIS statrup:

```
Task exception was never retrieved
future: <Task finished name='Task-36' coro=<TornadoSubscriptionServer._process_message() done, defined at ~/cylc-uiserver/cylc/uiserver/graphql/tornado_ws.py:173> exception=UnboundLocalError("local variable 'iterator' referenced before assignment")>
  File "~/cylc-uiserver/cylc/uiserver/graphql/tornado_ws.py", line 189, in _process_message
    return await self.on_start(connection_context, op_id, params)
  File "~/cylc-uiserver/cylc/uiserver/graphql/tornado_ws.py", line 308, in on_start
    await iterator.aclose()
UnboundLocalError: local variable 'iterator' referenced before assignment
```

The server stays up, but this does seem to be a code error, I've worked around it like so:

```diff
diff --git a/cylc/uiserver/graphql/tornado_ws.py b/cylc/uiserver/graphql/tornado_ws.py
index e42052f..5916d89 100644
--- a/cylc/uiserver/graphql/tornado_ws.py
+++ b/cylc/uiserver/graphql/tornado_ws.py
@@ -286,6 +286,7 @@ class TornadoSubscriptionServer:
 
         params['kwargs']['root_value'] = op_id
         execution_result = await self.execute(params)
+        iterator = None
         try:
             if isawaitable(execution_result):
                 execution_result = await execution_result
@@ -305,7 +306,8 @@ class TornadoSubscriptionServer:
         except Exception as e:
             await self.send_error(connection_context, op_id, e)
         finally:
-            await iterator.aclose()
+            if iterator:
+                await iterator.aclose()
         await self.send_message(connection_context, op_id, GQL_COMPLETE)
         await connection_context.unsubscribe(op_id)
         await self.on_operation_complete(connection_context, op_id)
```

cylc/uiserver/graphql/tornado_ws.py
oliver-sanders: Have checked the subprotocol, it appears to be producing identical results and is following the correct subprotocol for the current UI implmentation.

cylc/uiserver/resolvers.py
```diff
@@ -332,7 +335,7 @@ async def play(
                     results[wflow] = 'started'
 
             except Exception as exc:
-                # oh noes, something went wrong, send back confirmation
+                log.exception(exc)
```
MetRonnie, line 338: Ditto (almost identical change made in #689)

MetRonnie (changes requested):

cylc/uiserver/graphql/tornado.py
dwsutherland: We're keeping this one. (the only way we remove is if a non-dead integration is created)
I cannibolised a few different implementations to create this one..

cylc/uiserver/graphql/tornado_ws.py
dwsutherland: Again, keeping until better alternative is available.

cylc/uiserver/graphql/tornado.py
dwsutherland: I'm not sure the line with which code becomes our own.. it's probably 1/3 graphene django 1/3 graphene-tornado 1/3 ours (maybe a little inspired by graphql-server)..

cylc/uiserver/resolvers.py
```diff
@@ -257,8 +260,8 @@ async def clean(
             elif isinstance(exc, WorkflowFilesError):  # Expected error
                 msg = str(exc)
             else:  # Unexpected error
+                log.exception(exc)
                 msg = f"{type(exc).__name__}: {exc}"
-                log.exception(msg)
```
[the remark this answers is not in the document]
dwsutherland: Odd, I seem to have got these changes from you  😄 :
https://github.com/dwsutherland/cylc-uiserver/pull/4

resolved now

cylc/uiserver/resolvers.py
dwsutherland: resolved

dwsutherland:
> FYI: You accidentally [force-pushed](https://github.com/cylc/cylc-uiserver/compare/5ea2845516fb665f2ac098ec014425879612795d..ac0589ea899d89efb1bc31ec8523c1c2ad9f2370) over some suggestions so will need to re-apply them.

No idea how that happened, must have push from an old version? .. fixed now

dwsutherland:
> This does seem to be a code error, I've worked around it like so:

Applied the same workaround.. From looking at the code, this could happen when the execution result is an error or something not async iterator.. Not sure what actions this translates to for the user, you may be correct.

cylc/uiserver/graphql/tornado.py
```diff
@@ -0,0 +1,465 @@
...
+#
+# ----------------------------------------------------------------------------
+#
+# This is a Cylc port of the graphene-django graphene integration:
+# https://github.com/graphql-python/graphene-django
+# with reference to:
+# https://github.com/graphql-python/graphene-tornado
+# https://github.com/graphql-python/graphql-server
+#
+# Excludes GraphiQL
```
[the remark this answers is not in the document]
dpmatthews: Looks fine

cylc/uiserver/graphql/tornado_ws.py
```diff
@@ -0,0 +1,399 @@
...
+# ----------------------------------------------------------------------------
+#
+# This code was derived from coded in the graphql-ws project, and code that
+# was offered to the graphql-ws project by members of the Cylc
+# development team but not merged.
+#
+# * https://github.com/graphql-python/graphql-ws
+# * https://github.com/graphql-python/graphql-ws/pull/25/files
+#
+# It has been evolved to suit and ported to graphql-core v3.
```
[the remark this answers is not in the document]
dpmatthews: Looks fine

oliver-sanders (commented):
This code contains instance attributes which shadow class attributes which is a tad confusing.

#### Declaring attributes

This pattern declares `Foo.variable`, then shadows it with `self.variable`:

```python
class Foo():

    variable = None  # Foo.variable (class variable)

    def (self, variable):
        self.variable = variable # Foo().variable (instance variable)
```

We only need one variable in this case, so the example can be re-written like so:

```diff
 class Foo():

-    variable = None
-
     def (self, variable):
         self.variable = variable
```

#### Declaring defaults for attributes

```python
class Foo():

    variable = None  # class attribute - only used once

    def (self, variable):
        self.variable = variable or self.variable  # class attribute used here only
```

This should be written like this (remove class variable):

```diff
 class Foo():

-    variable = None
-
     def (self, variable):
-        self.variable = variable or self.variable
+        self.variable = variable or None
```

Or this (avoid shadowing class variable):

```diff
 class Foo():

-    variable = None
+    VARIABLE = None  # constant - do not change

     def (self, variable):
-        self.variable = variable or self.variable
+        self.variable = variable or self.VARIABLE
```

Or this (hardcode defaults in method signature):

```diff
 class Foo():

-    variable = None
-
-    def (self, variable):
+    def (self, variable=None):
-        self.variable = self.variable
+        self.variable = variable
```

#### Declaring types

This sets a class attribute that is not used:

```python
class Foo():

    variable: Optional[str] = None

    def (self, variable: str):
        self.variable = variable
```

But we can still declare the type without assigning

```diff
 class Foo():

-    variable: Optional[str] = None
+    variable: str

    def (self, variable: str):
        self.variable = variable
```

cylc/uiserver/handlers.py
```diff
@@ -322,32 +315,37 @@ class UIServerGraphQLHandler(CylcAppHandler, TornadoGraphQLHandler):
...
+                self.middleware = middleware
+            else:
+                self.middleware = list(instantiate_middleware(middleware))
+        self.root_value = root_value
+        self.pretty = pretty or self.pretty
+        self.batch = batch or self.batch
+        self.execution_context_class = (
+            execution_context_class or self.execution_context_class
+        )
+
```
oliver-sanders, line 338: With the exception of the `self.auth = auth` line, this duplicates `TornadoGraphQLHandler.initialize` so is overwriting it?!

I think we can replace this with:

```suggestion
        super(TornadoGraphQLHandler, self).initialize(
            schema=None,
            middleware=None,
            root_value=None,
            pretty=False,
            batch=False,
            execution_context_class=None,
            validation_rules=None,
        )
        self.auth = auth
```

oliver-sanders:
@dwsutherland, raised a PR with the requested changes above: https://github.com/dwsutherland/cylc-uiserver/pull/5

dwsutherland:
> @dwsutherland, raised a PR with the requested changes above: [dwsutherland#5](https://github.com/dwsutherland/cylc-uiserver/pull/5)

Thanks, merged

oliver-sanders (changes requested):
Code looks good and performs as intended for single-user cases.

But, unfortunately, it has failed security testing as the authorisation layer is no longer being engaged with these changes :(

I think the `AuthorizationMiddleware` needs to be updated to reflect changes in the `info` object. At present, it looks like the [`if len(info.path) > 1: return next_(root, info, **args)`](https://github.com/dwsutherland/cylc-uiserver/blob/cb90911d30c26620c2835b4380461d17ac629cbe/cylc/uiserver/authorise.py#L512-L513) return is always activated, which has the effect of bypassing the authorisation.

Testing this requires two user accounts, here's an example setup that grants `other-user` the permission to view your workflows, and run the `poll`, `play`, `pause` and `stop` mutations:

```python
# ~/.cylc/uiserver/jupyter_config.py

c.Authenticator.allow_all = True

c.CylcUIServer.site_authorization = {
    "*": {  # For all ui-server owners,
        "*": {  # Any authenticated user
            "limit": ["READ", "CONTROL"],
        },
    },
}

    'other-user': ['READ', 'Poll', 'Play', 'Pause', 'Stop'],
}
```

Then launch the hub:

```console
$ cylc hub
```

Next:
* Sign in as yourself and start your server.
* Get `other-user` to sign in. This will result in an error (expected user permissions error).
* Get them to navigate to your server (send them your URL).
* Get them to navigate to the GraphiQL page and issue some test mutations.

Note, at present, an issue with the Cylc UI Server means that the UI is not made aware of the whitelisted mutations (`poll`, `play`, `pause` and `stop`), so these buttons will remain greyed out in the GUI, despite `other-user` having the permissions to run these mutations, hence the need to use GraphiQL to test this at the moment. Note, this is a separate bug not related to this PR (and not a security issue).

dwsutherland:
> I think the AuthorizationMiddleware needs to be updated to reflect changes in the info object. At present, it looks like the [if len(info.path) > 1: return next_(root, info, **args)](https://github.com/dwsutherland/cylc-uiserver/blob/cb90911d30c26620c2835b4380461d17ac629cbe/cylc/uiserver/authorise.py#L512-L513) return is always activated, which has the effect of bypassing the authorisation.

I think I know the differences in info.path that caused this.. I had to deal with this in the other middleware..

hjoliver:
> Although, I haven't been about to setup permissions properly for my other-user .

What do you mean by "privileges" - just fine grained authorization of Cylc actions?

dwsutherland:
> > Although, I haven't been about to setup permissions properly for my other-user .
> 
> What do you mean by "privileges" - just fine grained authorization of Cylc actions?

Yeah, tried it this morning on my VM with.. it's picking it up, but setting to read:
```
[I 2025-05-23 14:16:16.406 CylcUIServer] Serving UI from: /home/sutherlander/cylc/cylc-ui/dist
{'other-suther': ['READ', 'Poll', 'Play', 'Pause', 'Stop']}
{'*': {'*': {'limit': ['READ', 'CONTROL']}}}
[I 2025-05-23 14:16:16.408 CylcUIServer] User other-suther authorized permissions: ['read']
{'read'}
```
just have to dig deeper as to why it's setting it to `read`

hjoliver:
Is it this: https://github.com/cylc/cylc-uiserver/issues/692

dwsutherland:
> Is it this: #692

Fixed:

https://github.com/cylc/cylc-uiserver/pull/693

MetRonnie (commented):
Just quickly skimmed through the tornado code and spotted a couple of small things

cylc/uiserver/graphql/tornado.py
```diff
@@ -0,0 +1,455 @@
...
+MUTATION_ERRORS_FLAG = "graphene_mutation_has_errors"
+MAX_VALIDATION_ERRORS = None
+
+
+def get_content_type(request: 'HTTPServerRequest') -> str:
+    content_type = request.headers.get(
+        "Content-Type",
+        request.headers.get("Content-Type", "")
+    )
+    return content_type.split(";", 1)[0].lower()
```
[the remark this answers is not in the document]
dwsutherland: Yeah, that's a weird one.. Lost in translation maybe? can't imagine someone writing this haha

cylc/uiserver/authorise.py
```diff
@@ -509,9 +510,12 @@ class AuthorizationMiddleware:
     def resolve(self, next_, root, info, **args):
         current_user = info.context["current_user"]
         # We won't be re-checking auth for return variables
-        if len(info.path) > 1:
+        if len(info.path.as_list()) > 1:
             return next_(root, info, **args)
```
MetRonnie, line 514: Might be worth adding a bit more detail to this comment, while we're here? Currently it is not very edifying to me

MetRonnie (commented):
Think I'm done with this one, will approve after cylc-flow PR 🚀

cylc/uiserver/authorise.py
oliver-sanders: Can you add a comment to explain what this line is achieving to ensure this context is remembered going forward.

dwsutherland: Ok, I've added a more detailed comment.

dwsutherland:
> Couple of changes accidentally reverted in latest force push (I've unresolved the conversations)

Yeah, I probably didn't fetch these locally

oliver-sanders (approved):
Whoop whoop, hello Python 3.10+!

dwsutherland:
Will fix any in the morning (if that's ok)

dwsutherland:
Tests are passing, however, coverage is a bit rubbish with the new code.
We need to add tests that spin up a real UIS, so I can add a subscription test(s).. (but perhaps that should be a follow-on PR?)

dwsutherland:
Tests appear to be failing because of the new change in master?
```
File "/home/runner/work/cylc-uiserver/cylc-uiserver/cylc/uiserver/__init__.py", line 20, in <module>
    from cylc.uiserver.app import CylcUIServer
  File "/home/runner/work/cylc-uiserver/cylc-uiserver/cylc/uiserver/app.py", line 90, in <module>
    from cylc.flow.network.graphql import (
  File "/home/runner/micromamba/envs/cylc-uiserver/lib/python3.8/site-packages/cylc/flow/network/graphql.py", line 41, in <module>
    from cylc.flow.network.schema import NODE_MAP
  File "/home/runner/micromamba/envs/cylc-uiserver/lib/python3.8/site-packages/cylc/flow/network/schema.py", line 636, in <module>
    class Workflow(ObjectType):
  File "/home/runner/micromamba/envs/cylc-uiserver/lib/python3.8/site-packages/graphene/types/objecttype.py", line 44, in __new__
    dataclass = make_dataclass(name_, fields, bases=())
  File "/home/runner/micromamba/envs/cylc-uiserver/lib/python3.8/dataclasses.py", line 1230, in make_dataclass
    return dataclass(cls, init=init, repr=repr, eq=eq, order=order,
  File "/home/runner/micromamba/envs/cylc-uiserver/lib/python3.8/dataclasses.py", line 1019, in dataclass
    return wrap(cls)
  File "/home/runner/micromamba/envs/cylc-uiserver/lib/python3.8/dataclasses.py", line 1011, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen)
  File "/home/runner/micromamba/envs/cylc-uiserver/lib/python3.8/dataclasses.py", line [86](https://github.com/cylc/cylc-uiserver/actions/runs/15758596335/job/44419593075?pr=672#step:9:90)1, in _process_class
    cls_fields = [_get_field(cls, name, type)
  File "/home/runner/micromamba/envs/cylc-uiserver/lib/python3.8/dataclasses.py", line 861, in <listcomp>
    cls_fields = [_get_field(cls, name, type)
  File "/home/runner/micromamba/envs/cylc-uiserver/lib/python3.8/dataclasses.py", line 745, in _get_field
    raise ValueError(f'mutable default {type(f.default)} for field '
ValueError: mutable default <class 'list'> for field log_records is not allowed: use default_factory
```

oliver-sanders:
gdammit.

Looks like [this](https://github.com/cylc/cylc-flow/pull/6755) works under graphql-core v2 but not v3?!

oliver-sanders:
This seems to do the trick: https://github.com/cylc/cylc-flow/pull/6803

cylc/uiserver/tests/test_schema.py
```diff
@@ -30,3 +31,19 @@ def test_node_map():
...
+
+    with pytest.raises(Exception) as exc:
+        await mutator(
+            None,
+            Info(),
+            command='NotACommand',
+            workflows=None,
+            args={'This': 'That'}
+        )
+    assert 'NoneType' in f'{exc.value}'
```
MetRonnie, line 49: This doesn't seem right to me. It looks like you're trying to test that `NotACommand` is handled properly? But the exception is `AttributeError: 'NoneType' object has no attribute 'log'` which is unrelated (due the line above `context = {'resolvers': None,`)

dwsutherland: Yeah, it was just a cheap way of hitting most the code without creating a resolver object.. But I suppose I could

dwsutherland:
I've been trying to get the subscription working, have got to trying with these:
```
@pytest.fixture
def gql_sub(jp_ws_fetch):
    """Perform a GraphQL request.
    E.G:
    conn = await gql_sub(
        *('cylc', 'subscriptions'),
        query='''
            subscription {
                workflows {
                    id
                    status
                }
            }
        ''',
    )
    while True:
        msg = await conn.read_message()
        # . . .
    """
    async def _fetch(*endpoint, query=None, headers=None):
        headers = headers or {}
        #headers = {
        #    **headers,
        #    'Content-Type': 'application/json'
        #}
        return await jp_ws_fetch(
            *endpoint,
            headers=headers,
            params={'query': query},
            subprotocols=['graphql-ws'],
        )
    return _fetch
```
```
async def test_subscription(gql_sub, dummy_workflow):
    """Test sending the most basic GraphQL subscription."""
    # configure two dummy workflows so we have something to look at
    await dummy_workflow('foo')
    await dummy_workflow('bar')
    # perform the request
    conn = await gql_sub(
        *('cylc', 'subscriptions'),
        query='''
            subscription {
                workflows {
                    id
                    status
                }
            }
        ''',
    )
    print(conn)
    while True:
        msg = await conn.read_message()
        print(msg)
        break
    assert False
```
However, I must be doing something wrong.. as it hangs on the `conn = await gql_sub` ... Ref to:
https://www.tornadoweb.org/en/stable/_modules/tornado/websocket.html#websocket_connect
and
https://github.com/jupyter-server/pytest-jupyter/blob/main/pytest_jupyter/jupyter_server.py

Will have another crack at it next week...

oliver-sanders:
If this is a new test, maybe remove it from this PR and prepare a new one?

cylc/uiserver/graphql/tornado_ws.py
```diff
@@ -0,0 +1,399 @@
...
+        await self.send_message(connection_context, op_id, GQL_COMPLETE)
+        await connection_context.unsubscribe(op_id)
+        await self.on_operation_complete(connection_context, op_id)
+
+    async def send_message(
+        self, connection_context, op_id=None, op_type=None, payload=None
+    ):
+        if op_id is None or connection_context.has_operation(op_id):
+            message = self.build_message(op_id, op_type, payload)
+            return await connection_context.send(message)
```
MetRonnie, line 337: Not a new problem in this PR, but I have figured out why we don't have any record of subscription errors anywhere. It's because this function doesn't do anything if the operation ID is not registered in the connection context, which it isn't for a graphql subscription error. 

But why doesn't it do anything? As far as I can tell, neither `self.build_message` nor `connection_context.send` make use of the iterator registered with the operation ID, so why not just do?

```suggestion
        message = self.build_message(op_id, op_type, payload)
        return await connection_context.send(message)
```

To test out what I'm talking about, try running the GUI with a typo in the `Workflows.vue` subscription, or by doing a GraphiQL subscription like
```gql
subscription App {
  deltas {
    id
    borked
  }
}
```

dwsutherland:
> If this is a new test, maybe remove it from this PR and prepare a new PR to add the tests?

Yes, will do..

cylc/uiserver/graphql/tornado.py
```diff
@@ -233,8 +233,15 @@ async def get_response(self, data):
...
+            try:
+                result = self.json_encode(response)
+            except TypeError:
+                # Back Compat < 8.5
+                response["errors"] = [f'{response["data"]}']
+                del response["data"]
 
-            result = self.json_encode(response)
+                status_code = 400
+                result = self.json_encode(response)
```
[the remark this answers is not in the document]
dwsutherland: Can't do that, it would mean `asyncio.gather` will stop at the first exception and not return the response from other workflows.. 

i.e. if you have one workflow running at `8.4.x` and another at `8.5.x`
```
mutation {
  reload(
      workflows: ["~sutherlander/five/run1", "~sutherlander/five/run2"],
      reloadGlobal: true) {
    result
  }
}
```
I've changed the commit to be more general and sift through the response to separate out the errors/exceptions from the others:

image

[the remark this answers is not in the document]
MetRonnie: > I've changed the commit to be more general and sift through the response to separate out the errors/exceptions from the others

Ok this does the trick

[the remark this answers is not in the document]
dwsutherland: > Is this actually a problem?

We need multiple mutations, and at some point we should use the response (especially for the failed ones)..
Either way, this solves the problem..

>  this is leading to this problem and the problem of the nested reload response

I actually had a solution in, but I think you removed it at some point (?) because it may have had other consequences (or it broke something)..
Thing is, it's one request from the UI/API so it needs to be folded into one response .. but sure, could be better.

oliver-sanders (approved):

PR #674, opened 2025-03-10
Make `/version` endpoint return cylc-flow and UIS versions as JSON

MetRonnie:
This is probably more useful info than the UI build version that we currently show. At the very least it is useful to developers who can check whether they are on master or the stable branch.

Plus:
- ~fix `c.CylcUIServer.log_level` not working in jupyter config.~
- copied ruff/black/isort config from cylc-flow

**Check List**

- [x] Tests are not needed.
- [x] Changelog entry not included as minor change
- [x] No docs needed

### DIFF
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -15,33 +15,42 @@
 
 from asyncio import Queue
 from functools import wraps
-import json
 import getpass
+import json
 import os
 import re
-from typing import TYPE_CHECKING, Callable, Dict
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    Dict,
+)
 
+from cylc.flow import __version__ as cylc_flow_version
 from graphene_tornado.tornado_graphql_handler import TornadoGraphQLHandler
 from graphql import get_default_backend
 from graphql_ws.constants import GRAPHQL_WS
 from jupyter_server.base.handlers import JupyterHandler
-from tornado import web, websocket
+from tornado import (
+    web,
+    websocket,
+)
 from tornado.ioloop import IOLoop
 
-from cylc.flow.scripts.cylc import (
-    get_version as get_cylc_version,
-    list_plugins as list_cylc_plugins,
+from cylc.uiserver import __version__
+from cylc.uiserver.authorise import (
+    Authorization,
+    AuthorizationMiddleware,
 )
-
-from cylc.uiserver.authorise import Authorization, AuthorizationMiddleware
 from cylc.uiserver.utils import is_bearer_token_authenticated
 from cylc.uiserver.websockets import authenticated as websockets_authenticated
 
+
 if TYPE_CHECKING:
-    from cylc.uiserver.resolvers import Resolvers
-    from cylc.uiserver.websockets.tornado import TornadoSubscriptionServer
     from graphql.execution import ExecutionResult
     from jupyter_server.auth.identity import User as JPSUser
+
+    from cylc.uiserver.resolvers import Resolvers
+    from cylc.uiserver.websockets.tornado import TornadoSubscriptionServer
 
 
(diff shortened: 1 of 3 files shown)


cylc/uiserver/app.py
```diff
@@ -463,13 +465,17 @@ def initialize_settings(self):
...
         # startup messages
-        self.log.info("Starting Cylc UI Server")
-        self.log.info(f'Serving UI from: {self.ui_path}')
+        self.log.info(f"Starting Cylc UI Server v{__version__}")
+        self.log.info(f"Serving UI from: {self.ui_path}")
+
+        try:
+            self.log.setLevel(self.log_level)
+        except Exception as exc:
+            self.log.warning(exc)
```
oliver-sanders, line 474: I'm not sure whether this is the correct behaviour for the `log_level` trait. It's a bit complicated, but it is used [by traitlets] as the default log level for terminal output, not for the logger itself.

cylc/uiserver/handlers.py
```diff
@@ -263,19 +275,19 @@ def set_default_headers(self) -> None:
     def get(self):
         user_info = {
             **self.current_user.__dict__,
-            **get_user_info(self)
+            **get_user_info(self),
+            'uis_version': __version__,
```
oliver-sanders, line 279: We do have a `cylc/version` endpoint which already contains this information, albeit in a less machine parsable format. Maybe we should highlight this information in the GUI somewhere?

If we want to put this in the application footer, then userprofile makes more sense (easier). But do we really want to display both the UI and UIS versions to users?

MetRonnie: Getting the `cylc/version` endpoint to be JSON is either involving refactoring some `cylc-flow` code, or reimplementing it in a different way in `cylc-uiserver`. And we would need to decide what info we want to include. At the moment it looks like:
```
8.4.1 (~/apps/cylc-8.4.1-1/bin/cylc)

Plugins:
  cylc-uiserver  1.6.1  ~/apps/cylc-8.4.1-1/lib/python3.9/site-packages  
  cylc-rose      1.5.0  ~/apps/cylc-8.4.1-1/lib/python3.9/site-packages  

Entry Points:
  cylc.command:                                                                                                          
    gui                cylc-uiserver  cylc.uiserver.scripts.gui:main           
    hub                cylc-uiserver  cylc.uiserver.scripts.hub:main [hub]     
    hubapp             cylc-uiserver  cylc.uiserver.scripts.hubapp:main [hub]  
  cylc.post_install:                                                                                                     
    rose_opts          cylc-rose      cylc.rose.entry_points:post_install      
  cylc.pre_configure:                                                                                                    
    rose               cylc-rose      cylc.rose.entry_points:pre_configure  
```

MetRonnie: I have made the `version` endpoint simply return
```json
{
  "cylc-flow": "<version>",
  "cylc-uiserver": "<version>",
}
```

MetRonnie:
Build check should be fixed by https://github.com/cylc/cylc-uiserver/pull/680

setup.cfg
```diff
@@ -103,6 +103,7 @@ tests =
     # https://github.com/pytest-dev/pytest/issues/12263
     pytest>=6,<8.2
     towncrier>=24.7.0
+    setuptools>=71.1
```
oliver-sanders, line 106: For typing?

oliver-sanders (approved):

setup.cfg
MetRonnie: Yes, it includes some needed type hints for `pkg_resources`, or something

MetRonnie:
```suggestion
    # Library stubs for "pkg_resources":
    setuptools>=71.1
```

PR #675, opened 2025-03-12
CPU and Max RSS Analysis tools

ChrisPaulBennett:
This apart of 3 pull requests for adding CPU time and Max RSS analysis to the Cylc UI.

This adds the Max RSS and CPU time (as measured by cgroups) to the table view, box plot and time series views.

Linked to;
https://github.com/cylc/cylc-flow/pull/6663
https://github.com/cylc/cylc-ui/pull/2100

**Check List**


cylc/uiserver/schema.py
```diff
@@ -287,9 +289,50 @@ async def get_elements(root, info, **kwargs):
...
+                          for item in json.loads(namespace.runtime.directives)}
+
+            memory = int(directives.get('--mem', 0))
+            cpus = math.ceil(int(directives.get('--tasks', 0)) / 2)
+            time = float(directives.get('--time', 0))
+
+            # TODO: put the results into the format GraphQL is expecting
+            ret[...] = {'memory': memory, 'cpus': cpus, 'time': time}
+    breakpoint()
+    return ret
```
oliver-sanders, line 333: Strongly recommend keeping these PRs to the profiler functionality alone so we can get that functionality released and in use more quickly.

We can come back to requested resource in a later PR.

ChrisPaulBennett: Done

MetRonnie:
Tests are failing

ChrisPaulBennett:
I've merged in the previous changes to the SQL infrastructure that deals with task states. @oliver-sanders I've had to change some of your code that creates parts of the SQL query, mostly just to take into account the new join.
The codecov failures are from code that I haven't touched.

MetRonnie:
Sorry for the delay, will get round to this soon but first needs deconflicting

cylc/uiserver/schema.py
```diff
@@ -684,40 +767,63 @@ def run_jobs_query(
     # filter by task name (special UIS argument for namespace queries)
     if tasks:
         where_stmts.append(
-            r'(name = '
-            + r" OR name = ".join('?' for task in tasks)
+            r'(data.name = '
+            + r" OR data.name = ".join('?' for task in tasks)
```
MetRonnie, line 771: There are other `WHERE` conditions above, do these also need `data.` prepended to the variables in them?

cylc/uiserver/schema.py
```diff
@@ -738,7 +844,19 @@ def run_jobs_query(
...
+            'job_ID': row['job_ID'],
+            'submit_num': row['submit_num'],
+            'name': row['name'],
+            'platform': row['platform'],
+            'submitted_time': row['submitted_time'],
+            'total_time': row['total_time'],
+            'run_time': row['run_time'],
+            'queue_time': row['queue_time'],
+            'max_rss': row['max_rss'],
+            'cpu_time': row['cpu_time']
```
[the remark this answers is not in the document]
ChrisPaulBennett: That syntax breaks my tiny brain. I think it's much less readable. But I'll defer to your better judgement

cylc/uiserver/schema.py
ChrisPaulBennett: As above they aren't needed. But adding to keep the code consistent

cylc/uiserver/schema.py
```diff
@@ -345,8 +345,43 @@ def run_task_query(conn, workflow):
...
+  CAST(SUBSTR(te.message,
+        INSTR(te.message, 'cpu_time ') + 9,
+        INSTR(te.message, ' max_rss') - (INSTR(te.message, 'cpu_time ') + 9)
+      ) AS INT ) AS cpu_time
+  FROM data
+  LEFT JOIN task_events te ON data.name = te.name AND data.cycle = te.cycle
+  AND data.submit_num = te.submit_num
+  WHERE te.message LIKE '%cpu_time%'
+),
+data2 AS (
```
MetRonnie, line 383: Could do with clearer names than `data1`, `data2` etc

cylc/uiserver/schema.py
```diff
@@ -747,33 +853,55 @@ def run_jobs_query(
...
     std_dev_run_time = graphene.Int()
-    total_quartiles = graphene.List(
+    run_quartiles = graphene.List(
         graphene.Int,
         description=sstrip('''
-            List containing the first, second,
-            third and forth quartile total times.'''))
+                List containing the first, second,
+                third and forth quartile run times.'''))
+    max_rss = graphene.String()
```
MetRonnie, line 884: Shouldn't this be?
```suggestion
    max_rss = graphene.Int()
```

cylc/uiserver/schema.py
```diff
@@ -747,33 +853,55 @@ def run_jobs_query(
...
+    max_rss_quartiles = graphene.List(
+        graphene.Int,
+        description=sstrip('''
+                List containing the first, second,
+                third and forth quartile for Max RSS.'''))
+    min_cpu_time = graphene.Int()
+    mean_cpu_time = graphene.Int()
+    max_cpu_time = graphene.Int()
+    total_cpu_time = graphene.Int()
+    std_dev_cpu_time = graphene.Int()
```
MetRonnie, line 898: Actually shouldn't all these integers actually be floats?

MetRonnie (commented):
These changes appear to have broken compatibility with workflows run in earlier versions of Cylc. I tested the `tasks` GraphQL query on a workflow that ran in 8.4.4.dev and got an empty result
```gql
query {
  tasks(live: false, workflows: ["wflow"]) {
    name
  }
}
```
```json
{
  "data": {
    "tasks": []
  }
}
```

ChrisPaulBennett:
This been fixed. Max RSS and CPU time will be 0, if no data is present. An update to the cylc-ui needs to be made to not show cpu time / max rss charts if there is no data

cylc/uiserver/schema.py
ChrisPaulBennett: CPU time should be integers, but standard deviation should be a float, yes

MetRonnie:
> This been fixed. Max RSS and CPU time will be 0, if no data is present. An update to the cylc-ui needs to be made to not show cpu time / max rss charts if there is no data

Is there a way to make this null instead of 0? Because currently the UI can't tell the difference between no data and times that are actually zero (to the nearest whole second).

MetRonnie:
I don't think that fix is valid - the UI should treat 0 as 0 seconds, and null/undefined should be treated as empty. So the uiserver should send null when there is no data rather than 0. Some tasks may run in ~0 seconds so 0 is a valid normal value.

cylc/uiserver/schema.py
```diff
@@ -670,60 +772,92 @@ def run_jobs_query(
...
+            ) AS INT
+        ) AS mem_alloc,
+        CAST(
+            SUBSTR(
+                te.message,
+                INSTR(te.message, 'max_rss ') + 9,
+                INSTR(te.message, ' mem_alloc') -
+                (INSTR(te.message, 'max_rss ') + 9)
+            ) AS INT
+        ) AS max_rss,
```
oliver-sanders: An alternative to this string processing would be JSON processing.

E.G, messages would look like `profiler:{"max_rss": 1234, "mem_alloc": 2345, "cpu_time": 3456}` with no constraint on field order.

The query would look like:

```sql
SELECT
  j.name
  j.cycle
  j.run_status
  JSON_EXTRACT(SUBSTR(e.message, 8), '$/cpu_time') cpu_time
  JSON_EXTRACT(SUBSTR(e.message, 8), '$/max_rss') max_rss
  JSON_EXTRACT(SUBSTR(e.message, 8), '$/mem_alloc') mem_alloc
FROM
  task_jobs j
LEFT JOIN
  task_events e
ON
  j.name = e.name
  AND ...
  AND e.message LIKE 'profiler:%'
```

This would remove the need for the `CAST` and `INSTR` logic.

I think it would also remove the need for the `WITH data AS` part. Hopefully that would make it easier to turn the `LEFT JOIN` on/off depending on whether we require the additional values or not.

oliver-sanders (commented):
Tested this end-to-end and can confirm it is working as intended.

I have a couple of suggestions to simply things:
1. Switch the message format from string to JSON.
   * SQL/Sqlite has JSON functionality which makes pulling multiple messages out of a single field easier.
   * This removes the need for many of the `INSTR` and `SUBSTR` operators.
   * This also removes the need for the `CAST` operators. 
2. Only retrieve/process these new fields when they were requested.
   * The jobs query is used for other things in the GUI.
   * The tasks query isn't yet, but will be one day.
   * To reduce the overhead added with this PR, we aught to avoid doing any heavy lifting when it isn't necessary.

I have implemented these suggestions in:

* https://github.com/ChrisPaulBennett/cylc-flow/pull/2
* https://github.com/ChrisPaulBennett/cylc-uiserver/pull/7

See what you think.

cylc/uiserver/schema.py
```diff
@@ -670,59 +816,108 @@ def run_jobs_query(
...
+        'name',
+        'cycle_point',
+        'submit_num',
+        'submit_status',
+        'run_status',
+        'started_time',
+    }
+
+    # the SQL statements required to extract each field
+    _fields = {
```
MetRonnie: Perhaps rename this to `fields_map` or something?

cylc/uiserver/schema.py
```diff
@@ -362,113 +431,189 @@ def run_task_query(conn, workflow):
...
+  MAX(CASE WHEN cpu_time_quartile = 2 THEN cpu_time END)
+  AS cpu_time_quartile_2,
+  MAX(CASE WHEN cpu_time_quartile = 3 THEN cpu_time END)
+  AS cpu_time_quartile_3,
+
+  COUNT(*) AS n
+FROM time_stats
+GROUP BY name, platform_name;
 '''):
+        total_of_totals += row[40]
```
MetRonnie:
```suggestion
        total_of_totals += row['n']
```

cylc/uiserver/schema.py
```diff
@@ -362,113 +431,189 @@ def run_task_query(conn, workflow):
...
         tasks.append({
             'id': workflow.duplicate(
                 cycle=row[1],
                 task=row[0],
                 job=row[2]
```
MetRonnie:
```suggestion
                cycle=row['cycle'],
                task=row['name'],
                job=f"{row['submit_num']:02d}",
```

cylc/uiserver/tests/test_workflow_retrieval.py
```diff
@@ -62,21 +62,37 @@ def make_db(*task_entries):
     '''
     )
 
+    conn.execute(
+        '''
+        CREATE TABLE 
+            task_events(
+                cycle TEXT, 
+                name TEXT, 
+                submit_num INTEGER,
```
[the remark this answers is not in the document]
MetRonnie: The order is correct now

MetRonnie (commented):
Some of the added test parameters are taking up a lot of vertical space. I've opened a PR to blacken these: https://github.com/ChrisPaulBennett/cylc-uiserver/pull/8

cylc/uiserver/schema.py
ChrisPaulBennett: Done

cylc/uiserver/schema.py
ChrisPaulBennett: Done

cylc/uiserver/schema.py
ChrisPaulBennett: Done

cylc/uiserver/schema.py
```diff
@@ -289,6 +291,26 @@ class Meta:
     result = GenericScalar()
 
 
+def _get_requested_fields(query_type, info):
```
oliver-sanders: (Worth copying the `Literal` type hint into this function declaration)

cylc/uiserver/schema.py
```diff
@@ -362,113 +431,189 @@ def run_task_query(conn, workflow):
...
-            'queue_quartiles': [row[14],
-                                row[14] if row[15] is None else row[15],
-                                row[14] if row[16] is None else row[16]],
+            'queue_quartiles': [row["queue_quartile_1"],
+                                row["queue_quartile_1"]
+                                if row["queue_quartile_2"] is None
+                                else row["queue_quartile_2"],
+                                row["queue_quartile_1"]
+                                if row["queue_quartile_3"] is None
+                                else row["queue_quartile_3"]],
```
oliver-sanders: 🤮, these quartiles blocks are quite hard to follow, consider formatting with black/ruff.

MetRonnie:
> > > This been fixed. Max RSS and CPU time will be 0, if no data is present. An update to the cylc-ui needs to be made to not show cpu time / max rss charts if there is no data
> > 
> > 
> > Is there a way to make this null instead of 0? Because currently the UI can't tell the difference between no data and times that are actually zero (to the nearest whole second).
> 
> This has been fixed in the UI repo

I'm still seeing zeros for tasks where the profiler was not enabled:
```json
                "maxMaxRss": 0,
                "meanMaxRss": 0,
                "stdDevMaxRss": 0,
                "minMaxRss": 0,
                "maxRssQuartiles": [0, 0, 0],
                "maxCpuTime": 0,
                "meanCpuTime": 0,
                "stdDevCpuTime": 0.0,
                "minCpuTime": 0,
                "totalCpuTime": 0,
                "cpuTimeQuartiles": [0, 0, 0],
                "totalOfTotals": 1294,
                "memAlloc": 0,
                "__typename": "UISTask"
```
I don't think this can be fixed UI-side because there would be no way for it to know if a zero is a zero or a lack of data

MetRonnie:
I'm also seeing some errors in the graphql query response. Looks like a type mismatch between the schema and the returned data.
```json
    "errors": [
        {
            "message": "Int cannot represent non-integer value: 120276004.88281766",
            "locations": [{ "line": 23, "column": 5 }],
            "path": ["tasks", 0, "stdDevMaxRss"]
        },
        {
            "message": "Int cannot represent non-integer value: 10362273.587175837",
            "locations": [{ "line": 23, "column": 5 }],
            "path": ["tasks", 1, "stdDevMaxRss"]
        },
        {
            "message": "Int cannot represent non-integer value: 11066.333333333334",
            "locations": [{ "line": 27, "column": 5 }],
            "path": ["tasks", 1, "meanCpuTime"]
        }
    ],
```

cylc/uiserver/schema.py
```diff
@@ -362,113 +431,189 @@ def run_task_query(conn, workflow):
...
+)
+SELECT
+  name,
+  cycle,
+  submit_num,
+  submit_status,
+  run_status,
+  time_run,
+  time_run_exit,
+  job_id,
```
MetRonnie: Why select unique columns like these in the final SELECT? These will return an arbitary row value when all we care about are the aggregate stats plus name and platform.

E.g., the cycle point returned will be arbitary and not useful since we are getting aggregate stats over all cycle points.

cylc/uiserver/schema.py
```diff
@@ -362,113 +431,189 @@ def run_task_query(conn, workflow):
...
+  AS stddev_total_time,
+  MAX(CASE WHEN total_time_quartile = 1 THEN total_time END)
+  AS total_quartile_1,
+  MAX(CASE WHEN total_time_quartile = 2 THEN total_time END)
+  AS total_quartile_2,
+  MAX(CASE WHEN total_time_quartile = 3 THEN total_time END)
+  AS total_quartile_3,
+
+  -- Calculate RSS stats
+  MIN(max_rss) AS min_max_rss,
```
MetRonnie, line 481: I have noticed this and min CPU will always be zero while a job is running

oliver-sanders: Aah, does this mean running jobs will skew the stats?

Shouldn't we have a state filter to prevent this (e.g, we don't want failed jobs skewing the stats either).

cylc/uiserver/schema.py
MetRonnie: Could create a helper function for this, as it's repeated several times

ChrisPaulBennett: I've blackened schema.py

cylc/uiserver/schema.py
ChrisPaulBennett: Done

cylc/uiserver/schema.py
ChrisPaulBennett: Removed some of them. Cycle is still needed for `id`. Though, I'm not sure what that is doing / whether it is needed or not, I'll investigate some more

MetRonnie: I don't think we can legitimately provide an ID, as the cycle point from the query will be arbitrary. We can only legitimately provide the task name, right?

cylc/uiserver/schema.py
```diff
@@ -363,113 +404,211 @@ def run_task_query(conn, workflow):
...
+                        row["queue_quartile_1"]
+                        if row["queue_quartile_2"] is None
+                        else row["queue_quartile_2"]
+                    ),
+                    (
+                        row["queue_quartile_1"]
+                        if row["queue_quartile_3"] is None
+                        else row["queue_quartile_3"]
+                    ),
+                ],
```
MetRonnie: To reduce repetition, how about something like
```python
def get_quartiles(prop: str) -> list[float]:
    return [
        row[f"{prop}_quartile_1"],
        (
            row[f"{prop}_quartile_1"]
            if row[f"{prop}_quartile_2"] is None
            else row[f"{prop}_quartile_2"]
        ),
        (
            row[f"{prop}_quartile_1"]
            if row[f"{prop}_quartile_3"] is None
            else row[f"{prop}_quartile_3"]
        ),
    ]
```

cylc/uiserver/schema.py
ChrisPaulBennett: Need to have a think about how best to solve this.

cylc/uiserver/schema.py
ChrisPaulBennett: You're right. It's not needed. I've removed them.

cylc/uiserver/schema.py
ChrisPaulBennett: Done

cylc/uiserver/schema.py
ChrisPaulBennett: After some digging, I think the issue is caused by the formatting functions in `utils/tasks.js`.
The code in UI server does pass a `NULL` as expected, but's it's not being handled correctly for CPU time

MetRonnie: To clarify: I was getting zero values from the SQL query, so this is not related to the UI (unless there is a separate issue there too)

cylc/uiserver/tests/test_workflow_retrieval.py
```diff
@@ -223,7 +223,7 @@ def test_make_task_query_2():
                 '2022-12-14T16:10:00Z',
                 1,
                 'message debug',
-                '_cylc_profiler: {"cpu_time": 1994, "max_rss": 50064,'
+                '_cylc_profiler: {"cpu_time": 1994, "max_rss": 50064,',
                 ' "memory_allocated": 1048576}',
```
oliver-sanders, line 227: I don't think the addition of this comma was correct?

cylc/uiserver/schema.py
```diff
@@ -738,49 +893,79 @@ def run_jobs_query(
...
+
     min_total_time = graphene.Int()
-    mean_total_time = graphene.Int()
+    mean_total_time = graphene.Float()
     max_total_time = graphene.Int()
-    std_dev_total_time = graphene.Int()
-    queue_quartiles = graphene.List(
+    std_dev_total_time = graphene.Float()
+    total_quartiles = graphene.List(
         graphene.Int,
```
oliver-sanders, line 920: This says `int`, but the signature of `get_quartiles` says `float`?

oliver-sanders:
(FYI: Plz only run code formatters on the lines changed in a PR, not the whole file to avoid extraneous diff)

cylc/uiserver/tests/test_workflow_retrieval.py
ChrisPaulBennett: Done

cylc/uiserver/schema.py
ChrisPaulBennett: Right you are. The type hint was wrong

cylc/uiserver/schema.py
ChrisPaulBennett: Have you got an offending DB file @MetRonnie? I'm struggling to recreate zero values from an SQL query.

oliver-sanders: @ChrisPaulBennett, from the description above, this should happen for running tasks, so try `sleep 1000`.

oliver-sanders: @MetRonnie, I seem to be getting `null` values for tasks & jobs queried via GraphiQL.

oliver-sanders:
There are a few threads of discussion in an unclear state that we need to resolve.

Zeros in GraphQL response:

* [x] https://github.com/cylc/cylc-uiserver/pull/675#discussion_r3638270304
* [x] https://github.com/cylc/cylc-uiserver/pull/675#issuecomment-5058341915

Errors in response:

* [x] https://github.com/cylc/cylc-uiserver/pull/675#issuecomment-5058415188

Reduce SQL query requirements for non-analysis view usage:

* [x] https://github.com/cylc/cylc-uiserver/pull/675#discussion_r2686987094

Note to self, remember to [re-test back-compatibility](https://github.com/cylc/cylc-uiserver/pull/675#pullrequestreview-3017171155) in light of recent changes

cylc/uiserver/schema.py
MetRonnie: OK this was fixed since my comment, by removing the `COALESCE` from the query in d77092878c0c66d53864a8227ebf0cce12fb98fb

But there are a few other examples of `COALESCE` that presumably need to be removed? https://github.com/cylc/cylc-uiserver/blob/71db9345fd53bb763065a0abb3b82e4adc917ed0/cylc/uiserver/schema.py#L836-L850

cylc/uiserver/schema.py
```diff
@@ -738,49 +893,79 @@ def run_jobs_query(
...
+    std_dev_run_time = graphene.Float()
+    run_quartiles = graphene.List(
         graphene.Int,
         description=sstrip('''
-            List containing the first, second,
-            third and forth quartile total times.'''))
+                List containing the first, second,
+                third and forth quartile run times.'''),
+    )
+    max_rss = graphene.BigInt()
```
MetRonnie: This is not returned by the SQL query, so is always `null`. (Note in the SQL query it is going to vary by job. We cannot return this, only the aggregate stats)
```suggestion
```

cylc/uiserver/schema.py
```diff
@@ -346,130 +356,226 @@ async def list_elements(query_type, workflows: 'Iterable[Tokens]', **kwargs):
...
+  AS stddev_max_rss,
+  MAX(CASE WHEN max_rss_quartile = 1 THEN max_rss END) AS max_rss_quartile_1,
+  MAX(CASE WHEN max_rss_quartile = 2 THEN max_rss END) AS max_rss_quartile_2,
+  MAX(CASE WHEN max_rss_quartile = 3 THEN max_rss END) AS max_rss_quartile_3,
+
+  -- Calculate CPU time
+  MIN(cpu_time) AS min_cpu_time,
+  AVG(cpu_time) AS mean_cpu_time,
+  MAX(cpu_time) AS max_cpu_time,
+  TOTAL(cpu_time) AS total_cpu_time,
```
MetRonnie: This is returning `0` when `cpu_time` is `null`/`None`. Use `SUM` instead: https://sqlite.org/lang_aggfunc.html#sumunc

cylc/uiserver/schema.py
```diff
@@ -346,130 +356,226 @@ async def list_elements(query_type, workflows: 'Iterable[Tokens]', **kwargs):
...
+    tj.name,
+    tj.cycle,
+    tj.submit_num,
+    tj.submit_status,
+    tj.time_run,
+    tj.time_run_exit,
+    tj.job_id,
+    tj.platform_name,
+    tj.time_submit,
+    tj.run_status,
```
MetRonnie: Can you please check whether all of these really need to be returned by the SELECT statement?

cylc/uiserver/schema.py
```diff
@@ -738,49 +893,79 @@ def run_jobs_query(
...
-            third and forth quartile total times.'''))
+                List containing the first, second,
+                third and forth quartile run times.'''),
+    )
+    max_rss = graphene.BigInt()
+    min_max_rss = graphene.BigInt()
+    mean_max_rss = graphene.Float()
+    max_max_rss = graphene.BigInt()
+    std_dev_max_rss = graphene.Float()
+    max_rss_quartiles = graphene.List(
```
MetRonnie: Actually these names are somewhat confusing. Perhaps "peak RSS" would be clearer, so then it would be `min_peak_rss`, `max_peak_rss` etc?

cylc/uiserver/schema.py
ChrisPaulBennett: Done

ChrisPaulBennett: Done

cylc/uiserver/schema.py
ChrisPaulBennett: Done

cylc/uiserver/schema.py
ChrisPaulBennett: Right you are

cylc/uiserver/schema.py
ChrisPaulBennett: Done

cylc/uiserver/tests/test_workflow_retrieval.py
```diff
@@ -63,119 +63,181 @@ def make_db(*task_entries):
...
+    assert ret['min_cpu_time'] == 994
     assert ret['name'] == 'Task_1'
     assert ret['platform'] == 'MyPlatform'
-    assert ret['started_time'] == '2022-12-14T15:01:00Z'
-    assert ret['state'] == 'succeeded'
     assert ret['std_dev_queue_time'] == pytest.approx(0.0, 0.01)
     assert ret['std_dev_run_time'] == pytest.approx(0.0, 0.01)
     assert ret['std_dev_total_time'] == pytest.approx(0.0, 0.01)
-    assert ret['submit_num'] == 1
-    assert ret['submitted_time'] == '2022-12-14T15:00:00Z'
```
oliver-sanders, line 124: Are we not testing started_time/submitted_time/state/submit_num any more?

cylc/uiserver/tests/test_workflow_retrieval.py
```diff
@@ -187,87 +249,115 @@ def test_make_task_query_2():
     assert ret['min_total_time'] == 600
     assert ret['name'] == 'Task_1'
     assert ret['platform'] == 'MyPlatform'
-    assert ret['started_time'] == '2022-12-15T15:01:16Z'
-    assert ret['state'] == 'succeeded'
     assert ret['std_dev_queue_time'] == pytest.approx(8.00, 0.01)
     assert ret['std_dev_run_time'] == pytest.approx(52.0, 0.01)
     assert ret['std_dev_total_time'] == pytest.approx(60.0, 0.01)
-    assert ret['submit_num'] == 1
-    assert ret['submitted_time'] == '2022-12-15T15:00:00Z'
```
oliver-sanders, line 196: Are we not testing these fields any more?

cylc/uiserver/tests/test_workflow_retrieval.py
```diff
@@ -279,43 +369,29 @@ def test_make_task_query_3():
     assert ret['min_total_time'] == 600
     assert ret['name'] == 'Task_1'
     assert ret['platform'] == 'MyPlatform'
-    assert ret['started_time'] == '2022-12-16T15:01:16Z'
-    assert ret['state'] == 'succeeded'
     assert ret['std_dev_queue_time'] == pytest.approx(7.54, 0.01)
     assert ret['std_dev_run_time'] == pytest.approx(49.02, 0.01)
     assert ret['std_dev_total_time'] == pytest.approx(56.56, 0.01)
-    assert ret['submit_num'] == 1
-    assert ret['submitted_time'] == '2022-12-16T15:00:00Z'
```
oliver-sanders, line 288: Are we not testing these fields any more?

oliver-sanders (commented):
Some fields removed from the tests, not sure if this was purposeful?

cylc/uiserver/schema.py
oliver-sanders: Not quite sure what "done" means.

Note, this SQL statement has three `SELECT`s, with some overlap, e.g, all three request `name` and the first two both select `time_run`.

cylc/uiserver/schema.py
```diff
@@ -815,49 +882,78 @@ def run_jobs_query(
...
+    total_cpu_time = graphene.Int()
+    std_dev_cpu_time = graphene.Float()
+    cpu_time_quartiles = graphene.List(
+        graphene.Int,
+        description=sstrip('''
+                List containing the first, second,
+                third and forth quartile for CPU time.'''),
+    )
+    total_of_totals = graphene.Int()
+    mem_alloc = graphene.BigInt()
```
oliver-sanders: Graphene seems to support this, but, from a quick search, it's unclear whether GraphQL does.

Does the JS end of the stick still work when this integer is maxed out?

ChrisPaulBennett: It's a been a while since I did this. But this was the cause issue's with mem alloc not getting displayed.
Now I think about it, dividing down to megabytes would make the number smaller. Similar to what I do with CPU time

oliver-sanders: Reducing the units would make sense, I can't think why we would care about anything less than 1 megabyte! If doing this, make sure to do it consistently across all fields (we don't want some reporting bytes and others megabytes).

Would be nice to add some `description` lines here to convey the units.

ChrisPaulBennett: I'd rather leave the BigInt in there and leave it as is, unless there is an issue with it?

cylc/uiserver/schema.py
ChrisPaulBennett: Ah yes, `time_run` is only need for the initial calculation and doesn't need to be carried forward. 
`name` is needed for the `GROUP BY` statement at the end.

cylc/uiserver/schema.py
oliver-sanders: > unless there is an issue with it?

As I said:

> Graphene seems to support this, but, from a quick search, it's unclear whether GraphQL does.

Here's an example:

> there's no concept like bigInt in graphQL.
>
> -- https://stackoverflow.com/questions/62893664/how-can-i-handle-long-int-with-graphql

Graphene (the Python framework we're using) appears to support this, but it's unclear whether this is actually supported in general.

E.g, Does Apollo JS support this?

What I'm asking is, does this BitInt support actually work!

cylc/uiserver/schema.py
oliver-sanders, line 412: ([RE](https://github.com/cylc/cylc-uiserver/pull/675#discussion_r3959630317)) This one also looks like it's unused.

cylc/uiserver/schema.py
```diff
@@ -408,10 +405,7 @@ def run_task_query(conn, workflow):
 time_stats AS (
   SELECT
     name,
-    time_run,
-    time_run_exit,
     platform_name,
```
oliver-sanders, line 408: Ditto?

cylc/uiserver/schema.py
ChrisPaulBennett: I've had to create a new PR for flow to implement the change of memory values to megabytes. UI changes have also been made

cylc/uiserver/tests/test_workflow_retrieval.py
ChrisPaulBennett: No. They we aren't. These values are not apart of the task query anymore. I've also removed submit_num.

cylc/uiserver/schema.py
ChrisPaulBennett: These are used in the next select statement

cylc/uiserver/schema.py
ChrisPaulBennett: Done

cylc/uiserver/schema.py
```diff
@@ -738,49 +876,78 @@ def run_jobs_query(
...
+    std_dev_run_time = graphene.Float()
+    run_quartiles = graphene.List(
         graphene.Int,
         description=sstrip('''
-            List containing the first, second,
-            third and forth quartile total times.'''))
+                List containing the first, second,
+                third and forth quartile run times.'''),
+    )
+    min_peak_rss = graphene.Int()
```
MetRonnie: All RSS fields in the schema should be floats now, to take advantage of the size afforded by double precision

cylc/uiserver/schema.py
```diff
@@ -346,130 +356,215 @@ async def list_elements(query_type, workflows: 'Iterable[Tokens]', **kwargs):
...
+                        row["peak_rss_quartile_1"]
+                        if row["peak_rss_quartile_2"] is None
+                        else row["peak_rss_quartile_2"]
+                    ),
+                    (
+                        row["peak_rss_quartile_1"]
+                        if row["peak_rss_quartile_3"] is None
+                        else row["peak_rss_quartile_3"]
+                    ),
+                ],
```
MetRonnie:
```suggestion
                'peak_rss_quartiles': get_quartiles(row, 'peak_rss'),
```

cylc/uiserver/schema.py
```diff
@@ -346,130 +356,215 @@ async def list_elements(query_type, workflows: 'Iterable[Tokens]', **kwargs):
...
+                        row["cpu_time_quartile_1"]
+                        if row["cpu_time_quartile_2"] is None
+                        else row["cpu_time_quartile_2"]
+                    ),
+                    (
+                        row["cpu_time_quartile_1"]
+                        if row["cpu_time_quartile_3"] is None
+                        else row["cpu_time_quartile_3"]
+                    ),
+                ],
```
MetRonnie:
```suggestion
                'cpu_time_quartiles': get_quartiles(row, 'cpu_time'),
```

cylc/uiserver/schema.py
ChrisPaulBennett: Silly me!

cylc/uiserver/schema.py
```diff
@@ -346,130 +356,189 @@ async def list_elements(query_type, workflows: 'Iterable[Tokens]', **kwargs):
     return elements
 
 
+def get_quartiles(row, prop: str) -> list[int]:
+    # Prevents null entries when there are too few
+    # tasks for quartiles
```
MetRonnie, line 361: This doesn't seem to do the job quite correctly, as sometimes I see Q2 has a value and the others are null.

GraphQL response:
```json
{
        "name": "p",
        "platform": "pbs-profiler",
        "count": 2,
        "meanTotalTime": 33,
        "stdDevTotalTime": 0,
        "minTotalTime": 33,
        "totalQuartiles": [
          null,
          33,
          null
        ],
        "maxTotalTime": 33,
        "meanRunTime": 25,
        "stdDevRunTime": 0,
        "minRunTime": 25,
        "runQuartiles": [
          null,
          25,
          null
        ],
        "maxRunTime": 25,
        "meanQueueTime": 8,
        "stdDevQueueTime": 0,
        "minQueueTime": 8,
        "queueQuartiles": [
          null,
          8,
          null
        ],
        "maxQueueTime": 8,
        "maxPeakRss": 430850048,
        "meanPeakRss": 430850048,
        "stdDevPeakRss": 0,
        "minPeakRss": 430850048,
        "peakRssQuartiles": [
          null,
          430850048,
          null
        ],
        "maxCpuTime": 26118,
        "meanCpuTime": 26118,
        "stdDevCpuTime": 0,
        "minCpuTime": 26118,
        "totalCpuTime": 26118,
        "cpuTimeQuartiles": [
          null,
          26118,
          null
        ],
        "totalOfTotals": 49914,
        "memAlloc": 0,
        "__typename": "UISTask"
      },
```

cylc/uiserver/schema.py
```diff
@@ -346,130 +356,189 @@ async def list_elements(query_type, workflows: 'Iterable[Tokens]', **kwargs):
...
     # https://github.com/cylc/cylc-uiserver/issues/440
     tasks = []
+    total_of_totals = 0
     for row in conn.execute('''
-SELECT
+WITH profiler_stats AS (
+  SELECT
+    tj.name,
+    tj.platform_name,
+    JSON_EXTRACT(SUBSTR(te.message, 17), '$.memory_allocated') AS mem_alloc,
```
MetRonnie, line 388: This is likely a cylc-flow issue, but I am seeing zero memory allocated for PBS jobs. It doesn't matter if I specify the `-l mem` directive. Can you reproduce?

ChrisPaulBennett: Looking into it

cylc/uiserver/tests/test_workflow_retrieval.py
oliver-sanders: I think we've removed some of the fields that these functions used to return. E.g, `id` was returned before, but isn't now.

Note, this is unrelated to tidying unused parts of the SQL queries used within the functions.

Screenshot from 2026-09-24 11-15-44

PR #676, opened 2025-03-12
actions: provision system deps from conda-forge

oliver-sanders:
* Decouple from the OS image.
* Allow testing against a broader range of Python versions without having to manage GH actions runner images.
* Remove homebrew package manager use.
* Add pipefail protection to the default test shell.

Needs updstream change to release-actions to pass.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,6 +15,10 @@ concurrency:
 env:
   FORCE_COLOR: 2
 
+defaults:
+  run:
+    shell: bash -c "exec $CONDA_PREFIX/bin/bash -elo pipefail {0}"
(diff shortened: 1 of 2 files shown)


MetRonnie (changes requested):
MacOS is running Python 3.13 for some reason, instead of 3.8

oliver-sanders:
Fixed. It was caused GitHub actions annoying habit of silently tolerating unset template variables. In this case `${{ matrix.python }}`.

Do you know if there is any way of forcing a brittle failure for an unset variable?

MetRonnie:
Don't think so

MetRonnie (approved):

Issue #682, opened 2025-04-10
Severe memory leak/accumulation - Due to ZeroMQ subscription data build up


dwsutherland:
### Description

At NIWA we've noticed the consumption of roughly 5Gb a day of memory from long running UIS  (via `cylc gui`):

![Image](https://github.com/user-attachments/assets/bdaec071-41c7-41c3-9213-8da1e3c09e66)

As a work around we've been restarting the UIS every 4-5 days...

We're not stopping and starting workflows under different runs all the time, we have a fairly static ~57 workflows in the operation, and a similar number in the test system..

### Reproducible Example

Just running a UIS instance with one workflow running will, given time, show a reduction in free memory on the machine.

After analyzing objects in the UIS the problem didn't appear to be _normal_ objects like the data store and other UIS attributes:

![Image](https://github.com/user-attachments/assets/d684704c-16ba-4db1-8a7c-3d060ad3f861)

![Image](https://github.com/user-attachments/assets/044279a8-f4ed-4ce4-b0f3-83ab0cc2983b)

However, using this tool:
https://github.com/bloomberg/memray
on a UIS with 5 small workflows I was able to find the exact location of problem:

![Image](https://github.com/user-attachments/assets/650448b2-ad51-4e91-855d-32b19e872c1f)

Shows an increase over time:

![Image](https://github.com/user-attachments/assets/f05f2ccf-9f66-4f36-abca-8978140fee4f)

![Image](https://github.com/user-attachments/assets/32a86a63-b57d-4351-a1f3-1f58a250043c)

And It appeared to be the executor threads of the data-store used for ZeroMQ subscriptions (getting data from the Scheduler(s)), threads which this tool can isolate:

![Image](https://github.com/user-attachments/assets/f8ad1847-5c1a-41f5-81ce-6101318284e8)

More specifically the delta processing after receipt is accumulating in size:

![Image](https://github.com/user-attachments/assets/ac13ee5c-9dde-4180-b8c1-e6133567885b)

Which I suspect is something to do with the data going from the subscription thread to the main thread..

### Expected Behaviour

For a relatively static load of workflows, the UIS should consume a static amount of memory (not grow).

dwsutherland:
Well it appears not to be the threading, I took the processing out of the thread and put the incoming deltas into a queue:

![Image](https://github.com/user-attachments/assets/ec2cec21-43e5-43fc-850e-2e3a8a8f71cf)

(iterating over the queue with a periodic callback)

[...]

PR #683, opened 2025-04-12
Remove ZeroMQ subscription threading

dwsutherland:
Just fooling around with no-threading options for receiving ZeroMQ subscription data from the scheduler(s).
However, this seems quite responsive, so thought I'd post it.

I was testing whether the memory accumulation at the UIS was related to threading or not:
https://github.com/cylc/cylc-uiserver/issues/682
Turns out threading is not a problem in this regard.

**Check List**


oliver-sanders:
See also https://github.com/cylc/cylc-uiserver/pull/574

PR #684, opened 2025-04-14
Drop usage of pkg_resources

MetRonnie:
> Use of pkg_resources is deprecated in favor of [importlib.resources](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3.11/library/importlib.metadata.html#module-importlib.metadata) and their backports ([importlib_resources](https://pypi.org/project/importlib_resources), [importlib_metadata](https://pypi.org/project/importlib_metadata)).

Also, currently `cylc hub` is broken if you don't have setuptools installed (setuptools was never a dependency)!

**Check List**

- [x] Applied any dependency changes to `setup.cfg` (no `conda-environment.yml` present).
- [x] Tests are not needed
- [x] Changelog entry not needed as not user-facing
- [x] No docs needed

### DIFF
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -63,7 +63,6 @@ from types import SimpleNamespace
 from typing import List, Optional, Union
 
 from jupyter_server.extension.application import ExtensionApp
-from pkg_resources import parse_version
 from tornado import ioloop
 from tornado.web import RedirectHandler
 from traitlets import (
@@ -385,7 +384,10 @@ class CylcUIServer(ExtensionApp):
                 for version in path.glob('[0-9][0-9.]*')
                 if version
             ),
-            key=parse_version
+            key=lambda x: tuple(
+                int(x) if x.isdecimal() else x
+                for x in x.split('.')
+            )
         )
 
     @default('ui_path')
--- a/cylc/uiserver/jupyter_config.py
+++ b/cylc/uiserver/jupyter_config.py
@@ -16,11 +16,19 @@
 # Configuration file for jupyterhub.
 
 from pathlib import Path
-import pkg_resources
+import sys
+
+
+if sys.version_info[:2] > (3, 8):
+    from importlib.resources import files
+else:
+    # BACK_COMPAT: importlib_resources
+    from importlib_resources import files
 
 from cylc.uiserver import (
     __file__ as uis_pkg,
-    getenv)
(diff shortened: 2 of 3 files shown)


cylc/uiserver/app.py
```diff
@@ -385,7 +384,10 @@ def _list_ui_versions(path: Path) -> List[str]:
                 for version in path.glob('[0-9][0-9.]*')
                 if version
             ),
-            key=parse_version
+            key=lambda x: tuple(
+                int(x) if x.isdecimal() else x
+                for x in x.split('.')
+            )
```
oliver-sanders, line 390: We might as well use the importlib-metadata interface for parsing the version here as it is already a dependency of cylc-flow.

MetRonnie: This is parsing a NodeJS semver version which is not guaranteed to be compatible with Python versioning syntax. I figure this custom sort key is good enough

oliver-sanders: IMO, it's nicer to use the interface we've already got installed (and we know this package conforms to semver, because it's ours) but will leave it to you if you're keen not to.

hjoliver (approved):
I'm happy either way on the version parsing question.

cylc/uiserver/app.py
oliver-sanders: Have just realised that the `isdecimal` logic here does not work:

```python
>>> sorted(['1.1', '1.a1'], key=lambda x: tuple(int(x) if x.isdecimal() else x for x in x.split('.')))
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'str' and 'int'
```

(it did at Python 2, but Python 3 doesn't allow comparisons of str and int)

MetRonnie: Ah woops, could just do?
```suggestion
            key=lambda x: tuple(
                int(i) if i.isdecimal() else -1
                for i in x.split('.')
            )
```

(Also realised I was confusingly redefining `x` in the comprehension)

oliver-sanders: > (Also realised I was confusingly redefining x in the comprehension)

I spotted that, but it seems to work (confusingly), not quite sure how.

> Ah woops, could just do?

I would just use an existing version parser, there's a reason I've done this in the past!

PR #689, opened 2025-05-14
Ensure resolver errors are logged properly for `play` and `clean`

MetRonnie:
Currently, if playing a workflow from the UI fails, there is no evidence in the UIS logs.

This fix ensures the error is logged.

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] Changelog entry not needed as minor
- [x] No docs needed

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -16,17 +16,18 @@
 """GraphQL resolvers for use in data accessing and mutation of workflows."""
 
 import asyncio
-import errno
-import os
-import signal
 from contextlib import suppress
 from copy import deepcopy
+import errno
 from getpass import getuser
+import os
+import signal
 from subprocess import (
     DEVNULL,
     PIPE,
     Popen,
 )
+from textwrap import indent
 from time import time
 from typing import (
     TYPE_CHECKING,
@@ -41,13 +42,17 @@ from typing import (
     Union,
 )
 
-import psutil
 from cylc.flow.data_store_mgr import WORKFLOW
-from cylc.flow.exceptions import ServiceFileError, WorkflowFilesError
+from cylc.flow.exceptions import CylcError
 from cylc.flow.id import Tokens
 from cylc.flow.network.resolvers import BaseResolvers
-from cylc.flow.scripts.clean import CleanOptions, run
+from cylc.flow.scripts.clean import (
+    CleanOptions,
+    run,
+)
 from graphql.language.base import print_ast
+import psutil
+
 
 if TYPE_CHECKING:
     from concurrent.futures import Executor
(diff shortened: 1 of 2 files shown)


cylc/uiserver/resolvers.py
```diff
@@ -251,11 +255,9 @@ async def clean(
                 executor, _clean, workflow_ids, opts
             )
         except Exception as exc:
-            if isinstance(exc, ServiceFileError):  # Expected error
-                # The "workflow still running" msg is very long
-                msg = str(exc).split('\n')[0]
```
MetRonnie, line 256: The error caught here is not `ServiceFileError`, but a `CylcError`: see https://github.com/cylc/cylc-flow/blob/98181b6a80d89f5fbc57cbee964609bb320d3bde/cylc/flow/scripts/clean.py#L231-L240

We should rethink that error message, see https://github.com/cylc/cylc-flow/issues/5324

hjoliver (approved):
LGTM. Just needs a test to cover the `clean` resolver?

cylc/uiserver/resolvers.py
```diff
@@ -251,11 +255,9 @@ async def clean(
...
-            elif isinstance(exc, WorkflowFilesError):  # Expected error
+            if isinstance(exc, CylcError):  # Expected error
                 msg = str(exc)
```
oliver-sanders, line 259: We should probs format this the same way we would for the CLI:

```suggestion
                msg = f'{exc.__class__.__name__}: {exc}'
```

MetRonnie:
> LGTM. Just needs a test to cover the `clean` resolver?

> We should probs format this the same way we would for the CLI

Both done

cylc/uiserver/resolvers.py
```diff
@@ -251,13 +255,10 @@ async def clean(
...
-                # The "workflow still running" msg is very long
-                msg = str(exc).split('\n')[0]
-            elif isinstance(exc, WorkflowFilesError):  # Expected error
-                msg = str(exc)
+            msg = f"{type(exc).__name__}: {exc}"
+            if isinstance(exc, CylcError):  # Expected error
+                log.error(msg)
             else:  # Unexpected error
-                msg = f"{type(exc).__name__}: {exc}"
                 log.exception(msg)
```
oliver-sanders, line 262: Ooh, shouldn't this be `log.exception(exc)`?

oliver-sanders: No scratch that:

> Exception info is added to the logging message. This method should only be called from an exception handler.
>
> -- https://docs.python.org/3/library/logging.html#logging.Logger.exception

cylc/uiserver/resolvers.py
```diff
@@ -322,11 +323,10 @@ async def play(
...
 
                 if ret_code:
-                    # command failed
+                    msg = f"Command failed ({ret_code}): {cmd_repr}"
                     out, err = proc.communicate()
-                    results[wflow] = err.strip() or out.strip() or (
-                        f'Command failed ({ret_code}): {cmd_repr}'
-                    )
+                    results[wflow] = err.strip() or out.strip() or msg
+                    log.error(f"{msg}\n{results[wflow]}")
```
oliver-sanders, line 329: This could log `{msg}\n{msg}`?

MetRonnie: Yes, but unlikely as most of the time there should be stderr or stdout

oliver-sanders:
```suggestion
                    log.error(f"{msg}\n{err.strip() or out.strip()}")
```

?

oliver-sanders: Or:

```suggestion
                    log.error(f"{msg}\nstdout: {out.strip()}\nstderr: {err.strip()}")
```

cylc/uiserver/resolvers.py
oliver-sanders: Just spotted, David has changed this to `log.exception(exc)` on his branch.

PR #693, opened 2025-05-23
fix authorization limit-permission case mismatch

### LINKED ISSUE #692: authorisation: permissions reported can be lower than configured
We have spotted a reporting issue where the UIS logs that a user has lower permissions than they have actually been granted in some case.

This will cause the UI to erroneously grey out the corresponding commands, however, if manually actioned (e.g. via GraphiQL) the authorisation continues to work as configured.

There is no security compromise here, everything is working as intended, but permissions are being erroneously reported which will cause confusion and result in reduced functionality in the UI.

#### Reproducible Example

Grant a user read permissions, along with selected commands:

```python
    'other-user': ['READ', 'Poll', 'Play', 'Pause', 'Stop'],
}
```

When `other-user` logs on, the UIS will report them as having `[read]` permissions, but not `poll`, `play`, `pause` or `stop`. The corresponding commands will be "greyed" out in the UI because this erroneous report makes it into the user profile request leading the UI to think the user has lesser permissions and not offer these commands. However, if you action these mutations via GraphiQL, it works as configured.

dwsutherland:
closes #692

Before this change a config like this:
```

```
Would internally be expanded as:
https://github.com/cylc/cylc-uiserver/blob/98c9e397ba23c978beeb87067635cab864abe72b/cylc/uiserver/authorise.py#L305
https://github.com/cylc/cylc-uiserver/blob/98c9e397ba23c978beeb87067635cab864abe72b/cylc/uiserver/authorise.py#L191
```
limits_owner_can_give: {'CONTROL', 'READ'}
user_conf_permitted_ops: {'Play', 'Stop', 'READ', 'Poll', 'Pause'}
expanded user_conf_permitted_ops: {'Play', 'read', 'Stop', 'Poll', 'Pause'}
expanded limits_owner_can_give: {'trigger', 'pause', 'scan', 'set_verbosity', 'play', 'ext_trigger', 'set', 'read', 'hold', 'release_hold_point', 'message', 'clean', 'stop', 'kill', 'resume', 'set_graph_window_extent', 'release', 'remove', 'poll', 'reload', 'set_hold_point'}
```
The via:
```
        # subtract permissions that the site does not permit to be granted
        allowed_operations = limits_owner_can_give.intersection(
            user_conf_permitted_ops
        )
```
give `read`
```
[I 2025-05-23 14:27:42.958 CylcUIServer] User other-suther authorized permissions: ['read']
{'read'}
```

With this small fix, the permissions are correct:
```
[I 2025-05-23 16:09:28.239 CylcUIServer] User other-suther authorized permissions: ['pause', 'play', 'poll', 'read', 'stop']
{'poll', 'play', 'stop', 'pause', 'read'}
```
And UI menu correctly making available the corresponding options:
image

**Check List**

### DIFF
--- a/cylc/uiserver/authorise.py
+++ b/cylc/uiserver/authorise.py
@@ -227,6 +227,9 @@ class Authorization:
                 permission_set.remove(action_group)
                 permission_set.update(expansion)
 
+        # lowercase all before negation i.e. Play -> play
+        permission_set = {p.lower() for p in permission_set}
+
         # Remove negated permissions
         remove = set()
         for perm in permission_set:


dwsutherland:
> We should probably just validate the config to prevent this mistake being made?

Well it's documented as capitalized (i.e. `Play` or `Stop`) but the get_mutations returns all lowercase:
```
    @property
    def ALL_OPS(self) -> List[str]:
        """ALL OPS constant, returns list of all mutations."""
        return get_list_of_mutations()
    @property
    def CONTROL_OPS(self) -> List[str]:
        """CONTROL OPS constant, returns list of all control mutations."""
        return get_list_of_mutations(control=True)
```

At least with this PR it is essentially case insensitive (aside from the grouped i.e. `CONTROL` etc), which shouldn't be a problem?

oliver-sanders:
Thanks for looking into this. I've taken a look and I think this is just a documentation error, looks like it's been lowercase since auth weas introduced in 0.6.0.

I think we should probably:

* Correct the documentation.
* Reject unknown items.

Converting to lowercase is slightly dubious as the same thing can be configured multiple ways, the difference between item and group is blurred and potential for misunderstanding increased.

dwsutherland:
@oliver-sanders - If you're happy with just the documentation fix (merged), and a validate check.. I suppose we can close this?

PR #709, opened 2025-07-24
GH Actions: fix create PR step

MetRonnie:
Follow-up to #705, #706, #707

I think the problem may be to do with adding the assignee and/or reviewer, so I've split them off into separate commands that are allowed to fail.

I think it's a bug in GH CLI or the GitHub token permissions: https://github.com/orgs/community/discussions/113519


oliver-sanders:
Ach, so we can revert our attempted fixes?

MetRonnie:
Well it looks like GH CLI accepts either `GITHUB_TOKEN` or `GH_TOKEN` so no need to change anything back

PR #715, opened 2025-08-01
Use a human friendly sort order for log files

jfrost-mo:
Compared to the default by character sort this ensures that 2 is sorted before 10, etc.

Also add James Frost to list of contributors.

Fixes: https://github.com/cylc/cylc-ui/issues/2263

**Check List**


jfrost-mo:
The added sorting function has a test at https://github.com/MetOffice/CSET/blob/2f798783a33a989fafb18ecc8af12b32c0e23376/tests/test_common.py#L285-L291. I haven't ported it over as the function it is contained in is also untested.

jfrost-mo:
Updated to use the cylc flow natural_sort_key function. I've used the key rather than the full sort function as we need to reverse the order for this usage, and arguments are not passed through to the sort underlying natural_sort.

hjoliver (approved):

PR #740, opened 2025-10-03
Fixing memory leak

### LINKED ISSUE #699: WebSocketClosedError on write_message (coincidental memory accumulation)
### Description
Cause not known, but it would appear a message is send before closure of the websocket could complete.

```
Task exception was never retrieved
future: <Task finished name='Task-10268090' coro=<TornadoSubscriptionServer.on_start() done, defined at /opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/cylc/uiserver/websockets/tornado.py:129> exception=WebSocketClosedError()>
  File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/cylc/uiserver/websockets/tornado.py", line 148, in on_start
    await self.send_execution_result(
  File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/cylc/uiserver/websockets/tornado.py", line 173, in send_execution_result
    await BaseSubscriptionServer.send_execution_result(
  File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/graphql_ws/base_async.py", line 181, in send_message
    return await connection_context.send(message)
  File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/cylc/uiserver/websockets/tornado.py", line 49, in send
    await self.ws.write_message(data)
  File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/tornado/websocket.py", line 360, in write_message
    raise WebSocketClosedError()
tornado.websocket.WebSocketClosedError

  File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/cylc/uiserver/websockets/tornado.py", line 151, in on_start
    await self.send_error(connection_context, op_id, e)
  File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/graphql_ws/base_async.py", line 181, in send_message
    return await connection_context.send(message)
  File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/cylc/uiserver/websockets/tornado.py", line 49, in send
    await self.ws.write_message(data)
  File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/tornado/websocket.py", line 360, in write_message
    raise WebSocketClosedError()
tornado.websocket.WebSocketClosedError
```

**Note:** This code has changed location content since 1.6.x/8.4.x, however the send mechanism is likely similar so a fix should be against 1.7.0 once understood

Also, this was found in NIWA operations, and there are indications this may have a large contribution to memory accumulation of the UI Server...

Servers usually start with about ~3-4Gb memory used on their designated VMs and currently, after a similar load and run duration, the `oper` and `test` systems differ in their memory used:
Oper:
```
$ free -h
              total        used        free      shared  buff/cache   available
Mem:            31G         16G         13G         57M        664M         13G
Swap:            0B          0B          0B
```
Test:
```
              total        used        free      shared  buff/cache   available
Mem:            15G        3.8G        592M         61M         11G         10G
Swap:            0B          0B          0B
```

The only difference found was `oper` has 36 of these errors in the log (test none):
```
[ecox_oper@w-ec-admin02 ~]$ grep -c 'During handling of the above exception' nohup.out 
36
```
where `oper` UIS appears to have a compounding memory build..

Perhaps a coincidence, however, if subscriptions aren't cleaned up properly their queues could be accumulating deltas.. (which would explain the apparent compounding effect as the error count increases)

### Reproducible Example

Unknown.

### Expected Behaviour

Send attempt on closed web-socket is handled gracefully, ending all associated subscriptions (etc...) like `on_close`

### LINKED ISSUE #732: small memory leak for new connections/subscriptions
A small amount of memory (~1kb) is allocated with new connections or subscriptions (haven't asserted which yet) which doesn't appear to be released when the connection / subscription terminates.

The amount of memory is very small so is unlikely to be noticed in real-world situations. However, if for some reason, a client were to time out and reconnect with a high frequency, this can lead to a slow memory leak.

> [!IMPORTANT]
>
> Recommend investigating with https://github.com/cylc/cylc-flow/pull/6985 merged in.

### Steps To Reproduce

Configure a frequent websocket ping with a completely unreasonable timeout:

```python
c.ServerApp.websocket_ping_interval = 3
c.ServerApp.websocket_ping_timeout = 0.000001
```

You will need to hack jupyter_server to allow you to use float values for the ping_timeout, here's the patch (will raise as a PR to jupyter_server in due course):

```diff
diff --git a/jupyter_server/serverapp.py b/jupyter_server/serverapp.py
index 1afbef4d0..53ecc04a7 100644
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1649,7 +1649,7 @@ class ServerApp(JupyterApp):
             return "jupyter_server.gateway.connections.GatewayWebSocketConnection"
         return ZMQChannelsWebsocketConnection
 
-    websocket_ping_interval = Integer(
+    websocket_ping_interval = Float(
         config=True,
         help="""
             Configure the websocket ping interval in seconds.
@@ -1666,12 +1666,16 @@ class ServerApp(JupyterApp):
             the connection will be closed from the server side.
         """,
     )
-    websocket_ping_timeout = Integer(
+    websocket_ping_timeout = Float(
         config=True,
         help="""
             Configure the websocket ping timeout in seconds.
 
-            See ``websocket_ping_interval`` for details.
+            The default is the same as the ping interval. Set to ``0`` to turn
+            the ping timeout off.
+
+            Note the the websocket ping timeout requires the ping interval to
+            be set, see ``websocket_ping_interval`` for details.
         """,
     )
```

This will cause the client to reconnect every ~3 seconds.

```bash
# launch the server
cylc gui --new

# connect the client in a new brower tab

# wait for the client to connect, then attach your profiler of choice
mprof attach <pid>

# wait for a few connect/disconnect cycles to pass, then collect your results
mprof plot <pid>
```

Image

### Profiling

Because the amount of memory leaked is so small, it is quite hard to trace.

The best approach I have come up with so far is to compare memory snapshots, the diff of which should contain the answer.

There are two tools I'm aware of which can do this:
* `tracemalloc.take_snapshot.compare_to(previous, 'lineno')`
* `pympler.tracker.SummaryTracker.print_diff`

Here's a patch which uses `tracemalloc`:

```diff
diff --git a/cylc/uiserver/graphql/tornado_ws.py b/cylc/uiserver/graphql/tornado_ws.py
index fd58032..de24895 100644
--- a/cylc/uiserver/graphql/tornado_ws.py
+++ b/cylc/uiserver/graphql/tornado_ws.py
@@ -23,6 +23,7 @@
 #
 # It has been evolved to suit and ported to graphql-core v3.
 
+import tracemalloc
 import asyncio
 from asyncio.queues import QueueEmpty
 from contextlib import suppress
@@ -46,6 +47,8 @@ from cylc.flow.network.graphql_subscribe import subscribe
 from cylc.uiserver.authorise import AuthorizationMiddleware
 from cylc.uiserver.schema import SUB_RESOLVER_MAPPING
 
+tracemalloc.start()
+SNAP = tracemalloc.take_snapshot()
 
 NO_MSG_DELAY = 1.0
 
@@ -166,7 +169,7 @@ class TornadoSubscriptionServer:
     ):
         self.schema = schema
         self.loop = loop
-        self.middleware = middleware
+        self.middleware = middleware or []
         self.execution_context_class = execution_context_class
         self.auth = auth
 
@@ -310,6 +313,25 @@ class TornadoSubscriptionServer:
     async def on_start(self, connection_context, op_id, params):
         # Attempt to unsubscribe first in case we already have a subscription
         # with this id.
+        global SNAP
+        snapshot = tracemalloc.take_snapshot()
+        cmp = [
+            s
+            for s in snapshot.compare_to(SNAP, 'lineno')
+            if 'cylc/' in str(s)
+            # or 'tornado/' in str(s)
+            # or 'graphql/' in str(s)
+        ]
+        print()
+        print('+/-', sum(stat.size_diff for stat in cmp))
+        # with open('log', 'a') as log_file:
+        #     log_file.write(str(sum(stat.size_diff for stat in cmp)) + '\n')
+        #     log_file.flush()
+        for stat in sorted(cmp, key=lambda x: x.size_diff, reverse=True):
+            if stat.size_diff != 0:
+                print(f'  {stat}')
+        print()
+        SNAP = snapshot
         await connection_context.unsubscribe(op_id)
 
         params['kwargs']['root_value'] = op_id
```

This diff is filtering memory allocated from within the Cylc codabase.

Note that memory diffs can be a bit jumpy, sometimes memory doesn't get deallocated for a while, but does eventually.

Remove the filter and uncomment the `with open('log` bit to get the full memory diffs and log the net change to a file. Plot the result and you'll get something like this:

Image

It jumps around, the leak is barely visible in the long term trend.

### Leaks

The leak may not be in the Cylc codebase at all.

So far, I have found one minor leak originating from Cylc, however, this doesn't seem to make a big difference: https://github.com/cylc/cylc-flow/pull/6985

ChrisPaulBennett:
This PR hopefully closes #699 and closes #732

#732 has a write up of Oliver's investigation into this matter (and also how to recreate the memory leak in a predictable way)
I am doing a full write up of my investigation here as there appears to be another small memory leak here, and this write up should stop any duplicate work.

This memory leak was very tricky to pin down. This because the leak is caused by improper use of the tornado library. The tornado library calls some other code (most likely C) as part of it's functionality. This means that the leak is not seen from within python. All tools I used from within python showed no memory leak (Heapy, tracemalloc etc). But if you profile the program outside of python, using mprof in this case, the memory leak is easy to see. (mprof is a python library, but it will profile any executable you give it, no matter the language used to make it).

In a weird quirk, Valgrind (pronounced val - grinned for some reason) did not detect a memory leak. Valgrind appears to be the go to memory leak detection software on linux, it's bundled in with some distro's of linux. I'm not sure why valgrind was unable to see the memory leak, but mprof was.
Thanks to the information from #699, we knew that the memory leak was something to do with network connections, which thankfully narrowed down the problem to only a few files. 
This GitHub conversation details some memory leak problems that other projects were having. The Tornado developers claim that the issue is caused by improper use of the library.

https://github.com/tornadoweb/tornado/issues/2515

**Check List**

### DIFF
--- a/cylc/uiserver/graphql/tornado_ws.py
+++ b/cylc/uiserver/graphql/tornado_ws.py
@@ -106,9 +106,8 @@ class TornadoConnectionContext:
             raise ConnectionClosedException()
 
     async def send(self, data):
-        if self.closed:
-            return
-        await self.ws.write_message(data)
+        if not self.ws.ws_connection.is_closing():
+            await self.ws.write_message(data)
 
     @property
     def closed(self):


oliver-sanders:
Have had a quick skim of https://github.com/tornadoweb/tornado/issues/2515, there are several different examples on that thread but I don't think this issue is related. The examples on that thread do indeed look like misuse of the library. If you try to send messages faster than the client can process them, you're going to run into problems (with any framework). Another issue mentioned there is attempting to send messages from the wrong thread (?), we should probs look into this just incase we are doing it without realising. Again, legitimate misuse.

But what's going on here is different. The theory of this PR is that attempting to send a message to a client in the process of being disconnected from the server side, results in the message being persisted in memory. This makes some sense as the message probably isn't going to be sent if the connection is in the process of being closed.

Need to dig into the code and understand this better. It's possible that this is something Tornado could/should defend against, if so, we should open an issue over there. But it's also possible that this is an issue of our own creation ;(, need to dig.

ChrisPaulBennett:
I've done a bit more digging into the tornado library. The "close_code" variable (and it's accompanying "close_reason" variable) stores the reason why the connection was closed. When a web socket object is created, this variable is set to `None`. When a connection is closed, using the `on_ws_connection_close` method is called, it set's the `close_code` and `close_reason` variable and invokes the `on_connection_close()` method. I'm assuming that if this connection is not closed properly then these variables don't get set and the `close_code` variable stays at `None`. So a misbehaving / unstable network connection would cause issue's.
I would argue that using `close_code()` is misusing the library.
But I also totally agree that Tornado should be defending against this. This is a very subtle / non-obvious leak.

oliver-sanders:
Profiling results, left:before, right:after:

Screenshot from 2025-10-16 10-19-41

The first 10 seconds show the GUI with no clients attached to provide a baseline, at 10s the browser is opened.

Memory usage appears to be consistently starting from a *higher* start point *after* these changes. I have no idea why. Something about this branch is having an impact which is promising, but the stepped pattern of memory increase doesn't appear to have altered :(

dwsutherland:
> I am very ready for this to also be wrong, but hey ho.

Welcome to the club 😞

dwsutherland:
> Another issue mentioned there is attempting to send messages from the wrong thread (?), we should probs look into this just incase we are doing it without realising. Again, legitimate misuse.

I think the only threading is the ZeroMQ subscription to the schedulers (unless Tornado is doing something under the hood), and these just pump deltas into queues on the main thread (I believe).

oliver-sanders (approved):
Thanks for your tireless efforts on this one Chris, it was gnarly!

dwsutherland (approved):
LGTM - Regardless of the result (and there's some evidence of improvement), testing for a connection that isn't closing before send is sensible.

PR #746, opened 2025-10-29
services: validate-reinstall

### LINKED ISSUE #526: services: vr (validate-reinstall-apply)
Add support for the `cylc vr` command from the GUI.

This will allow users to more easily sync changes from workflow's working directories.

Note, the `cylc vr` command currently only operates on one workflow at a time, this will change with: https://github.com/cylc/cylc-flow/issues/5837

oliver-sanders:
Closes #526

I somehow forgot that we have `cylc vr` support in the Tui, but not the GUI!

Trivial to add support. Supporting the `cylc install` command is also possible, however, that will require #512.

UI sibling: https://github.com/cylc/cylc-ui/pull/2347

**Check List**

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -28,6 +28,7 @@ from subprocess import (
     DEVNULL,
     PIPE,
     Popen,
+    TimeoutExpired,
 )
 from textwrap import indent
 from time import time
@@ -309,6 +310,7 @@ class Services:
                 The application log, used to record this command invocation.
             timeout:
                 Length of time to wait for the command to complete.
+                The command will be killed if the timeout elapses.
             success_msg:
                 Message to be used in the response if the command succeeds.
             fail_msg:
@@ -344,7 +346,7 @@ class Services:
                     env.pop('CYLC_ENV_NAME', None)
                     env['CYLC_VERSION'] = cylc_version
 
-                # run cylc play
+                # run command
                 proc = Popen(
                     cmd,
                     env=env,
@@ -353,11 +355,21 @@ class Services:
                     stderr=PIPE,
                     text=True
                 )
-                ret_code = proc.wait(timeout=timeout)
+
+                try:
+                    ret_code = proc.wait(timeout=timeout)
+                except TimeoutExpired as exc:
+                    proc.kill()
+                    ret_code = 124  # mimic `timeout` command error code
+                    # NOTE: preserve any stderr that the command produced this
+                    # far as this may help with debugging
+                    out, err = proc.communicate()
+                    err = str(exc) + (('\n' + err) if err else '')
+                else:
+                    out, err = proc.communicate()
 
                 if ret_code:
                     msg = f"{fail_msg} ({ret_code}): {cmd_repr}"
-                    out, err = proc.communicate()
                     results[wflow] = err.strip() or out.strip() or msg
                     log.error(
                         f"{msg}\n"
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -219,31 +219,34 @@ class Services:
     # log file stream lag
     CAT_LOG_SLEEP = 1
 
+    #  command timeout for commands which start schedulers
+    START_TIMEOUT = 120
+
     @staticmethod
     def _error(message: Union[Exception, str]):
         """Format error case response."""
-        return [
+        return (
             False,
             str(message)
-        ]
+        )
 
     @staticmethod
     def _return(message: str):
         """Format success case response."""
-        return [
+        return (
             True,
             message
-        ]
+        )
 
     @classmethod
     async def clean(
         cls,
+        workflows_mgr: 'WorkflowsManager',
         workflows: Iterable['Tokens'],
         args: dict,
-        workflows_mgr: 'WorkflowsManager',
         executor: 'Executor',
         log: 'Logger'
-    ):
+    ) -> tuple[bool, str]:
         """Calls `cylc clean`"""
         # Convert Schema options → cylc.flow.workflow_files.init_clean opts:
         opts = _schema_opts_to_api_opts(args, schema=CleanOptions)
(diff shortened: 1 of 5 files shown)


cylc/uiserver/resolvers.py
```diff
@@ -273,25 +277,51 @@ async def scan(
...
     @classmethod
-    async def play(
+    async def run_command(
         cls,
+        command: Iterable[str],
         workflows: Iterable[Tokens],
         args: Dict[str, Any],
-        workflows_mgr: 'WorkflowsManager',
         log: 'Logger',
-    ) -> List[Union[bool, str]]:
```
oliver-sanders: Changed from `list[bool | str]` to `tuple[bool, str]` as this is safer for the way we are trying to use it (`('msg', False)` would have been acceptable before).

oliver-sanders:
FYI: style tests will fail until #748

wxtim (approved):
- [x] Read Code
 - [x] Tried in practice.

n.b. I think that VR should be on the shortened list of commands, but that's a Cylc UI change

cylc/uiserver/schema.py
```diff
@@ -241,6 +241,44 @@ class Arguments:
...
+            Validate, reinstall, then reload or restart as appropriate.
+
+            This command updates a workflow to reflect any new changes made in
+            the workflow source directory since it was installed.
+
+            The workflow will be reinstalled, then either:
+            * Reloaded (if the workflow is running),
+            * or restarted (if it is stopped).
+        ''')
+        # Note we have the "resume" mutation for paused workflows.
```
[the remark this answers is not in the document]
oliver-sanders: > Validate-reinstall will not play a paused workflow

It will play.

(comment copied from the play section)

```suggestion
```

wxtim (approved):

cylc/uiserver/schema.py
```diff
@@ -241,6 +241,48 @@ class Arguments:
...
+            This command updates a workflow to reflect any new changes made in
+            the workflow source directory since it was installed.
+
+            The workflow will be reinstalled, then either:
+            * Reloaded (if the workflow is running),
+            * or restarted (if it is stopped).
+        ''')
+        resolver = partial(mutator, command='validate_reinstall')
+
+    class Arguments:
```
MetRonnie: There are more arguments that this shares with `play` - worth extracting out into a class that can be inherited by both `Arguments` classes?

cylc/uiserver/tests/test_resolvers.py
```diff
@@ -169,37 +173,38 @@ async def test_play(
             [Tokens('wflow1')],
             [1],
             ("something", "bad things!!"),
-            "bad things!!",
-            ["Command failed (1): cylc play", "something", "bad things!!"],
+            r"bad things!!.*",
+            ["Command failed \(1\): cylc ", "something", "bad things!!"],
```
MetRonnie:
```suggestion
            [r"Command failed \(1\): cylc ", "something", "bad things!!"],
```

cylc/uiserver/resolvers.py
```diff
@@ -273,25 +277,51 @@ async def scan(
         cls,
         args: dict,
         workflows_mgr: 'WorkflowsManager',
-    ):
+    ) -> tuple[bool, str]:
         await workflows_mgr.scan()
         return cls._return("Scan requested")
 
     @classmethod
```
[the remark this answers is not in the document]
oliver-sanders: > As opposed to normal methods? 

I'm guessing you mean functions rather than methods (methods have the same drawbacks as classmethods).

The classmethods in this class make use of other properties, not just `_return` and `_error` but also `CAT_LOG_SLEEP` and `enqueue`. These things are a part of the class for namespacing (tidies related functionality).

We could spread this throughout the module, or move it all into another module (would would provide the same namespacing).

I tend to lean toward functions myself, but I don't think it really matters here.

oliver-sanders:
(drafting whilst I look at how hard it will be to combine ALL play and reload opts)

oliver-sanders (commented):
This is now the "full" solution which should support all relevant usages/args.

Reviewers, please give attention to the newly-added arguments to ensure they are working as intended:

https://github.com/cylc/cylc-uiserver/pull/746/changes#r2833158708

cylc/uiserver/resolvers.py
```diff
@@ -155,6 +156,33 @@ def _build_cmd(cmd: List, args: Dict) -> List:
...
+    Returns:
+        The provided dictionary post-modification.
+
+    Examples:
+        >>> _substitute_keys({'a': 1, 'b': 2}, [('b', 'c')])
+        {'a': 1, 'c': 2}
+
+        >>> _substitute_keys({'a': 1, 'b': 2}, [('b', 'c'), ('c', 'd')])
+        {'a': 1, 'd': 2}
+
```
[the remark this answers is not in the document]
oliver-sanders: Yes, they could be combined, but it's probs simpler to have one test to demonstrate each thing than to merge them.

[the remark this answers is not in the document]
wxtim: Happy to let that ride/

cylc/uiserver/schema.py
```diff
@@ -241,6 +242,86 @@ class Arguments:
...
+class Reinstall:
+    class Arguments:
+        workflows = graphene.List(WorkflowID, required=True)
+        cylc_version = CylcVersion(
+            description=sstrip('''
+                Set the Cylc version that the workflow starts with.
+            ''')
+        )
+        # cylc-rose
+        opt_conf_key = graphene.List(
```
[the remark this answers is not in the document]
oliver-sanders: The consequences of the opt config are written to `log/config/TIMESTAMP-rose-suite.conf`.

e.g:

```
cylc install -O '(foo)'
cylc reinstall <wflow> -O '(bar)'
```

The latest `rose-suite.conf` in that dir will read:

```
!opts=(foo) (bar)

[env]
# ROSE_ORIG_HOST set by cylc install.
ROSE_ORIG_HOST=...

[template variables]
# ROSE_ORIG_HOST set by cylc install.
ROSE_ORIG_HOST=...
```

The previous will read `!opts=(foo)`.

cylc/uiserver/schema.py
```diff
@@ -241,6 +242,86 @@ class Arguments:
...
 
 
+class Reinstall:
+    class Arguments:
+        workflows = graphene.List(WorkflowID, required=True)
+        cylc_version = CylcVersion(
+            description=sstrip('''
+                Set the Cylc version that the workflow starts with.
+            ''')
+        )
```
MetRonnie: I don't think this applies to reinstall, surely?
```suggestion
```

MetRonnie (changes requested):
I tried adding opt conf keys in the command form but they were not passed to the `cylc reinstall command`

cylc/uiserver/schema.py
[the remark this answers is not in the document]
oliver-sanders: If you made this change it would break.

The GraphQL field name must match the CLI field name, otherwise, we would have to maintain a mapping for each field / CLI option. Possible, but a maintenance pain.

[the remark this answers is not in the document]
MetRonnie: They are all plural in the CLI: `dest="opt_conf_keys"`, `dest="defines"`, 
`dest="rose_template_vars"`

https://github.com/cylc/cylc-flow/blob/520b04105ecdd2c330bc3f8aa48a635842720502/cylc/flow/option_parsers.py#L478-L511

Maybe that's why it isn't working correctly at the moment?

> I tried adding opt conf keys in the command form but they were not passed to the cylc reinstall command

oliver-sanders:
(tests should pass now)

cylc/uiserver/schema.py
```diff
@@ -241,6 +242,87 @@ class Arguments:
...
+                specifying `[template variables]FOO=1` to the `define`
+                argument.
+            ''')
+        )
+
+
+class ReinstallReload(graphene.Mutation):
+    class Meta:
+        description = sstrip('''
+            Validate, reinstall, then reload workflows.
```
MetRonnie, line 291: To match how other commands are written in the schema
```suggestion
            Validate, reinstall, then reload the workflow.
```

cylc/uiserver/schema.py
```diff
@@ -241,6 +242,87 @@ class Arguments:
...
+        """Compound command inherits arguments from the subcommands it proxies.
+        """
+
+    result = GenericScalar()
+
+
+class ReinstallRestart(graphene.Mutation):
+    class Meta:
+        description = sstrip('''
+            Validate, reinstall, then restart workflows.
```
MetRonnie, line 310:
```suggestion
            Validate, reinstall, then restart the workflow.
```

MetRonnie (commented):
Still have the duplicated cylc version option (https://github.com/cylc/cylc-uiserver/pull/746#discussion_r2860175782) but otherwise I think this is all good

cylc/uiserver/schema.py
oliver-sanders: This mutation can operate on multiple workflows right?

cylc/uiserver/schema.py
oliver-sanders: This mutation can operate on multiple workflows right?

MetRonnie: Yes, but singular would be consistent with all other mutations' descriptions

MetRonnie (approved):
👍

PR #721, opened 2025-08-21
fix UIS-Schd dump delta race condition

dwsutherland:
closes #700 

This fix increases the amount of time a delta waits for the initial dump, and carries on otherwise (relying on the data reconciliation/validation to redo the store if needed).

Hard to test without decreasing this wait time again, but can demonstrate it has done the job by running a highly parallel `cylc-flow` integration test with a UIS running.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [ ] Tests are included (or explain why tests are not needed).
- [x] Changelog entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

Issue #700, opened 2025-07-04
Apparent race condition between deltas and initial dump on start up


dwsutherland:
### Description

At NIWA, UI Server start-up (only) produces can produce the following log output:
```
.
.
.
[I 2025-07-04 01:30:09.293 CylcUIServer] [data-store] register_workflow('~ecox_oper/westport-inundation-forecast-nzcsm/run1', True
)
[I 2025-07-04 01:30:09.295 CylcUIServer] [data-store] connect_workflow('~ecox_oper/westport-inundation-forecast-nzcsm/run1', <dict
>)
[E 2025-07-04 01:30:11.220 CylcUIServer] 'family_proxies'
    Traceback (most recent call last):
      File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/cylc/uiserver/data_store_mgr.py", line 310, in _apply_all_delta
        if delta_time >= self.data[w_id]['delta_times'][field.name]:
    KeyError: 'family_proxies'
[E 2025-07-04 01:30:11.222 CylcUIServer] 'jobs'
    Traceback (most recent call last):
      File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/cylc/uiserver/data_store_mgr.py", line 310, in _apply_all_delta
        if delta_time >= self.data[w_id]['delta_times'][field.name]:
    KeyError: 'jobs'
[E 2025-07-04 01:30:11.222 CylcUIServer] 'task_proxies'
    Traceback (most recent call last):
      File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/cylc/uiserver/data_store_mgr.py", line 310, in _apply_all_delta
        if delta_time >= self.data[w_id]['delta_times'][field.name]:
    KeyError: 'task_proxies'
[E 2025-07-04 01:30:11.235 CylcUIServer] 'family_proxies'
    Traceback (most recent call last):
      File "/opt/niwa/share/cylc/8.4.3/lib/python3.9/site-packages/cylc/uiserver/data_store_mgr.py", line 310, in _apply_all_delta
        if delta_time >= self.data[w_id]['delta_times'][field.name]:
    KeyError: 'family_proxies'
[E 2025-07-04 01:30:11.236 CylcUIServer] 'jobs'
    Traceback (most recent call last):
[...]

PR #760, opened 2025-12-03
Fix long time for deltas to come through on workflow startup 

MetRonnie:
**Check List**

- [x] No dependency changes 
- [x] Docs not needed

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -93,8 +93,6 @@ class DataStoreMgr:
 
     """
 
-    INIT_DATA_WAIT_CHECKS = 30  # check attempts
-    INIT_DATA_RETRY_DELAY = 0.5  # seconds
     RECONCILE_TIMEOUT = 5.  # seconds
     PENDING_DELTA_CHECK_INTERVAL = 0.5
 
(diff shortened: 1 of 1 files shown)


cylc/uiserver/data_store_mgr.py
```diff
@@ -271,16 +268,6 @@ def _update_workflow_data(self, topic, delta, w_id):
             w_id (str): Workflow external ID.
 
         """
-        # Wait until data-store is populated for this workflow,
-        # carry on otherwise, errors will be reconciled with data validation.
-        if self.data[w_id][WORKFLOW].last_updated == 0:
-            loop_cnt = 0
-            while loop_cnt < self.INIT_DATA_WAIT_CHECKS:
```
dwsutherland: Hold on.. I just increased this check time recently, because waiting a shorter time was causing problems for larger workflows..

Are you sure we can just delete it?

dwsutherland: https://github.com/cylc/cylc-uiserver/pull/721

That contains two fixes:
- increase the time waiting for the initial dump
- make sure deltas get applied even if the store doesn't have a previous update time.

So what it will do is allow any deltas to come in and get applied, even if the initial dump hasn't been applied.. Which could mean they get overwritten by the dump (so we miss deltas)

MetRonnie: Sorry I was planning to tag you and ask about this. I have observed in testing a workflow started via the GUI that this `while` loop is blocking any deltas coming in for the maximum wait time and removing it leads to the tasks showing much faster in the GUI. I tried increasing `INIT_DATA_WAIT_CHECKS` to 120 so that it would wait 60s and it still does not receive what it expects. Removing it did not seem to cause any problems that I observed, and did not bring back the bug reported in #700 while running the cylc-flow integration tests at `-n 8` parallelisation.

Maybe there is some other fix to make this `while` loop work as intended? But at present it does not seem to be working correctly.

dwsutherland: fix up:
https://github.com/cylc/cylc-uiserver/pull/761

fixes this and the other long standing problem

Issue #221, opened 2021-05-25
graphql: delta fixes


oliver-sanders:
Address remaining delta issues:

* Added delta on workflow start (should be updated)
* Added delta on workflow reload (should be updated)
* Empty deltas

See https://github.com/cylc/cylc-uiserver/pull/212#issuecomment-841107713

dwsutherland:
I don't think it should be `updated`, the UI should rebuild it's store on:
- Start
- Reload
- UIS resync (due to failing validation)

Main reason being; there may be no way to reconcile the update with the old UI store  (i.e. the workflow changed it's structure, so historical nodes may never get a pruned delta)..
It's the same reason I don't update the scheduler store on reload... We can still show orphaned tasks (we don't at present, issue up somewhere), and change it's family hierarchy (by throwing it under root), but it's best start fresh on these rare occurrences.

oliver-sanders:
Ok, I was unaware that these extra added deltas were supposed to be signals to the UI to rebuild its store (I'm not sure whether the UI is aware of that either?). It's not all that obvious that `added(status:running)` actually means "delete everything and rebuild from scratch".

If we have to wipe the entire UI data store entry then the UI will have to clear all DOM elements relating to those entries meaning that on (re)start/reload the workspace would go blank before reloading and the user would loose all of their data, selection and position within the app (but the tabs would be preserved at least).

I'm not convinced that we should ever need to rebuild the UI store (other than in the event of validation failure at the UI end)? In the event of "start" and "reload" the UIS should be able to tell what's changed and send that in an updated delta? Or at least tell what items have changed and send a "pruned" and "added" delta for each? Can the UIS tell what's changed something like this?:

```python
pruned = set(before[taskProxies)) - set(after[taskProxies])
added = set(after[taskProxies]) - set(before[taskProxies))
updated = set(before[taskProxies)) & set(after[taskProxies])
```

If we do need to tell the UI to rebuild the store could we send an explicit signal to the UI telling it what to clear rather than leaving it to interpret the signal itself. Perhaps something along these lines:

```
# restart
pruned(taskProxies:*, jobs:*, tasks:*)
updated(status:running)
```

One problem that we are going need to address soon(ish) is that the GraphQL deltas will need to be used to provide both offline and live data. The user might be looking at cycles outside of the `n=x` window, viewing the static graph or job log files. In this case we wouldn't want the UI to delete all of that data, this would be basically equivalent to refreshing the browser tab.

[...]

PR #761, opened 2025-12-05
fix empty status and block on scheduler startup/init-sync

### LINKED ISSUE #485: running workflows showing with the "?" symbol
When starting a stopped workflow, it will sometimes show in the GUI with the "?" symbol rather than the "▶" icon for a running workflow.

From inspecting the UI data store, when the workflow is started, the status and statusMsg fields are initially provided correctly:

```
 status: "running", statusMsg: "running"
```

But are then quickly set to `""` in a subsequent delta:

```
status: "", statusMsg: ""
```

Then, this happens again leaving the workflow with status `""` causing the workflow-icon component to display the `?` and preventing mutations from being run against the workflow.

dwsutherland:
closes #485, closes #760
partially address #221

The empty status bug appeared to be cause by a contact file update/delta on connect:
image
image

This contact file update (and resulting delta) on connect is actually unnecessary, because all the fields it updates are also updated by the initial dump from the scheduler (or initial delta if the scheduler is still loading)..

If the scheduler is still loading, then the empty status would remain for the duration of a data-store initial checking loop..
This was a bug, as the checking loop expected a non-zero `updated_time`, however, a workflow starting up would produce an empty initial dump (hence zero `updated_time`, and empty status)

These bugs would give the appearance of a slow and uncertain scheduler startup..

Both of these related bugs are fixed here.

**Check List**

### DIFF
--- a/changes.d/761.fix.md
+++ b/changes.d/761.fix.md
@@ -1,1 +1,1 @@
-Fixed a bug causing empty status and data-store initiation delay on scheduler startup.
+Fixed a bug causing "unknown" status and delay to data appearing on workflow startup.


cylc/uiserver/data_store_mgr.py
```diff
@@ -180,9 +180,6 @@ async def connect_workflow(self, w_id, contact_data):
             self.log.info(f'failed to connect to {w_id}')
             self.disconnect_workflow(w_id)
             return False
-        else:
-            # don't update the contact data until we have successfully updated
-            self._update_contact(w_id, contact_data)
```
oliver-sanders: It seems like this was somehow responsible for delta with a blank status field.

But this is rather odd as the delta created by this method will only contain a status field if one is explicitly provided (which it isn't):

https://github.com/cylc/cylc-uiserver/blob/83934ec822e6be4983491fe10a7bbda4a410890e/cylc/uiserver/data_store_mgr.py#L473-L474

Any idea how this is happening?

If I'm understanding correctly, would adjusting `_update_contact` along the lines of this also work?

```diff
- if status is not None:
-     flow.status = status
+ flow.status = status or self.data[w_id].status  # untested
```

oliver-sanders: > Do we need to do stripNull=true for added too?

Doing this does not make sense, and should not be necessary, however, I suspect it *might* just work.

**Why this doesn't make sense:**

* An added delta is the data that creates/initialises an object in the store with its initial data (isn't really a delta, it's the full state).
* An updated delta is a true delta, containing a diff for the data store to apply.

We should receive one added delta, followed by a string of update deltas, followed by a prune delta.

**Why this might just work:**

Sadly, we had some issues with this due to the different lifecycle of schedulers versus the UI server.

From the scheduler's perspective, it produces an added delta when it starts up. The UIS, however, has been tracking the workflow since it installed. Unfortunately, the UIS did/does foreward the startup "added" delta through to GraphQL. As a result, we may receive, one added delta, followed by a chain of update deltas, followed by one added delta, followed by a chain of updated deltas, followed by a prune delta.

This is a long standing issue covered by https://github.com/cylc/cylc-uiserver/issues/221.

It's also why the UI used to frequently raise "added delta received after update" error messages when run in development mode.

To work around this, when the UI data store was re-written, it was done to treat added/updated equally and automatically create any missing objects.

changes.d/761.fix.md
```diff
@@ -0,0 +1 @@
+Fixed a bug causing empty status and data-store initiation delay on scheduler startup.
```
MetRonnie:
```suggestion
Fixed a bug causing "unknown" status and delay to data appearing on workflow startup.
```

PR #768, opened 2025-12-20
add graphql websocket subsciption tests

dwsutherland:
Adds an integration test for testing the graphql subscriptions via websocket connection.

**Check List**

### DIFF
--- a/cylc/uiserver/graphql/tornado.py
+++ b/cylc/uiserver/graphql/tornado.py
@@ -40,7 +40,6 @@ from typing import (
 from tornado import web
 from tornado.escape import json_encode
 from tornado.escape import to_unicode
-from tornado.httpclient import HTTPClientError
 from tornado.log import app_log
 from tornado.web import HTTPError
 
@@ -146,46 +145,35 @@ class TornadoGraphQLHandler(web.RequestHandler):
             self.handle_error(ex)
 
     async def run(self, *args, **kwargs):
-        try:
-            data = self.parse_body()
+        data = self.parse_body()
 
-            if self.batch:
-                responses = [
-                    await self.get_response(entry)
-                    for entry in data
-                ]
-                result = "[{}]".format(
-                    ",".join([response[0] for response in responses])
-                )
-                status_code = (
-                    responses
-                    and max(responses, key=lambda response: response[1])[1]
-                    or 200
-                )
-            else:
-                result, status_code = await self.get_response(data)
-
-            if status_code == 400:
-                self.set_status(status_code, reason=result)
-            else:
-                self.set_status(status_code)
-
-            self.set_header("Content-Type", "application/json")
-            self.write(result)
-            await self.finish()
-
-        except HTTPClientError as e:
-            response = e.response
-            response["Content-Type"] = "application/json"
-            response.content = self.json_encode(
-                {"errors": [self.format_error(e)]}
+        if self.batch:
+            responses = [
+                await self.get_response(entry)
+                for entry in data
+            ]
+            result = "[{}]".format(
+                ",".join([response[0] for response in responses])
             )
-            return response
+            status_code = (
+                responses
+                and max(responses, key=lambda response: response[1])[1]
+                or 200
+            )
+        else:
+            result, status_code = await self.get_response(data)
+
+        if status_code == 400:
+            self.set_status(status_code, reason=result)
+        else:
+            self.set_status(status_code)
+
+        self.set_header("Content-Type", "application/json")
+        self.write(result)
+        await self.finish()
 
     async def get_response(self, data):
-        query, variables, operation_name, _id = self.get_graphql_params(
-            self.request, data
-        )
+        query, variables, operation_name, _id = self.get_graphql_params(data)
 
         execution_result = await self.execute_graphql_request(
             data, query, variables, operation_name
@@ -253,10 +241,10 @@ class TornadoGraphQLHandler(web.RequestHandler):
                 request_json = json.loads(body)
                 if self.batch:
                     if not isinstance(request_json, list):
-                        raise AssertionError(
+                        raise AssertionError((
                             "Batch requests should receive a list"
                             ", but received {}."
-                        ).format(repr(request_json))
+                        ).format(repr(request_json)))
                     if len(request_json) <= 0:
                         raise AssertionError(
                             "Received an empty list in the batch request."
@@ -354,14 +342,14 @@ class TornadoGraphQLHandler(web.RequestHandler):
     async def execute(self, *args, **kwargs):
         return execute(*args, **kwargs)
 
-    def get_graphql_params(self, request, data):
+    def get_graphql_params(self, data):
         if self.graphql_params:
             return self.graphql_params
 
         single_args = {}
-        for key in request.arguments.keys():
+        for key in self.request.arguments.keys():
             single_args[key] = self.decode_argument(
-                request.arguments.get(key)[0])
+                self.request.arguments.get(key)[0])
 
         query = single_args.get("query") or data.get("query")
         variables = single_args.get("variables") or data.get("variables")
--- a/cylc/uiserver/graphql/tornado.py
+++ b/cylc/uiserver/graphql/tornado.py
@@ -235,7 +235,7 @@ class TornadoGraphQLHandler(web.RequestHandler):
             try:
                 body = self.request.body
             except Exception as e:
-                raise ExecutionError(400, e)
+                raise ExecutionError(400, [e])
 
             try:
                 request_json = json.loads(body)
@@ -337,7 +337,7 @@ class TornadoGraphQLHandler(web.RequestHandler):
 
             return result
         except Exception as e:
-            return ExecutionResult(errors=[e])
+            return ExecutionResult(data=None, errors=[e])
 
     async def execute(self, *args, **kwargs):
         return execute(*args, **kwargs)
--- a/cylc/uiserver/graphql/tornado.py
+++ b/cylc/uiserver/graphql/tornado.py
@@ -59,11 +59,7 @@ from graphql.execution.middleware import MiddlewareManager
 from graphql.pyutils import is_awaitable
 from graphql.validation import validate
 
-from cylc.flow.network.graphql import (
-    NULL_VALUE,
-    instantiate_middleware,
-    strip_null
-)
+from cylc.flow.network.graphql import instantiate_middleware
 
 if TYPE_CHECKING:
     from graphene import Schema
@@ -71,20 +67,6 @@ if TYPE_CHECKING:
 
 MUTATION_ERRORS_FLAG = "graphene_mutation_has_errors"
 MAX_VALIDATION_ERRORS = None
-
-
-def data_search_action(data, action):
-    if isinstance(data, dict):
-        return {
-            key: data_search_action(val, action)
-            for key, val in data.items()
-        }
-    if isinstance(data, list):
-        return [
-            data_search_action(val, action)
-            for val in data
-        ]
-    return action(data)
 
 
 def get_content_type(request: 'HTTPServerRequest') -> str:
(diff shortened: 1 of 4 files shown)


cylc/uiserver/graphql/tornado.py
```diff
@@ -407,24 +391,6 @@ async def execute_graphql_request(
...
-            else 0
-        )
-        json_priority = (
-            accepted_length - accepted.index("application/json")
-            if "application/json" in accepted
-            else 0
-        )
-
-        return html_priority > json_priority
-
```
oliver-sanders: [note to reviewers]

This is presumably something we inherited via graphene-tornado that's no longer required - https://github.com/graphql-python/graphene-tornado/blob/e4fa7d7b4c2256fa37f5ad89cbfd4d4bf6fdb606/graphene_tornado/tornado_graphql_handler.py#L379

Can't find any trace of it in Tornado.

oliver-sanders (commented):
LGTM, one question (see comment).

This pushes code coverage in the right direction :+1::

Screenshot from 2026-01-05 10-58-11

There are still a few gaps:

* Would be good to cover error handling, e.g, GraphQL validation errors.
* Would be good to cover more of the websocket subprocotocol (though note that [we will change subprotocol soon](https://github.com/cylc/cylc-ui/issues/1028))
* Would be good to cover connection termination (as this is something we've had issues with).

cylc/uiserver/graphql/tornado.py
```diff
@@ -92,25 +91,6 @@ def get_content_type(request: 'HTTPServerRequest') -> str:
...
-    raw_content_types = request.headers.get("Accept", "*/*").split(",")
-    qualified_content_types = map(qualify, raw_content_types)
-    return [
-        x[0]
-        for x in sorted(
-            qualified_content_types, key=lambda x: x[1], reverse=True)
-    ]
-
-
```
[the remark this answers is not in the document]
dwsutherland: Same as above

cylc/uiserver/tests/test_graphql.py
```diff
@@ -92,6 +155,76 @@ async def test_query(gql_query, dummy_workflow):
...
+    # perform the request
+    response = await gql_query(
+        *('cylc', 'graphql', 'batch'),
+        body=json.dumps(
+            [{'id': 1, 'query': query, 'operationName': 'fooFlow'}]
+        )
+    )
+    assert response.code == 200
+
+    # we should find the two dummy workflows in the response
```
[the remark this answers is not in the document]
dwsutherland: Yeah, this is a weird one.. Both `graphene-django` and `graphene-tornado` don't appear to be built to handle an actual batched response (or list of queries)...

I may have to change the code to "fix" this in accordance to the accepted behavior..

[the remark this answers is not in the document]
oliver-sanders: I more meant that the test code below doesn't seem to match the comment?

oliver-sanders:
83% project coverage :1st_place_medal: 

Ping me when you want a review.

dwsutherland:
> 83% project coverage 🥇
> 
> Ping me when you want a review.

Just need to bump up the websocket/subscription side of things first.

oliver-sanders:
(shifting this onto the 1.9.0 milestone as there are functional code changes)

oliver-sanders:
> (shifting this onto the 1.9.0 milestone as there are functional code changes)

FYI: We'll be looking to release 1.9.0 shortly once the new Cylc Review is in, would be great to merge this pre-release.

cylc/uiserver/tests/test_graphql.py
```diff
@@ -41,44 +45,112 @@ def gql_query(jp_fetch):
...
+        headers={'Content-Type': 'application/x-www-form-urlencoded'},
+        query=query,
+        method='GET'
     )
-    assert response.code == 200
+    assert get_response.code == 200
+
+    # test against same query different method
+    post_response = await gql_query(*('cylc', 'graphql'), query=query)
+    assert get_response.body == post_response.body
```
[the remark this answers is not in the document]
dwsutherland: So it appears that the only restriction is mutations have to be POST, it's the same with Django:
https://github.com/graphql-python/graphene-django/blob/f02ea337a23df06d166d463d6f92cb7505fbb435/graphene_django/views.py#L322-L337

I do test this..
```
    # try a mutation with GET method
    response = await gql_query(
        *('cylc', 'graphql'),
        headers={'Content-Type': 'application/x-www-form-urlencoded'},
        query='''
            mutation {
                stop(workflows: ["*"]) {
                    result
                }
            }
        ''',
        method='GET',
        raise_error=False
    )
    # Should be rejected for incorrect method.
    assert response.code == 405
    assert response.reason == 'Method Not Allowed'
```

So GET or POST for queries, only POST for mutations.. The POST just means we package the query into the body instead of URL..

So, "nothing to see here.."

dwsutherland:
> > (shifting this onto the 1.9.0 milestone as there are functional code changes)
> 
> FYI: We'll be looking to release 1.9.0 shortly once the new Cylc Review is in, would be great to merge this pre-release.

@oliver-sanders - This is ready to go.. (was a while ago but I thought the whole POST query would be more narly, turned out it was the norm amongst framework implementations of graphql)

(
seems coverage has dropped with Cylc Review and/or something else..
image

Maybe #797 and others ... But this PR does what it says and 100% on this patch... So follow-ons to get coverage up.
)

oliver-sanders:
(project coverage is just a target, you don't have to hit that number in this PR!)

dwsutherland:
> (project coverage is just a target, you don't have to hit that number in this PR!)

Yeah, there is one thing annoying me though.. The async gens are not being served:
image

(which doesn't seem to make sense, because how can the stuff at the beginning of `on_start` not be covered if the `try:` is hit ??)

I think I need to try push the subscriptions a little further.. Like you mentioned:
https://github.com/cylc/cylc-uiserver/pull/768#discussion_r2661092151

So will try get that covered at least

dwsutherland:
Not sure what's with coverage, that block is hit:
```
    async def on_start(self, connection_context, op_id, params):
        # Attempt to unsubscribe first in case we already have a subscription
        # with this id.

        print('HELLO')
        execution_result = await self.execute(params)
        iterator = None
        try:
            if isawaitable(execution_result):
                execution_result = await execution_result
            if not hasattr(execution_result, '__aiter__'):
                await self.send_execution_result(
                    connection_context, op_id, execution_result)
            else:
                print('THERE')
```
then forcing an assert error shows it in the stdout:
```
$ pytest cylc/uiserver/tests/test_graphql.py::test_subscription
.
.
.
--------------------------------- Captured stdout call ---------------------------------
HELLO
THERE
```
🤷

oliver-sanders:
(re-running CI on fresh merge to see if coverage reports change)

dwsutherland:
> (re-running CI on fresh merge to see if coverage reports change)

Looks like it's showing as covered now.. I don't want to be too finicky about `tornado_ws.py` with https://github.com/cylc/cylc-uiserver/pull/821
in bound..

So good to go as is, then I can build on it.. i.e. with #597

oliver-sanders:
(Codecov sometimes jitters, I think test re-runs can cause it, a fresh run fixes it)

oliver-sanders (approved):

cylc/uiserver/tests/test_handlers.py
```diff
@@ -20,17 +20,273 @@
...
+                    }
+                ),
+            ]
+        )
+
+    def _create_handler(
+        self,
+        handler_class=UIServerGraphQLHandler,
+        r_kwargs={},
+        h_kwargs={},
```
MetRonnie: Mutable default args. Are we not running flake8 on the tests?

MetRonnie (changes requested):

cylc/uiserver/tests/test_handlers.py
oliver-sanders: Ach, no, however, Scott has kindly synced the linting config from cylc-flow in https://github.com/cylc/cylc-uiserver/pull/823

oliver-sanders: #823 is now merged.

dwsutherland: Ah, yes, fixed.

cylc/uiserver/tests/test_handlers.py
```diff
@@ -19,17 +19,277 @@
...
+        handler = handler_class(
+            application=app,
+            request=request,
+            schema=schema,
+            **h_params,
+        )
+
+        if logged_in:
+            handler.get
+            _current_user = lambda: {'name': getuser()}
```
MetRonnie:
```suggestion
            handler.get_current_user = lambda: {'name': getuser()}
```

MetRonnie (changes requested):

MetRonnie:
Sorry David, I've spotted that https://github.com/cylc/cylc-uiserver/pull/823 didn't actually enable flake8 for the tests, please standby, pending https://github.com/cylc/cylc-uiserver/pull/842

MetRonnie:
Kicking tests

oliver-sanders:
@dwsutherland, have pushed some lint fixes to this branch in order to nudge this forward.

MetRonnie (commented):
LGTM but some of the tests are pretty slow, e.g. https://github.com/cylc/cylc-uiserver/blob/111aaebaf0d1f5347180362cd413f188a195496e/cylc/uiserver/tests/test_graphql.py#L432

takes >45s on my machine. Any idea why?

oliver-sanders:
Hmmm, that's concerning.

From a quick skim, `_update_workflow_data` will sleep for up to 15 seconds and is called by the test three times, so that's probably it.

MetRonnie:
Ack, yes. This was encountered before in https://github.com/cylc/cylc-uiserver/pull/790#discussion_r2847200683

oliver-sanders:
This sleep is pretty icky, if I remember correctly...

This sleep is there to ensure that the data store is populated before we yield any graphql updates. I.e, wait for the initial data store sync before yielding deltas.

Note, it's a `time.sleep` not an `asyncio.sleep` so presumably locks up the thread completely?!

We don't want to yield results before this point because key fields (namely status) will not be populated, causing problems with the UI.

So we wait for an arbitrary 15 seconds. If the data doesn't come in before that, then we send the partial delta anyway and let problems downstream materialise?

An alternative might be to set the workflow status as running/stopped (based on the presence or absence of the contact file), which is icky as the defaulted status could prove incorrect. Or to obtain the status via a direct request to the workflow (note, default comms timeout is PT5S) before issuing the data subscription?

If the sleep is hard to get rid of (and I suspect it might be), then we might be best off just hacking the sleep interval for tests.

oliver-sanders:
(the easiest solution is probably to `monkeypatch` out the `time.sleep` so that this doesn't affect testing)

oliver-sanders:
@dwsutherland, to help speed the PR along, this diff turns off any uses of `time.sleep` in the `data_store_mgr` which seems to do the trick (test time down from ~60s to ~20s at 4x parallelism).

```diff
diff --git a/cylc/uiserver/tests/conftest.py b/cylc/uiserver/tests/conftest.py
index e48a188..0f889be 100644
--- a/cylc/uiserver/tests/conftest.py
+++ b/cylc/uiserver/tests/conftest.py
@@ -280,6 +280,13 @@ def patch_conf_files(monkeypatch: pytest.MonkeyPatch):
     )
 
 
+@pytest.fixture(autouse=True)
+def remove_data_store_sleep(monkeypatch):
+    monkeypatch.setattr(
+        'cylc.uiserver.data_store_mgr.time.sleep', lambda x: None
+    )
+
+
 @pytest.fixture
 def cylc_uis(jp_serverapp):
     """Return the UIS extension for the JupyterServer ServerApp."""
```

dwsutherland:
> to help speed the PR along

Done and rebased

oliver-sanders:
Sleep redaction looks good to me, plz squash merge when happy.

MetRonnie (approved):

PR #755, opened 2025-11-26
Add Cylc Review (from Cylc 7)

wxtim:
Closes https://github.com/cylc/cylc-flow/issues/5937 and https://github.com/cylc/cylc-flow/issues/3441
Requires https://github.com/cylc/cylc-flow/pull/7068 to work correctly (else `cylc review` is labelled a dead-end).

Sort of requires https://github.com/cylc/release-actions/pull/136 To run coverage later.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).

Test working can be demonstrated by adding this to CI. 
```
          pip uninstall cylc-flow -y
          pip install git+https://github.com/wxtim/cylc@feat.cylc-review
```
https://github.com/cylc/cylc-uiserver/actions/runs/19763672460/job/56631293468

- [x] Changelog entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #805, opened 2026-04-10
review: move into own namespace package

oliver-sanders:
Follow up to #755

* Cylc Review has been ported into this repository as a stopgap measure.
* The long-term plan is to develop cylc-uiserver/cylc-ui until it possesses all of the functionalities/capabilities of Cylc Review.
* At which point Cylc Review will become deprecated and will eventually become removed.
* To facilitate this, Cylc Review has been moved into its own namespace package to help prevent its code becoming intertwined with UIS code.

To test:

```console
$ pip install -e .
$ cylc review start
```

It should behave identically as on master (note, you need to `pip install -e` every time you toggle between this branch and master)

Additionally:
* Run `isort`.
* Fix `annotated_types` import and missing dependency. 

**Check List**

- [x] Changelog entry included if this is a change that can affect users - unreleased code so no change


cylc/review/scripts/review.py
```diff
@@ -26,9 +26,14 @@
 import signal
 
 from cylc.flow.option_parsers import CylcOptionParser as COP
-from cylc.uiserver.review import CylcReviewService
```
oliver-sanders, line 29: I ran `isort` at the same time as making these changes, hence the import diff.

oliver-sanders:
Project coverage has gone down to ~22% which is an error, need to configure something...

oliver-sanders:
Config fixed, coverage back up to ~70%.

Note, the project coverage is still failing here, as it is on master, we've got some more test work to do before we reach our coverage target...

.coveragerc
```diff
@@ -30,14 +30,12 @@ omit =
     cylc/uiserver/tests/*
     cylc/uiserver/jupyter*_config.py
 parallel = True
-plugins=
-include=
-    cylc/uiserver/*
 timid = False
+source = ./cylc
```
oliver-sanders, line 34: Resolves a warning which was previously printed for every `cylc` command invocation during the Cylc Review tests.

.coveragerc
```diff
@@ -30,14 +30,12 @@ omit =
...
+relative_files=True
```
oliver-sanders, line 35: This along with the `coverage combine --append` fixes an issue where the coverage from the cylc.uiserver and cylc.review tests were not being combined. The proof of this working is in the project coverage score.

.github/workflows/test.yml
```diff
@@ -117,7 +117,7 @@ jobs:
 
       - name: Coverage report
         run: |
-          coverage combine .
+          coverage combine --append
```
oliver-sanders, line 120: `.` is implicit, `--append` is not.

Without the `--append` it was throwing away previous coverage data.

ChrisPaulBennett (approved):
Looks good

samuel-denton:
Everything seems to be working.  

I have found 1 slight difference to current cylc review, where some old workflows are appearing at the beginning of the list rather than in date order, but I think this is due to those workflows having been broken in some way and no longer properly reporting their metadata.  I have been unable to clean those same workflows for a few weeks, so I think this is something unique to my setup.  It's possible I've accidentally moved or deleted one of their associated directories.
I am trying to reproduce this currently, but I suspect it's unrelated to this PR.

oliver-sanders:
> I have found 1 slight difference to current cylc review, where some old workflows are appearing at the beginning of the list rather than in date order

Difference against the "site-deployed" Cylc Review, or the version on master?

samuel-denton:
> > I have found 1 slight difference to current cylc review, where some old workflows are appearing at the beginning of the list rather than in date order
> 
> Difference against the "site-deployed" Cylc Review, or the version on master?

Site-Deployed.  I haven't been able to reproduce this issue, I suspect those workflows are broken in some way, maybe I deleted some files without doing a clean first or something like that.  They do appear on the site-deployed cylc review:
image
Just at the bottom.  The version of cylc review from this PR puts them at the top.  I.E. for months I have been able to ignore those workflows but if it puts them at the top I keep being reminded to try and fix them :D

I was going to approve but figured I would get your opinion on those stuck workflows first.

oliver-sanders:
> The version of cylc review from this PR puts them at the top

I doubt it has anything to do with this PR, you want to test against master rather than the site-deployed as they are wildly different.

cylc/review/ws.py
```diff
@@ -161,7 +165,7 @@ def get_util_home(*args):
 
 
 def get_review_service_config(
-    port: Annotated[int, Ge(0)] = 8042,
+    port: 'Annotated[int, Ge(0)]' = 8042,
     service_root: str = 'services/cylc'
```
samuel-denton, line 169: This confused me and I had to look it up, but it appears you might want to 'stringify' an annotated type if there are import conflicts at runtime? 

"If an import isn’t available at runtime, quoting prevents Python from trying to evaluate it.  It’s just stored as a string and only interpreted later (e.g. by type checkers)."

I don't follow this well enough to confirm there are no issues caused by this change.  In my head it might alter how some tests work but thats a complete guess. 

@oliver-sanders if you can confirm you understand this change and it's implications I can approve the PR else I need to do a bit more reading and better understand what this function actually does and what the change might do.

samuel-denton: Thank you copilot:
"Ge is only imported under TYPE_CHECKING (which is False at runtime). Without the string quotes, Python would try to evaluate Annotated[int, Ge(0)] at runtime and raise a NameError because Ge doesn't exist. Wrapping it in quotes ('Annotated[int, Ge(0)]') makes it a string literal that Python stores as-is in __annotations__ without evaluating it."

samuel-denton: I've checked myself and agree there doesn't seem to be any tests relying on this type annotation.  (Copilot seems really good at finding tests for specific bits of code):

Image

samuel-denton (approved):
Looks good.

samuel-denton:
This looks bugged to me unless I don't understand what it's saying :D
image

I think the code coverage is ~71%, not sure what the 'patch 100%' means or if its a bug.

Given the last few PR's have been merged with similar coverage, is this one ok to merge as is @oliver-sanders?

EDIT: I guess 'patch 100%' is saying all the code added / changed in this PR is covered?

oliver-sanders:
Really not sure about that. Taking a look at codecov, the project coverage was ~71% before and remains ~71% after, so I think this is a reporting artifact?

cylc/review/ws.py
oliver-sanders: The reason we "stringify" type hints is to avoid having to import the thing unnecessarily.

The `TYPE_CHECKING` variable is always `False` unless a type checker (e.g, `mypy`) is in use, in which case it's `True`.

I.E, this:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
   from some_big_module import Whatevs

def foo(x: 'Whatevs'):
    ...
```

Is considered to be better than this:

```python
from typing import TYPE_CHECKING
from some_big_module import Whatevs

def foo(x: Whatevs):
    ...
```

Because it avoids explicitly importing `some_big_module`. It may be imported somewhere else, but at least we're not *just* importing it for the purpose of type checking.

Shemse1 (commented):
done

.coveragerc
Shemse1: cylc/review/static/js/lang-n.js

PR #823, opened 2026-05-20
sync linting configuration from cylc-flow

### LINKED ISSUE #809: sync linting from cylc-flow
Cylc Flow uses these Flake8 plugins:

https://github.com/cylc/cylc-flow/blob/f731469a46da66469af360b8ed2bf185111f59e8/setup.cfg#L100-L109

With a couple of rule exceptions:

https://github.com/cylc/cylc-flow/blob/master/tox.ini

Synchronise these plugins and exceptions to cylc-uiserver and fix any issues raised.

Note, you may find the `ruff check --select<code> --fix` command comes in handy here.

Crosslink:
* https://github.com/cylc/cylc-rose/issues/424
* https://github.com/cylc/cylc-uiserver/issues/809
* https://github.com/metomi/rose/issues/3034

Scott-Owen-James:
Syncing linting, closes #809 

**Check List**

### DIFF
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -446,10 +446,10 @@ class CylcUIServer(ExtensionApp):
             # pick the highest installed version by default
             try:
                 version = self._list_ui_versions(build_dir)[-1]
-            except IndexError as exc:
+            except IndexError:
                 raise Exception(
                     f'Could not find any UI builds in {build_dir}.'
-                ) from exc
+                ) from None
 
         ui_path = build_dir / version
         if (ui_path / 'index.html').exists():
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -446,10 +446,10 @@ class CylcUIServer(ExtensionApp):
             # pick the highest installed version by default
             try:
                 version = self._list_ui_versions(build_dir)[-1]
-            except IndexError:
+            except IndexError as exc:
                 raise Exception(
                     f'Could not find any UI builds in {build_dir}.'
-                )
+                ) from exc
 
         ui_path = build_dir / version
         if (ui_path / 'index.html').exists():
--- a/cylc/uiserver/graphql/tornado.py
+++ b/cylc/uiserver/graphql/tornado.py
@@ -300,7 +300,7 @@ class TornadoGraphQLHandler(web.RequestHandler):
             try:
                 body = self.request.body
             except Exception as e:
-                raise ExecutionError(400, e)
+                raise ExecutionError(400, e) from e
 
             try:
                 request_json = json.loads(body)
@@ -322,11 +322,11 @@ class TornadoGraphQLHandler(web.RequestHandler):
                 self.parsed_body = request_json
                 return self.parsed_body
             except AssertionError as e:
-                raise HTTPError(status_code=400, log_message=str(e))
-            except (TypeError, ValueError):
+                raise HTTPError(status_code=400, log_message=str(e)) from e
+            except (TypeError, ValueError) as e:
                 raise HTTPError(
                     status_code=400, log_message="POST body sent invalid JSON."
-                )
+                ) from e
 
         elif content_type in [
             "application/x-www-form-urlencoded",
@@ -441,11 +441,11 @@ class TornadoGraphQLHandler(web.RequestHandler):
         if variables and isinstance(variables, str):
             try:
                 variables = json.loads(variables)
-            except Exception:
+            except Exception as e:
                 raise HTTPError(
                     status_code=400,
                     log_message="Variables are invalid JSON."
-                )
+                ) from e
 
         operation_name = (
             single_args.get("operationName") or data.get("operationName")
--- a/cylc/uiserver/graphql/tornado.py
+++ b/cylc/uiserver/graphql/tornado.py
@@ -300,7 +300,7 @@ class TornadoGraphQLHandler(web.RequestHandler):
             try:
                 body = self.request.body
             except Exception as e:
-                raise ExecutionError(400, e) from None
+                raise ExecutionError(400, e) from e
 
             try:
                 request_json = json.loads(body)
--- a/cylc/uiserver/graphql/tornado.py
+++ b/cylc/uiserver/graphql/tornado.py
@@ -300,7 +300,7 @@ class TornadoGraphQLHandler(web.RequestHandler):
             try:
                 body = self.request.body
             except Exception as e:
-                raise ExecutionError(400, e) from e
+                raise ExecutionError(400, e) from None
 
             try:
                 request_json = json.loads(body)
--- a/cylc/uiserver/jupyter_config.py
+++ b/cylc/uiserver/jupyter_config.py
@@ -25,7 +25,7 @@ from cylc.uiserver import (
 from cylc.uiserver.app import USER_CONF_ROOT
 from cylc.uiserver.authorise import CylcAuthorizer
 
-
+c = {}
 # the command the hub should spawn (i.e. the cylc uiserver itself)
 c.Spawner.cmd = ['cylc', 'hubapp']
 
--- a/cylc/uiserver/jupyter_config.py
+++ b/cylc/uiserver/jupyter_config.py
@@ -25,7 +25,6 @@ from cylc.uiserver import (
 from cylc.uiserver.app import USER_CONF_ROOT
 from cylc.uiserver.authorise import CylcAuthorizer
 
-c = {}
 # the command the hub should spawn (i.e. the cylc uiserver itself)
 c.Spawner.cmd = ['cylc', 'hubapp']
 
--- a/cylc/uiserver/profilers.py
+++ b/cylc/uiserver/profilers.py
@@ -122,7 +122,7 @@ def get_profiler(profiler: str):
         return None
     try:
         return PROFILERS[profiler]
-    except KeyError as exc:
+    except KeyError:
         raise Exception(
             f'Unknown profiler: {profiler}'
             f'\nValid options: {", ".join(PROFILERS)}'
--- a/cylc/uiserver/profilers.py
+++ b/cylc/uiserver/profilers.py
@@ -126,4 +126,4 @@ def get_profiler(profiler: str):
         raise Exception(
             f'Unknown profiler: {profiler}'
             f'\nValid options: {", ".join(PROFILERS)}'
-        ) from exc
+        ) from None
--- a/cylc/uiserver/profilers.py
+++ b/cylc/uiserver/profilers.py
@@ -122,8 +122,8 @@ def get_profiler(profiler: str):
         return None
     try:
         return PROFILERS[profiler]
-    except KeyError:
+    except KeyError as exc:
         raise Exception(
             f'Unknown profiler: {profiler}'
             f'\nValid options: {", ".join(PROFILERS)}'
-        )
+        ) from exc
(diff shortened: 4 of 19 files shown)


tox.ini
```diff
@@ -1,24 +1,41 @@
...
-    tests,
     .git,
     __pycache__,
     .tox,
-    .eggs,
-    cylc/uiserver/jupyter_config.py,
-    *.svg,
-    ./src
+    **graphql_subscribe.py,
+    **data_messages_pb2.py
```
oliver-sanders: This is the list of files to exclude from linting.

It is specific to each project, e.g, in cylc-flow there's a file called `graphql_subscribe.py`, and in cylc-uiserver there's a file called `jupyter_config.py`.

Restore this bit how it was before, it should fix a bunch of problems.

cylc/uiserver/jupyter_config.py
```diff
@@ -25,7 +25,7 @@
 from cylc.uiserver.app import USER_CONF_ROOT
 from cylc.uiserver.authorise import CylcAuthorizer
 
-
+c = {}
```
oliver-sanders: This file is supposed to be excluded from linting:

```suggestion
```

cylc/uiserver/tests/test_workflows_mgr.py
```diff
@@ -34,7 +34,9 @@
     WorkflowsManager,
 )
 
-from .conftest import AsyncClientFixture
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    from .conftest import AsyncClientFixture
```
oliver-sanders, line 39: Anything within this `if TYPE_CHECKING` block will only be run if the module is loaded by a type-checker tool such as mypy.

Other tools (like regular Python) might fail due to this not being imported.

The workaround is to put the thing you've imported in quotes ahywhere it might cause problems.

E.g:

```python
def my_function(thinggy: 'AsyncClientFixture'`): pass
```

cylc/uiserver/tests/test_handlers.py
```diff
@@ -55,7 +54,10 @@ def _create_handler(self, logged_in=True):
 
         if logged_in:
             handler.get
-            _current_user = lambda: {'name': getuser()}
```
oliver-sanders, line 58: Well this is embarrassing.

This test code does not appear to have done anything.

I'm not sure what the code should be changed to, need to look this through...

cylc/uiserver/graphql/tornado.py
```diff
@@ -300,7 +300,7 @@ def parse_body(self):
             try:
                 body = self.request.body
             except Exception as e:
-                raise ExecutionError(400, e)
+                raise ExecutionError(400, e) from e
```
oliver-sanders, line 303: Question to reviewers.

Should these be `from e` or `from None`?

Do we want to preserve the primitive exception?

cylc/uiserver/tests/test_gui.py
```diff
@@ -76,18 +88,17 @@
...
-    existing_content,
-    workflow_id,
-    expected_updated_content,
-    hub_url,
-    mock_glbl_cfg):
+        existing_content,
+        workflow_id,
+        expected_updated_content,
+        hub_url,
+        mock_glbl_cfg):
```
oliver-sanders:
```suggestion
def test_update_html_file_updates_gui_file(
    existing_content,
    workflow_id,
    expected_updated_content,
    hub_url,
    mock_glbl_cfg
):
```

cylc/uiserver/tests/test_gui.py
```diff
@@ -76,18 +88,17 @@
...
     """Tests url is updated correctly"""
-    mock_glbl_cfg('cylc.uiserver.scripts.gui.glbl_cfg', 
-        f'''[hub]
-        url = {hub_url}
-        ''')
+    mock_glbl_cfg('cylc.uiserver.scripts.gui.glbl_cfg',
+                  f'''[hub]
+                  url = {hub_url}
+                  ''')
```
oliver-sanders: (preferable not to mix indentation styles)

```suggestion
    mock_glbl_cfg(
        'cylc.uiserver.scripts.gui.glbl_cfg',
        f'''[hub]
        url = {hub_url}
        '''
    )
```

cylc/uiserver/app.py
```diff
@@ -446,10 +446,10 @@ def _get_ui_path(self):
             # pick the highest installed version by default
             try:
                 version = self._list_ui_versions(build_dir)[-1]
-            except IndexError:
+            except IndexError as exc:
                 raise Exception(
                     f'Could not find any UI builds in {build_dir}.'
-                )
+                ) from exc
```
oliver-sanders: This `raise from <X>` feature is a relatively obscure one. It determines the way tracebacks appear in the terminal.
* If you add `from exc`, then you'll get a double traceback (the original traceback and another one for the new exception).
* If you add `from None`, then only the new exception `raise Exception(f'Could not ...'` will appear.

Generally, when we catch one exception only to raise another, the reason we are doing this is to prove a more user-friendly error message, so typically we want `from None` for a cleaner traceback, but it's contextual.

```suggestion
                ) from None
```

tox.ini
Scott-Owen-James: Made change

cylc/uiserver/tests/test_workflows_mgr.py
Scott-Owen-James: Corrected issue

cylc/uiserver/tests/test_gui.py
Scott-Owen-James: Change implemented.

cylc/uiserver/tests/test_gui.py
Scott-Owen-James: Change implemented.

cylc/uiserver/app.py
Scott-Owen-James: Change implemented.

cylc/uiserver/tests/conftest.py
```diff
@@ -215,7 +215,7 @@ def _write(**kwargs):
         conf = kwargs
 
     def _read(self):
-        nonlocal conf
+        conf
```
oliver-sanders:
```suggestion
```

cylc/uiserver/tests/conftest.py
```diff
@@ -433,14 +433,15 @@ def test_something(mock_glbl_cfg):
     """
     # TODO: modify Parsec so we can use StringIO rather than a temp file.
     def _mock_glbl_cfg(pypath: str, global_config: str) -> None:
-        nonlocal tmp_path, monkeypatch
+        tmp_path
+        monkeypatch
```
oliver-sanders:
```suggestion
```

cylc/uiserver/tests/conftest.py
```diff
@@ -433,14 +433,15 @@ def test_something(mock_glbl_cfg):
...
         global_config_path = tmp_path / 'global.cylc'
         global_config_path.write_text(global_config)
         glbl_cfg = ParsecConfig(SPEC, validator=cylc_config_validate)
         glbl_cfg.loadcfg(global_config_path)
 
         def _inner(cached=False):
-            nonlocal glbl_cfg
+            glbl_cfg
```
oliver-sanders:
```suggestion
```

cylc/uiserver/tests/test_handlers.py
```diff
@@ -55,7 +54,10 @@ def _create_handler(self, logged_in=True):
...
+
+            def current_user(name):
+                if name == 'name':
+                    return getuser()
```
oliver-sanders: Have traced this back to the commit where it became broken: https://github.com/cylc/cylc-uiserver/commit/91baadbceca3c7885169be9315ca5ce30971c5c3#diff-af7ba6b6a77e527c4b921c2e667f65da8f60bf738b3ea36df1100f094f59ec88R56

```suggestion
            def _current_user():
                return {'name': getuser()}

            handler.get_current_user = _current_user
```

cylc/uiserver/tests/conftest.py
Scott-Owen-James: Change made.

cylc/uiserver/tests/conftest.py
Scott-Owen-James: Change made.

cylc/uiserver/tests/conftest.py
Scott-Owen-James: Change made.

cylc/uiserver/graphql/tornado.py
Scott-Owen-James: Generally tried to keep consistent across the board in regard to when and when not to keep this, let me know if you want any I missed or you think need changing.

cylc/uiserver/tests/test_handlers.py
Scott-Owen-James: Dealt with.

cylc/uiserver/profilers.py
```diff
@@ -122,8 +122,8 @@ def get_profiler(profiler: str):
         return None
     try:
         return PROFILERS[profiler]
-    except KeyError:
+    except KeyError as exc:
         raise Exception(
             f'Unknown profiler: {profiler}'
             f'\nValid options: {", ".join(PROFILERS)}'
-        )
+        ) from exc
```
oliver-sanders:
```suggestion
        ) from None
```

cylc/uiserver/graphql/tornado.py
```diff
@@ -300,7 +300,7 @@ def parse_body(self):
             try:
                 body = self.request.body
             except Exception as e:
-                raise ExecutionError(400, e)
+                raise ExecutionError(400, e) from None
```
oliver-sanders:
```suggestion
                raise ExecutionError(400, e) from e
```

oliver-sanders (approved):
LGTM.

A couple of suggestions for exception raising.

Plz squash-merge.

cylc/uiserver/tests/test_handlers.py
```diff
@@ -55,7 +54,10 @@ def _create_handler(self, logged_in=True):
 
         if logged_in:
             handler.get
```
MetRonnie: Looks like this was supposed to be `handler.get_current_user = ...` (now correct below) - should remove this line that isn't doing anything
```suggestion
```

tox.ini
```diff
@@ -1,13 +1,33 @@
...
+    ; not all exception subclass __init__ args passed to super __init__
+    ; see https://github.com/cylc/cylc-flow/issues/7059
+    B042
+
+per-file-ignores=
+    ; TYPE_CHECKING block suggestions
+    ; B904 needs explicitly excepting because it's added in extend-select
+    tests/*: A, B, B904, C, E731, TC, ISC, N, SIM
+
+    ; for clarity we don't merge 'with Conf():' context trees
```
MetRonnie: This is referring to code that only exists in cylc-flow, and was also referring to a specific rule but the exclusions were broadened since then and the comment must not have been updated/removed
```suggestion
```

MetRonnie (changes requested):

cylc/uiserver/tests/test_handlers.py
oliver-sanders: Addresed in https://github.com/cylc/cylc-uiserver/pull/823#discussion_r3372254872

MetRonnie: Not fully addressed yet, hence my suggestion

MetRonnie (approved):
Many thanks 👌

PR #831, opened 2026-06-10
Cherry pick #823

MetRonnie:
I realised after merging #823 it was against master. But we're probably going to end up in a pickle if we don't have the change on 1.9.x


oliver-sanders:
It's not really appropriate to backport this, I'm not sure how much of a pickle you think this will cause, not too concerned.

MetRonnie:
We're going to end up with lots of lint failures in sync PRs

oliver-sanders:
We've only got 1 PR open for 1.9.x, so not too worried about that.

MetRonnie:
But anyway this is very low risk, no?

oliver-sanders:
There are functionalish changes to the source and no justification for those changes ending up in a bugfix.

If it was just test changes, that would be another matter.

MetRonnie:
Literally the only changes to source code are adding `... from exc/None` to raised exceptions, which has no real functional effect.

PR #840, opened 2026-06-15
Review: support alternate home location.

### LINKED ISSUE #837: review: support alternate home dirs
A minimal change to support viewing symlinked run dirs directly instead of via $HOME.

This issue was transferred from https://github.com/cylc/cylc-flow/pull/6875 which was built against the Cylc 7 code.

@hjoliver is this still needed?

hjoliver:
As described on the original PR here: https://github.com/cylc/cylc-flow/pull/6875

Close #837

**Check List**

### DIFF
--- a/cylc/review/review.py
+++ b/cylc/review/review.py
@@ -23,12 +23,13 @@ With no arguments, the status of the ad-hoc web service server is printed.
 
 For 'cylc review start', if 'PORT' is not specified, port 8080 is used.
 
-If $CYLC_REVIEW_HOME is defined look there for user run directories instead
-of in home directories. This can be used to view symlinked run directories
-directly when home directories are private. For example with symlinking as:
-   /home/USER/cylc-run/WORKFLOW -> /project/PROJECT/USER/cylc-run/WORKFLOW
-set the environment variable as:
-   CYLC_REVIEW_HOME=/project/PROJECT
+With --alt-home-dir=DIR look under DIR for user run directories instead
+of in $HOME. This can be used to view symlinked run directories directly
+when home directories are private. For example with symlinking as:
+   /home/$USER/cylc-run/<workflow> -> /project/test/$USER/cylc-run/<workflow>
+start Review like this:
+   cylc review start --alt-home-dir=/project/test
+
 """
 
 import contextlib
@@ -111,9 +112,10 @@ class CylcReviewService:
         'rose-suite.conf'
     ]
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, alt_home_dir, *args, **kwargs):
         self.exposed = True
-        self.suite_dao = CylcReviewDAO()
+        self.alt_home_dir = alt_home_dir
+        self.suite_dao = CylcReviewDAO(alt_home_dir)
         self.logo = os.path.basename(
             get_util_home("doc", "src", "cylc-logo.png"))
         self.title = self.TITLE
@@ -373,7 +375,7 @@ class CylcReviewService:
         # Set list of task states depending on Cylc version 7 or 8
         task_statuses_ordered = CYLC7_TASK_STATUSES_ORDERED
         suite_dir = os.path.join(
-            self._get_user_home(user),
+            self._get_user_home(user, self.alt_home_dir),
             "cylc-run",
             suite)
         is_c8 = self.is_cylc8(suite_dir)
@@ -947,8 +949,8 @@ class CylcReviewService:
         return path
 
     @staticmethod
-    def _get_user_home(user):
-        """Return user home dir or $CYLC_REVIEW_HOME if defined.
+    def _get_user_home(user, alt_home_dir=None):
+        """Return user home directory under alt_home_dir or $HOME.
 
         N.B. os.path.expanduser does not fail if ~user is invalid.
 
@@ -956,20 +958,17 @@ class CylcReviewService:
             cherrypy.HTTPError(404)
 
         """
-        if os.environ.get('CYLC_REVIEW_HOME', False):
-            return os.path.join(
-                os.environ['CYLC_REVIEW_HOME'],
-                str(user)
-            )
+        if alt_home_dir is not None:
+            return os.path.join(alt_home_dir, str(user))
         try:
             return pwd.getpwnam(user).pw_dir
         except KeyError:
             raise cherrypy.HTTPError(404) from None
 
-    def _get_user_suite_dir_root(self, user):
+    def _get_user_suite_dir_root(self, user, alt_home_dir=None):
         """Return, e.g. ~user/cylc-run/ for a cylc suite."""
         return self._check_dir_access(
-            os.path.join(self._get_user_home(user), "cylc-run"))
+            os.path.join(self._get_user_home(user, alt_home_dir), "cylc-run"))
 
     @staticmethod
     def _check_string_for_path(string):
@@ -1115,7 +1114,7 @@ class CylcReviewService:
         for path in paths:
             self._check_path_normalised(path)
         suite_dir = os.path.join(
-            self._get_user_home(user),
+            self._get_user_home(user, self.alt_home_dir),
             "cylc-run",
             suite)
         if not paths:
@@ -1126,6 +1125,7 @@ class CylcReviewService:
             # >>> os.path.join('/foo', '/bar')
             # '/bar'
             raise cherrypy.HTTPError(403)
+
         return self._check_dir_access(path)
 
     @staticmethod
--- a/cylc/review/review.py
+++ b/cylc/review/review.py
@@ -21,7 +21,15 @@ logs via an HTTP interface.
 
 With no arguments, the status of the ad-hoc web service server is printed.
 
-For 'cylc review start', if 'PORT' is not specified, port 8080 is used."""
+For 'cylc review start', if 'PORT' is not specified, port 8080 is used.
+
+If $CYLC_REVIEW_HOME is defined look there for user run directories instead
+of in home directories. This can be used to view symlinked run directories
+directly when home directories are private. For example with symlinking as:
+   /home/USER/cylc-run/WORKFLOW -> /project/PROJECT/USER/cylc-run/WORKFLOW
+set the environment variable as:
+   CYLC_REVIEW_HOME=/project/PROJECT
+"""
 
 import contextlib
 from fnmatch import fnmatch
@@ -940,7 +948,7 @@ class CylcReviewService:
 
     @staticmethod
     def _get_user_home(user):
-        """Return, e.g. ~/cylc-run/ for a cylc suite.
+        """Return user home dir or $CYLC_REVIEW_HOME if defined.
 
         N.B. os.path.expanduser does not fail if ~user is invalid.
 
@@ -948,6 +956,11 @@ class CylcReviewService:
             cherrypy.HTTPError(404)
 
         """
+        if os.environ.get('CYLC_REVIEW_HOME', False):
+            return os.path.join(
+                os.environ['CYLC_REVIEW_HOME'],
+                str(user)
+            )
         try:
             return pwd.getpwnam(user).pw_dir
         except KeyError:
--- a/cylc/review/review_dao.py
+++ b/cylc/review/review_dao.py
@@ -43,13 +43,10 @@ TASK_STATUS_GROUPS = {
 }
 
 
-def get_prefix(user_name):
-    """Return user "home" dir under $CYLC_REVIEW_HOME, or else ~user_name."""
-    if os.environ.get('CYLC_REVIEW_HOME', False) and user_name:
-        prefix = os.path.join(
-            os.environ['CYLC_REVIEW_HOME'],
-            str(user_name)
-        )
+def get_prefix(user_name, alt_home_dir=None):
+    """Return user home dir under alt_home_dir or else ~user_name."""
+    if alt_home_dir is not None and user_name:
+        prefix = os.path.join(alt_home_dir, str(user_name))
     else:
         prefix = "~"
         if user_name:
@@ -129,7 +126,8 @@ class CylcReviewDAO:
     # Because that method needs self.TABLE_BROADCAST_STATES
     TABLE_BROADCAST_STATES = CylcWorkflowDAO.TABLE_BROADCAST_STATES
 
-    def __init__(self):
+    def __init__(self, alt_home_dir=None):
+        self.alt_home_dir = alt_home_dir
         self.daos = {}
 
     def _db_init(self, user_name, suite_name):
@@ -148,7 +146,7 @@ class CylcReviewDAO:
             for name in [os.path.join("log", "db"), "cylc-suite.db"]:
                 db_f_name = os.path.expanduser(
                     os.path.join(
-                        get_prefix(user_name),
+                        get_prefix(user_name, self.alt_home_dir),
                         os.path.join("cylc-run", suite_name, name)
                     )
                 )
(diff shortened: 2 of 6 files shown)


cylc/review/review.py
```diff
@@ -20,7 +20,15 @@
...
 
-For 'cylc review start', if 'PORT' is not specified, port 8080 is used."""
+For 'cylc review start', if 'PORT' is not specified, port 8080 is used.
+
+If $CYLC_REVIEW_HOME is defined look there for user run directories instead
+of in home directories. This can be used to view symlinked run directories
+directly when home directories are private. For example with symlinking as:
+   /home/USER/cylc-run/WORKFLOW -> /project/PROJECT/USER/cylc-run/WORKFLOW
+set the environment variable as:
+   CYLC_REVIEW_HOME=/project/PROJECT
```
oliver-sanders: We might want to consider configuring this via a CLI/API option rather than an env var.

As part of porting cylc-review to cylc-uisever, the `get_review_service_config` function has been created as a sort of configuration object:

https://github.com/cylc/cylc-uiserver#review

This function currently has two arguments for configuring the `port` and `service_root` that Cylc Review will use:

https://github.com/cylc/cylc-uiserver/blob/557d03e76bc1709c87052e09edbfcd157d0fa154/cylc/review/ws.py#L167-L187

It would be handy if this new option could be specified there rather than requiring an additional configuration.

hjoliver: Good idea. I've started on this as we need it pretty badly here.

cylc/review/review.py
```diff
@@ -939,24 +949,26 @@ def _check_dir_access(cls, path):
...
             return pwd.getpwnam(user).pw_dir
         except KeyError:
             raise cherrypy.HTTPError(404) from None
 
-    def _get_user_suite_dir_root(self, user):
+    def _get_user_suite_dir_root(self, user, alt_home_dir=None):
         """Return, e.g. ~user/cylc-run/ for a cylc suite."""
         return self._check_dir_access(
-            os.path.join(self._get_user_home(user), "cylc-run"))
+            os.path.join(self._get_user_home(user, alt_home_dir), "cylc-run"))
```
oliver-sanders, line 971: This is a method so can access `self.alt_home_dir` directly rather than as optional argument (avoids the potential for this arg to be missed off by accident).

cylc/review/scripts/review.py
```diff
@@ -21,6 +21,16 @@
...
+
+For 'cylc review start', if 'PORT' is not specified, port 8080 is used.
+
+With --alt-home-dir=DIR look under DIR for user run directories instead
+of in $HOME. This can be used to view symlinked run directories directly
+when home directories are private. For example with symlinking as:
+   /home/$USER/cylc-run/<workflow> -> /project/test/$USER/cylc-run/<workflow>
+start Review like this:
+   cylc review start --alt-home-dir=/project/test
+
```
oliver-sanders, line 33: You might kick me for only raising this now....

But, we could potentially get `cylc review` to look for workflows in the directory configured by `[install][install target]run` (defaulting to `~/cylc-run` as usual).

cylc/review/review.py
```diff
@@ -939,24 +949,26 @@ def _check_dir_access(cls, path):
         return path
 
     @staticmethod
-    def _get_user_home(user):
-        """Return, e.g. ~/cylc-run/ for a cylc suite.
+    def _get_user_home(user, alt_home_dir=None):
```
oliver-sanders, line 952: Note to reviewers:

This staicmethod is used elsewhere (review_dao) where it does not have access to the instance, so cannot be converted to a method.

cylc/review/review_dao.py
```diff
@@ -134,11 +146,16 @@ def _db_init(self, user_name, suite_name):
             for name in [os.path.join("log", "db"), "cylc-suite.db"]:
```
oliver-sanders, line 146: `prefix` now unused?

cylc/review/scripts/review.py
hjoliver: All good, I'll consider that.  Somewhat relatedly, this is not gonna be usable as a hub service at ESNZ any which way, need to think about that. I'll comment on Element ...

oliver-sanders:
Drafting this in line with https://github.com/cylc/cylc-uiserver/pull/840#discussion_r3772272967

hjoliver:
@oliver-sanders - 

> But, we could potentially get cylc review to look for workflows in the directory configured by [install][install target]run (defaulting to ~/cylc-run as usual).

I don't think this idea will work for us.  The problem is at ESNZ some users work in multiple projects, with symlink-dirs determined on the fly according to `$PROJECT` in the environment. 

On the other hand this PR as-is will work nicely with the jupyter server proxy approach, because users can configure a separate Review command per project (with different `--alt-home-dir`) and they'll all appear as separate launchers in Jupyter Lab - easy.

So if you've no objection I'll undraft this and finish it off.

oliver-sanders:
With the Jupyter Server Proxy approach, Cylc Review would always be run *as* the user (just like the UIS), so would always be able to access a read-only `$HOME`, removing the need for alternate run dirs.

So I don't see how this would help?

hjoliver:
> With the Jupyter Server Proxy approach, Cylc Review would always be run as the user (just like the UIS), so would always be able to access a read-only $HOME,

Yes, but:

> , removing the need for alternate run dirs.

No, because the point of this PR is to allow a user to see other project members' run dirs as well as their own. 

When viewing our symlinked project locations, a Review service running as me can also see the run dirs of all other project members.  

Each user has to start their own service, but that service can see all their project colleagues too.

Obviously that's not as efficient as a single service per project, say, but that doesn't really matter. My review instance runs in a hub launched PBS allocation and will die when the PBS jobs ends. And no privileged service user is needed.

hjoliver:
Update from offline convo with @oliver-sanders - 

My approach does work as described (tested), but it requires some user config per project, and this PR.

The approach he's suggesting (hub spawns Review as the target user) will work without user config or this PR, with the proviso that manual URL manipulation is needed for the moment - but we can fix that later. And there may be a path to eventual UIS integration.

Upshot: I'll test his approach tomorrow and abandon this PR if it's a goer.

hjoliver:
OK, sigh, yesterday I had forgotten some of my thoughts on this from a couple of weeks back when first discussed.

In current circumstances, here, my approach - with this PR - is feasible; but I can't yet use the other at all.

(A) My approach (I start my own Review server, pointed at a symlinked shared project space) works now (with this PR, obvs)
 - does not require any hub config 
     - (I don't have control over our hub config) 
 - does not even require me to be able to start other users' servers 
      - (enabled for our ops hub, but not the research one due to unresolved concerns about external users)
 - leverages existing filesystem sharing permissions that our admins like
     - (my server can see all member run dirs, and no one else's)
 - I can configure a different review server per project, if I belong to multiple projects

(B) The other approach (the hub starts the target user's Review server for me, pointed at their home dir)
- Requires the hub to allow me to start other users' servers
  - (see above: still not enabled for our research hub)
- Requires significant hub config (RBAC) to ensure I can only start the servers of colleagues in the same projects as me
  - (admins not happy with me being able to read all hub users run dirs)

I agree that (B) is better in principle - but (A) allows sharing workflow logs easily with project members now, despite the fact that our researchers can't even use cross-user UIS auth yet (ops can),

oliver-sanders:
Ok, I don't quite understand the situation at your end.

Just to make you aware, you don't actually have to mainline this functionality in order to use it.

Here's a simple shim which encompasses the functionality in this PR:

```python
import os.path
from pathlib import Path
import sys

from cylc.review.review import CylcReviewService
import cylc.review.review_dao
from cylc.review.scripts.review import get_option_parser
from cylc.review.ws import _ws_init


ALT_RUN_DIR = '/var/tmp/tmp.CAEuQjoRVa'


def _get_user_home(*args):
    return os.path.join(ALT_RUN_DIR, args[-1])


def _expanduser(path):
    path = Path(path).expanduser()
    parts = path.parts
    if 'cylc-run' in parts:
        return str(
            Path(
                ALT_RUN_DIR,
                parts[parts.index('cylc-run') :],
            )
        )
    return str(path)


CylcReviewService._get_user_home = _get_user_home
cylc.review.review_dao.expanduser = _expanduser


parser = get_option_parser()
opts, args = parser.parse_args([], sys.argv)

_ws_init(
    service_cls=CylcReviewService,
    port=opts.port,
    service_root=opts.service_root,
    single_user=opts.single_user,
)
```

(Shouldn't actually need that expanduser shim, that code should use the same `get_user_home`)

Rather than configuring Jupyter Hub to run `cylc review start --port...`, configure it to call `python <path-to-shim> --port...`.

Example:

```console
$ cd $(mktemp -d)
$ mkdir $USER/cylc-run
$ cp -r ~/cylc-run/<my-workflow> $USER/cylc-run/
$ python <path-to-shim> --port=8000
$ firefox 0.0.0.0:8000
```

hjoliver:
> Ok, I don't quite understand the situation at your end. 

More explanation given in offline chat - hopefully that did it.

hjoliver:
I like the shim idea, thanks.

It needed some tweaking:

There seems to be no `opts.single_user` arg - what's that about?

The `expanduser` replacement doesn't work (it needs to replace `<module>.os.path.expanduser` not `<module>.expanduser`) but it still *seems* to work if your `$HOME/cylc-run` is readable because it still gets the logs from there.

Also had call the original `os.path.expanduser`  instead of `Path.expanduser` inside the shim function, to avoid a recursion bomb when `Path.expanduser` calls `os.path.expanduser` which we replaced...

I'll post the working version to the chat later, with a question about how best to invoke it...

oliver-sanders:
> There seems to be no opts.single_user arg - what's that about?

Whoops, I tried to snip out a part of a larger diff.

I had started looking into how to handle the user-switching problem for your deployment pattern where it would be necessary to switch servers in order to view other user's workflows.

Easily doable, though I failed to get the JS part working quickly.

```diff
diff --git a/cylc/review/review.py b/cylc/review/review.py
index cb2ad95..c506af5 100644
--- a/cylc/review/review.py
+++ b/cylc/review/review.py
@@ -25,6 +25,7 @@ For 'cylc review start', if 'PORT' is not specified, port 8080 is used."""
 
 import contextlib
 from fnmatch import fnmatch
+from getpass import getuser
 from glob import glob
 import json
 import mimetypes
@@ -44,13 +45,13 @@ import traceback
 from urllib.parse import quote
 
 import cherrypy
+from cylc.flow import __version__ as CYLC_VERSION
+from cylc.flow.task_state import TASK_STATUSES_ORDERED
+from cylc.flow.workflow_files import WorkflowFiles
 import jinja2
 from jinja2 import select_autoescape
 import markupsafe
 
-from cylc.flow import __version__ as CYLC_VERSION
-from cylc.flow.task_state import TASK_STATUSES_ORDERED
-from cylc.flow.workflow_files import WorkflowFiles
 from cylc.review.review_dao import (
     TASK_STATUS_GROUPS,
     CylcReviewDAO,
@@ -103,7 +104,7 @@ class CylcReviewService:
         'rose-suite.conf'
     ]
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, single_user=True, **kwargs):
         self.exposed = True
         self.suite_dao = CylcReviewDAO()
         self.logo = os.path.basename(
@@ -111,6 +112,7 @@ class CylcReviewService:
         self.title = self.TITLE
         self.host_name = gethostname()
         self.cylc_version = CYLC_VERSION
+        self.single_user = single_user
         # Autoescape markup to prevent code injection from user inputs.
         template_env = jinja2.Environment(
             loader=jinja2.PackageLoader(
@@ -129,6 +131,9 @@ class CylcReviewService:
     @cherrypy.expose
     def index(self, form=None):
         """Display a page to input user ID and suite ID."""
+        if self.single_user:
+            return self.suites(user=getuser())
+
         data = {
             "logo": self.logo,
             "title": self.title,
diff --git a/cylc/review/scripts/review.py b/cylc/review/scripts/review.py
index 19b7b1f..63d78d3 100644
--- a/cylc/review/scripts/review.py
+++ b/cylc/review/scripts/review.py
@@ -21,6 +21,26 @@ Start/stop ad-hoc Cylc Review web service server for browsing users' suite
 logs via an HTTP interface.
 
 With no arguments, the status of the ad-hoc web service server is printed.
+
+Cylc Review can be started in one of two modes:
+  single-user (default):
+    Display the workflows of the user that Cylc Review is run under.
+    This is best for standalone installations and is suitable for deployment
+    using Jupyter Server Proxy - where one Cylc Review server will be run.
+    per-user.
+
+  multi-user:
+    Display the workflows of all users who's home directories are readable by
+    the account that Cylc Review is run under. This may work for multi-user
+    networks with liberal filesystem permissions. It is suitable for deployment
+    using Jupyter Hub services - where one Cylc Review service will be run
+    per-Cylc Hub deployment.
+
+The difference between these modes is mostly cosmetic. In multi-user mode, the
+homepage becomes a form for selecting the user who's workflows are to be
+viewed, whereas in single-user mode, the homepage displays the workflows for
+the user it is running under.
+
 """
 
 import os
@@ -88,13 +108,32 @@ def get_option_parser():
         default='/',
         dest="service_root",
     )
+    parser.add_option(
+        "--single-user",
+        help="Run Cylc Review in single-user mode (this is the default).",
+        action="store_true",
+        default=True,
+        dest='single_user'
+    )
+    parser.add_option(
+        "--multi-user",
+        help="Run Cylc Review in multi-user mode.",
+        action="store_false",
+        dest='single_user'
+    )
+
     return parser
 
 
 @cli_function(get_option_parser, remove_opts=IRRELEVANT_OPTS)
 def main(_, opts, *args):
     """Start/Stop the Cylc Review server."""
+    import sys
+    print(f'$ {sys.argv}')
     subcmd = args[0] if args else ''
+    print(f'{args=}')
+    print(f'{subcmd=}')
+    pass
 
     # Get current server status:
     status = _get_server_status(CylcReviewService)
@@ -105,6 +144,7 @@ def main(_, opts, *args):
             service_cls=CylcReviewService,
             port=opts.port,
             service_root=opts.service_root,
+            single_user=opts.single_user,
         )
     # User has asked to stop or get info on server, server _not_ running:
     elif not status:
diff --git a/cylc/review/static/js/cylc-review.js b/cylc/review/static/js/cylc-review.js
index 8efc45a..8fc2f58 100644
--- a/cylc/review/static/js/cylc-review.js
+++ b/cylc/review/static/js/cylc-review.js
@@ -139,3 +139,20 @@ $(function() {
         $("input.task_status[value=waiting]").prop("checked", false);
     });
 });
+
+function sitchUser(from, to, jupythub_base) {
+    to = to.trim()
+    if (!to) {
+        alert('Please enter a user name')
+    }
+
+    if (jupythub_base, String(window.location).includes(from)) {
+        window.location.replace(String(window.location).replace(to, from))
+    }
+
+    const url = new URL('/suites', location.origin);
+    url.searchParams.set('user', to);
+  alert(url)
+    // location.assign(url);
+    window.location.replace(url)
+}
diff --git a/cylc/review/template/suites.html b/cylc/review/template/suites.html
index 2c58552..2ab0069 100644
--- a/cylc/review/template/suites.html
+++ b/cylc/review/template/suites.html
@@ -35,10 +35,24 @@ rel="stylesheet" media="screen" />
         {{title}} <small>@ {{host}}</small>
       </a>
     </div>
+
     <div id="navbar" class="navbar-collapse collapse">
-      <ul class="nav navbar-nav">
-        <li class="active"><a href="#"><strong>{{user}}</strong></a></li>
-      </ul>
+      <form
+        class="navbar-form navbar-left"
+        onsubmit="sitchUser('{{user}}', this.username.value, '{{jupythub_base}}')"
+      >
+        <div class="form-group">
+          <input
+            id="username"
+            type="text"
+            class="form-control"
+            placeholder="user name"
+            value="{{user}}"
+          />
+        </div>
+        <button type="submit" class="btn btn-default">Go</button>
+      </form>
+
       <ul class="nav navbar-nav navbar-right">
         <li>
           <span class="navbar-text">
diff --git a/cylc/review/ws.py b/cylc/review/ws.py
index 5c19837..fb0c180 100644
--- a/cylc/review/ws.py
+++ b/cylc/review/ws.py
@@ -181,9 +181,12 @@ def get_review_service_config(
     return {
         "name": "cylc-review",
         "command": [
-            "cylc", "review", "start",
+            "cylc",
+            "review",
+            "start",
             f"--port={port}",
             f"--service-root={service_root}",
+            "--multi-user",
         ],
         "url": f"http://0.0.0.0:{port}/",
     }
```

PR #847, opened 2026-06-18
review: make robust to late arrival of DB file.

oliver-sanders:
* If Cylc Review attempted to connect to a DB before it appeared on the filesystem, it would fall back to the Cylc 7 DB path.
* This would not self-rectify as the decision would be remembered for future requests.
* This changes the logic so the DB file location is only cached once it has been determined to exist.

Note, this is just a quick patch to allow Cylc Review to recover. The situation should be handled better!

**Check List**

### DIFF
--- a/changes.d/847.fix.md
+++ b/changes.d/847.fix.md
@@ -1,0 +1,1 @@
+Cylc Review: fixed a bug where a newly-installed workflow would be stuck with an error banner incorrectly mentioning Cylc 8.0.x.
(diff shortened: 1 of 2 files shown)


samuel-denton (approved):
~~I was only able to reproduce the issue on site review not locally so I can't directly confirm it works, but the code changes read as though they should solve the immediate problem of caching DB locations without confirming first that they exist.~~
I managed to reproduce this locally and this fix does what I think it's indented to do.  It does not prevent CYLC trying to connect to a DB before it exists, so installed workflows still get the warning, but the db location is not cached so once the workflow is started, review can recover itself and find the correct db location.

MetRonnie (approved):
Tested, all good

MetRonnie:
Added a changelog entry

oliver-sanders:
Critical issue resolved, but will continue this in follow-on PRs.

PR #848, opened 2026-06-19
review: remove somewhat broken 8.0.x back-compat

oliver-sanders:
Cylc Review had some back-compat for 8.0.x workflows, now long outside of the support window.

This support wasn't quite working as intended, its error message would show in other circumstances than the ones described.

**Check List**

### DIFF
--- a/cylc/review/review_dao.py
+++ b/cylc/review/review_dao.py
@@ -21,7 +21,6 @@ import os
 from pathlib import Path
 import re
 import sqlite3
-from sqlite3 import OperationalError
 import tarfile
 
 from cylc.flow.rundb import CylcWorkflowDAO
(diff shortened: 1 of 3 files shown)


cylc/review/template/taskjobs.html
```diff
@@ -181,15 +181,6 @@
 <div class="container-fluid">
   <div class="row">
```
samuel-denton:
```suggestion
  <div class="row">
    <div class="col-md-11">
```

cylc/review/review_dao.py
samuel-denton:
```suggestion
```
I think this can go too

cylc/review/template/taskjobs.html
samuel-denton: This suggestion diff isn't quite working but I just mean to remove the newline.

oliver-sanders:
```suggestion
```

cylc/review/review_dao.py
oliver-sanders: Do you mean the `CANNOT_JOIN_FLOW_NUMS` lines above?

samuel-denton: Yeah not sure what happened to that suggested diff.  I don't think that variable is needed.

oliver-sanders: Agreed.

samuel-denton (approved):
LGTM

ChrisPaulBennett (approved):
LGTM

ChrisPaulBennett:
@samuel-denton Can you resolve the conversations (If they have been resolved that is the suggested diff's make it hard to tell what it is you wanted done)

PR #849, opened 2026-06-19
Fix flaky coverage failures

MetRonnie:
Something about the relative paths coverage setting was causing flaky failures of `coverage combine`. See https://github.com/coveragepy/coveragepy/issues/2195

**Check List**

- [x] Nothing else applicable

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -32,7 +32,6 @@ omit =
 parallel = True
 timid = False
 source = ./cylc
-relative_files=True
 
 
 [report]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,8 +117,7 @@ jobs:
 
       - name: Coverage report
         run: |
-          coverage combine --append
-          coverage xml --ignore-errors
+          coverage xml
           coverage report
 
       - name: Upload coverage artifact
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -117,8 +119,7 @@ jobs:
 
       - name: Coverage report
         run: |
-          coverage combine --append
```
MetRonnie, line 120: > As of version 7.14.0, files are combined by the [reporting commands](https://coverage.readthedocs.io/en/7.14.1/commands/cmd_reporting.html#cmd-reporting), so there is less need to use an explicit combine command.

https://coverage.readthedocs.io/en/7.14.1/commands/cmd_combine.html#cmd-combine

.coveragerc
```diff
@@ -32,7 +32,6 @@ omit =
 parallel = True
 timid = False
 source = ./cylc
-relative_files=True
```
oliver-sanders, line 35: I think this is no longer necessary after the `CYLC_COVERAGE_BASE` variable was added.

.github/workflows/test.yml
oliver-sanders: FYI: we run the same combine command in the cylc-flow functional tests, so I don't think this will affect stability?

oliver-sanders (approved):
Coverage seems unchanged.

Issue #845, opened 2026-06-18
restarted workflow not detected


oliver-sanders:
We've seen a situation where a restarted workflow was not detected by the UIS.

* Workflow was running on server 1.
* UIS had detected and connected to this workflow.
* Server 1 crashed, taking out the workflow.
* Workflow was restarted on server 2.
* GUI failed to detect the crash/restart stating only "Connection timed out".

The UIS should have spotted the server/port had changed and unregistered, registered, then connected to the workflow anew.

dwsutherland:
We've actually had something different, but similar .. Where the subscription between UIS and scheduler is broken, and the UIS has no detection and recovery mechanism (i.e. testing the connection/sub, and re-establishing subscription and/or cleaning up)..

If it had such a mechanism, it would probably deal with this problem also.. Because a crashed workflow has a contact file, so scan still thinks it's up, then when it's recovered on server 2 the UIS doesn't know anything has changed.

dwsutherland:
I've been able to replicate the problem by:
- `kill -9` the workflow process (leaving the contact file in place.
- restarting the workflow between scans (so the scan doesn't notice the contact file missing)
- releasing the workflow

Then the workflow doesn't update in the WUI, as the UIS has not ditched the previous sub:

<img width="1573" height="694" alt="Image" src="https://github.com/user-attachments/assets/6877d668-8b37-4095-b489-d8a0dc7bc0d9" />

Will have a look at ways to monitor whether a socket/pub is still alive..

dwsutherland:
I have a fix for this .. will post tonight/tomorrow (once finished)

dwsutherland:
Fix up:
https://github.com/cylc/cylc-uiserver/pull/851

No tests yet.. Can replicate the above more easily by increasing the scan interval.

Issue #781, opened 2026-01-14
service for workflow lifecycle management


oliver-sanders:
A UIS / Hub service for managing workflow runs.

At present, Cylc workflows are manually started, manually monitored throughout the life of the workflow, manually updated (`cylc vr`) when changes are made, until they are manually stopped when no longer needed.

While this approach will continue for high-stakes deployments where such oversight is desirable/necessary, today's increasingly automated work patterns don't always fit this model. It would be convenient if Cylc itself could provide some level of automation over the management of the lifecycle of workflows to better suit CD working patterns for lower-stakes, more hands-off deployments.

For example, a workflow's lifecycle might look like this:

1. Checkout a working copy of a workflow (`git clone ... ~/cylc-src/my-workflow`).
2. Start it (`cylc vip`).
3. Update working copy with changes (`git pull upstream/main`).
4. Apply changes (`cylc vr`).
5. Stop workflow (`cylc stop`).

An alternative, approach to this might be:

1. Register a workflow repository with the UIS.
  * Working copy automatically checked out.
  * Repository automatically monitored for changes, and updated live.
  * Workflow automatically restarted if it were ever to go down (e.g, critical infrastructure failure).
2. Unregister the workflow with the UIS.
  * Workflow automatically stopped, run archived and working copy removed.

This could potentially be orchestrated via a REST API (convenient for making one-off changes), or via a configuration file (convenient for more infrastructure as code type solutions), e.g:

```yaml
workflows:
  - repo: github.com/my-company/my-workflow
    name: local
    options:
      template variables:
        - mode: operational
        - archive: true
        - domain: regional-atmos

  - repo: github.com/my-company/my-workflow
    name: global-atmos
    options:
      template variables:
        - mode: operational
        - archive: true
        - domain: global

  - repo: github.com/user/research-workflow
    name: test/trial
    options:
      optional configs:
         - mode-research
         - site-my-site
         - grid-n1234
```

In conventional deployments, the lifecycle of the deployment is effectively coupled to the lifecycle of the VMs. In more cloudy deployments, lifecycle is coupled to that of the containers it orchestrates, the longest-lived of which is the Hub/UIS.

As such, the Hub/UIS becomes the top-level service (the thing the people running the orchestration need to ensure is always running when needed) making it the more natural place for such functionalities to lie.

<table>
<thead>
<tr>
<td>implementation</td>
<td>pros</td>
<td>cons</td>
</tr>
</thead>
<tbody>
<tr>
[...]

PR #851, opened 2026-06-23
Include active broken-connection/irresponsive-workflow checks

### LINKED ISSUE #857: detect network issues and disconnect/reconnect as necessary
It [has been suggested](https://github.com/cylc/cylc-uiserver/pull/851#issuecomment-4839325586) that there may be scenarios where UIS-Flow communications might need to be reset as the result of network outages.

I think ZMQ should attempt to re-establish the connection on the next publish event, so I'm not sure whether we need to disconnect/reconnect the subscriber explicitly at our layer, but am unsure over the scenario.

dwsutherland:
closes #857

There were two problems: 
- ~~DB restarts don't change the `CYLC_WORKFLOW_UUID`, and this was the criteria for reestablishing the UIS workflow.~~
- ~~old contact files of inactive workflows left an old subscription/workflow in the UIS.~~

~~These are addressed here by using PID as an alternate criteria,~~ and testing the PUB tcp listener of workflows (which will help clear old contact files and/or reestablish subscriptions)

**Update:**

I have an idea for another check, however, it would need some back compat added.. It would essentially check both the REQ/RES PUB/SUB are still connected/working.

I would create a new endpoint on the `cylc-flow` side that would take a REQ, and pass it back via the PUB. We would then pick that up via the existing UIS SUB, and record the timestamp. This would happen periodically (say every ~5min?), and any workflow with a last recorded time older than some threshold (say 11min) would be disconnected/unregistered/purged, and re-registered on the next scan (if available).

Why? Because it's good to check the existing connections (not just create new ones), and also another problem I've seen:
- Some workflows that are inactive for large periods of time (i.e. ones that run every month) stop receiving updates, as if the connections have timed out or failed in periods of inactivity..

Although I could investigate whether ZeroMQ has some kind of default inactivity timeout on connections, I think this will act as a check for any other connection disruptions.

I will crack on with it.

**Tasks:**

- [x] ~~Add criteria checks (`PID`, `HOST`) in addition to `UUID`.~~
- [x] Check both the REQ/RES PUB/SUB are still connected/working.

Sibling PR https://github.com/cylc/cylc-flow/pull/7355

**Check List**

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -34,6 +34,7 @@ Subscriptions are currently run in a different thread (via ThreadPoolExecutor).
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
+import json
 from pathlib import Path
 import time
 from typing import (
@@ -312,10 +313,19 @@ class DataStoreMgr:
             return
         elif (
             topic == 'ping'
-            and delta == b'pong'
             and w_id in self.workflows_mgr.workflows
         ):
             self.workflows_mgr.workflows[w_id]['pubsub_time'] = time.time()
+            # workflow is now clearly responsive
+            if w_id in self.workflows_mgr.irresponsive:
+                # Bring status up to date
+                status_data = json.loads(delta.decode('utf-8'))
+                self.create_status_delta(
+                    w_id,
+                    status=status_data['status'],
+                    status_msg=status_data['status_msg'],
+                )
+                self.workflows_mgr.irresponsive.remove(w_id)
             return
         self._apply_all_delta(w_id, delta)
         self._delta_store_to_queues(w_id, topic, delta)
@@ -486,11 +496,16 @@ class DataStoreMgr:
             # workflow has been removed - do nothing
             return False
 
-        delta = DELTAS_MAP[ALL_DELTAS]()
-        delta.workflow.time = time.time()
+        delta = self.create_status_delta(
+            w_id,
+            status=status,
+            status_msg=status_msg,
+            apply=False,
+            send=False,
+        )
+
         flow = delta.workflow.updated
-        flow.id = w_id
-        flow.stamp = f'{w_id}@{delta.workflow.time}'
+
         if contact_data:
             # update with contact file data
             flow.name = contact_data['name']
@@ -507,10 +522,6 @@ class DataStoreMgr:
             flow.port = 0
             flow.api_version = 0
 
-        if status is not None:
-            flow.status = status
-        if status_msg is not None:
-            flow.status_msg = status_msg
         if pruned:
             flow.pruned = True
             delta.workflow.pruned = w_id
@@ -523,6 +534,48 @@ class DataStoreMgr:
         self._delta_store_to_queues(w_id, ALL_DELTAS, delta)
 
         return True
+
+    def create_status_delta(
+        self,
+        w_id,
+        status=None,
+        status_msg=None,
+        apply=True,
+        send=True,
+    ):
+        """Create status delta for application and/or distribution.
+
+        Args:
+            w_id: Workflow ID.
+            status: Workflow status (e.g. "running").
+            status_msg: Workflow status message (e.g. "will stop at 2000").
+            apply: Apply delta to data-store.
+            send: Add delta to subscription queue.
+
+        Returns:
+            An ALL_DELTAS with updated.workflow fields status and/or msg.
+
+        """
+
+        delta = DELTAS_MAP[ALL_DELTAS]()
+        delta.workflow.time = time.time()
+        flow = delta.workflow.updated
+        flow.id = w_id
+        flow.stamp = f'{w_id}@{delta.workflow.time}'
+
+        if status is not None:
+            flow.status = status
+        if status_msg is not None:
+            flow.status_msg = status_msg
+
+        # apply delta to data-store
+        if apply:
+            self._apply_all_delta(w_id, delta)
+        # Queue delta for subscription push
+        if send:
+            self._delta_store_to_queues(w_id, ALL_DELTAS, delta)
+
+        return delta
 
     def _get_status_msg(self, w_id: str, is_active: bool) -> str:
         """Derive a status message for the workflow.
@@ -552,7 +605,7 @@ class DataStoreMgr:
                 client=info['req_client'],
                 command='reqpub',
                 log=self.log,
-                args={'topic': topic, 'payload': 'pong'},
+                args={'topic': topic, 'payload': 'status'},
             )
             for w_id, info in self.workflows_mgr.workflows.items()
             if (
@@ -560,7 +613,7 @@ class DataStoreMgr:
                 info.get('req_client')
                 and w_id in self.w_subs
                 # BACK COMPAT
-                and info.get(CFF.VERSION) > '8.6.5'
+                and info.get(CFF.VERSION) >= '8.6.6'
             )
         }
         results = await asyncio.gather(
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -119,7 +119,7 @@ class DataStoreMgr:
         self.log = log
         self.data = {}
         self.w_subs: Dict[str, WorkflowSubscriber] = {}
-        self.topics = {ALL_DELTAS.encode('utf-8'), b'shutdown'}
+        self.topics = {ALL_DELTAS.encode('utf-8'), b'shutdown', b'ping'}
         self.loop = None
         self.executor = ThreadPoolExecutor(max_threads)
         self.delta_queues = {}
@@ -178,6 +178,10 @@ class DataStoreMgr:
             return
 
         self.delta_queues[w_id] = {}
+
+        # Setup connection checker points
+        self.workflows_mgr.workflows[w_id]['reqres_time'] = time.time()
+        self.workflows_mgr.workflows[w_id]['pubsub_time'] = time.time()
 
         # Might be options other than threads to achieve
         # non-blocking subscriptions, but this works.
@@ -305,6 +309,13 @@ class DataStoreMgr:
             self._delta_store_to_queues(w_id, topic, delta)
             # close connections
             self._disconnect_workflow(w_id)
+            return
+        elif (
+            topic == 'ping'
+            and delta == b'pong'
+            and w_id in self.workflows_mgr.workflows
+        ):
+            self.workflows_mgr.workflows[w_id]['pubsub_time'] = time.time()
             return
         self._apply_all_delta(w_id, delta)
         self._delta_store_to_queues(w_id, topic, delta)
@@ -532,3 +543,31 @@ class DataStoreMgr:
         else:
             # the workflow has not yet run
             return 'not yet run'
+
+    async def connections_checker(self):
+        """Check REQ/RES by requesting a PUB/SUB check."""
+        topic = 'ping'
+        requests = {
+            w_id: workflow_request(
+                client=info['req_client'],
+                command='reqpub',
+                log=self.log,
+                args={'topic': topic, 'payload': 'pong'},
+            )
+            for w_id, info in self.workflows_mgr.workflows.items()
+            if (
+                # skip stopped workflows
+                info.get('req_client')
+                and w_id in self.w_subs
+                # BACK COMPAT
+                and info.get(CFF.VERSION) > '8.6.5'
+            )
+        }
+        results = await asyncio.gather(
+            *requests.values(), return_exceptions=True
+        )
+        for w_id, result in zip(requests, results):
+            if isinstance(result, Exception):
+                continue
+            if result == topic:
+                self.workflows_mgr.workflows[w_id]['reqres_time'] = time.time()
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -194,10 +194,18 @@ class DataStoreMgr:
             # something went wrong, undo any changes to allow for subsequent
             # connection attempts
             self.log.info(f'failed to connect to {w_id}')
-            self.disconnect_workflow(w_id)
+            self._disconnect_workflow(w_id)
 
     @log_call
     def disconnect_workflow(self, w_id, update_contact=True):
+        """Terminate workflow subscriptions.
+
+        This is called externally i.e. by the workflow manager when the
+        workflow has stopped or changed state.
+        """
+        self._disconnect_workflow(w_id, update_contact)
+
+    def _disconnect_workflow(self, w_id, update_contact=True):
         """Terminate workflow subscriptions.
 
         Call this when a workflow has stopped.
@@ -245,7 +253,7 @@ class DataStoreMgr:
         """Purge the manager of a workflow's subscription and data."""
         # Ensure no old/new subscriptions exist on purge,
         # this shouldn't happen if disconnect is run before unregister.
-        self.disconnect_workflow(w_id, update_contact=False)
+        self._disconnect_workflow(w_id, update_contact=False)
         if w_id in self.data:
             del self.data[w_id]
         if w_id in self.delta_queues:
@@ -296,7 +304,7 @@ class DataStoreMgr:
         if topic == 'shutdown':
             self._delta_store_to_queues(w_id, topic, delta)
             # close connections
-            self.disconnect_workflow(w_id)
+            self._disconnect_workflow(w_id)
             return
         self._apply_all_delta(w_id, delta)
         self._delta_store_to_queues(w_id, topic, delta)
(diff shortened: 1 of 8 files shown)


dwsutherland:
Will sort tests in morning .. I could just change socket testing from testing PUB to REQ/RES server

cylc/uiserver/workflows_mgr.py
```diff
@@ -217,15 +229,57 @@ async def _workflow_state_changes(self):
...
+                )
             ):
-                # this flow is running but it's a different run
+                # This flow is running but it's a different run.
+                # Note: UUID can be the same for a workflow restarted using
+                # the same DB. So check process ID (PID).
+                # Although it's possible for a restarted workflow to regain
+                # the same PID, it's highly unlikely. And more unlikely for
+                # both this and the below PUB listener check to also be on the
+                # same ports.
```
oliver-sanders: Aah, derp, that would explain it.

For double safety, we can also check `CFF.HOST`.

This part is going to be quick and easy to get in, it might be worth chipping it off from the rest of the PR which might require more head-scratching.

Note, the `CFF.UUID` is *always* the same when a workflow-run is restarted (it's preserved in the DB).

cylc/uiserver/workflows_mgr.py
```diff
@@ -217,15 +229,57 @@ async def _workflow_state_changes(self):
...
+                        yield (wid, '/inactive', 'inactive', flow)
+                except SchedulerAlive:
+                    # old contact file exists and the workflow process still
+                    # exists.. purge and reestablish anyway.
+                    active.add(wid)
+                    if wid in active_before:
+                        yield (wid, '/active', 'active', flow)
+                    else:
+                        yield (wid, '/inactive', 'active', flow)
+
```
oliver-sanders: This part is going one step further than #845 and getting the UIS to proactively detect crashed workflows, which is a sensible thing to do.

In cylc-flow we handle this using the `cylc scan --ping`  option which just adds a basic GraphQL query to the pipe:

https://github.com/cylc/cylc-flow/blob/26a5aaac7181459c470814cbad8785fea99e7788/cylc/flow/scripts/scan.py#L490-L493

When the scan happens, any issues with the GraphQL request (incl comms-timeout) should activate the `detect_old_contact_file` logic, which should delete the contact file, which should get detected by the UIS on the next scan.

I.e, Running `cylc scan --ping` *should* be enough to get the UIS to detect a crashed workflow. I **think** this is still the case, so I **think** we don't need *this* explicit block of code (with all the `active -> inactivate` state transition handling) as it **should** be covered by existing logic. We would just need to add the GraphQL ping / `listener_exists || detect_old_contact_file` call onto the scan pipeline (rather than expecting the user to run `cylc scan --ping` manually) and the rest would follow naturally.

BUT, unfortunately, the `detect_old_contact_file` logic has been quite fiddly and caused a number of problems, and might not work for all deployments (e.g, it assumes SSH access between UIS and FLOW hosts). So annoyingly this will require a little more thought. Will raise at our end.

Linking this to https://github.com/cylc/cylc-flow/issues/4858.

dwsutherland: Roger that, I'll undo the socket check then.. The scan does the deal with the contact files, after the timeout delay..

However I think testing the existing REQ/RES/PUB/SUB connections can probably get in..

dwsutherland:
Pushed a related fix .. The workflow manager wasn't disconnecting the workflow and cleaning it up properly when the workflow stopped, only the data-store manager was disconnecting properly.

This effected a fix I had in..

cylc/uiserver/workflows_mgr.py
dwsutherland: Ok, socket test removed..

dwsutherland:
I've amended the description, and added tasks (see above). These are reduced to:
- Add criteria checks (`PID`, `HOST`) in addition to `UUID`.
- Check both the REQ/RES PUB/SUB are still connected/working.

And I've removed the socket testing (as that's covered by the REQ/RES/PUB/SUB test) and the direct contact file deletion..
This PR now has a `cylc-flow` sibling: https://github.com/cylc/cylc-flow/pull/7355

I've testing the ping pong:
image

But am yet to add actual integration tests.. Will work on that this afternoon.

cylc/uiserver/workflows_mgr.py
```diff
@@ -178,6 +183,19 @@ async def _workflow_state_changes(self):
...
+            w_id
+            for w_id in inactive_before
+            if (
+                w_id in self.workflows
+                and self.workflows[w_id].get('req_client')
+            )
+        }
+        # ensure workflows get properly dealt with by the workflow manger
+        active_before |= w_stops
+        inactive_before -= w_stops
```
dwsutherland, line 222: This is to ensure that the workflow manager actually calls it's `_disconnect` method, and cleans up the workflow record on it's side.. (the `_disconnect` wasn't been called at all, so the `self.workflows[w_id].get('req_client')` was not being set to `None`)

cylc/uiserver/data_store_mgr.py
```diff
@@ -178,13 +182,21 @@ async def connect_workflow(self, w_id, contact_data):
...
     @log_call
     def disconnect_workflow(self, w_id, update_contact=True):
         """Terminate workflow subscriptions.
 
+        This is called externally i.e. by the workflow manager when the
+        workflow has stopped or changed state.
+        """
+        self._disconnect_workflow(w_id, update_contact)
+
+    def _disconnect_workflow(self, w_id, update_contact=True):
```
dwsutherland, line 213: `_disconnect_workflow` can be called from multiple places, so only wanted `@log_call` to activate once (when the workflow manager cleans up it's records and calls it's own `_disconnect`)

cylc/uiserver/workflows_mgr.py
dwsutherland: Done

oliver-sanders:
Hi, I raised the question of proactive crashed workflow detection at our end...

* I don't think we can make use of the `detect_old_contact_file` check at the UIS end, or at least can't right now without a rethink.
* We have some half-baked plans to change approach:
  * Mark crashed workflows: https://github.com/cylc/cylc-flow/issues/4858
  * Workflow monitoring service: https://github.com/cylc/cylc-uiserver/issues/781
  * Which put together would handle this situation so the UIS doesn't have to.
  * These are chalked up for one of the prioritisation discussions that we've not got around to :cry:
* To proactively detect crashed workflows in the UIS (right now):
  * We could consider a ping approach (you beat me to it!).
  * This would be unable to differentiate between workflow crashed and workflow slow to respond / hanging.
  * So we can't declare the workflow stopped and remove the contact file.
  * But we can log an error event for the workflow (which would appear in the GUI) communicating this.
  * The message could potentially tell the user to run `cylc scan --ping` to poke at the workflows.
  * With more work, we could potentially give the user a button to request the UIS to do this for them (which should be fine as it wouldn't be required for functionality).

This is a very different issue to #845, and would require a cylc-flow interface, so we should split the #845 bugfix (target 1.9.x) from the proactive crash detection (target 1.10.0).

oliver-sanders:
@dwsutherland

I have extracted [the part of this PR](https://github.com/cylc/cylc-uiserver/pull/851#discussion_r3468512847) which addresses #845 into PR https://github.com/cylc/cylc-uiserver/pull/856 so we can get it into 1.9.3 quickly - will try to get that released today as there are Cylc Review fixes we need.

I've chalked up a new issue to cover the proactive detection of crashed workflows that the remainder of the PR addresses - https://github.com/cylc/cylc-uiserver/pull/856

There are some vague ideas around this, but we haven't managed to give it much time yet. One idea outlined was to get cylc-flow to `touch` it's own contact file with each main-loop iteration. This would allow us to tell how long it has been since the main-loop turned over which would separate network issues from scheduler issues. It still wouldn't be enough for us to tell if the scheduler has crashed (it could just be hanging for some perverse reason), but it would be enough for us to log an error for the user to pick up.

The UIS might want to do something more networky if there are scenarios where disconnecting/reconnecting to the same scheduler instance might be advantageous. E.g, if there is some unrecoverable ZMQ or networking issue and the connection needs to be reset?

dwsutherland:
> I don't think we can make use of the `detect_old_contact_file` check at the UIS end, or at least can't right now without a rethink.

I already removed this part from this PR (and the socket test).

> This would be unable to differentiate between workflow crashed and workflow slow to respond / hanging.

> The UIS might want to do something more networky if there are scenarios where disconnecting/reconnecting to the same scheduler instance might be advantageous. E.g, if there is some unrecoverable ZMQ or networking issue and the connection needs to be reset?

This is the primary purpose of the ping/pong, to detect and refresh/reconnect after possible networking/ZMQ issues.. Not to detect crashed workflows, perhaps that can be another effort.

If the workflow doesn't respond after two ping/pong attempts, then it is reset.. (set to `~3min30`)...


> I have extracted [the part of this PR](https://github.com/cylc/cylc-uiserver/pull/851#discussion_r3468512847) which addresses #845 into PR #856 so we can get it into 1.9.3 quickly - will try to get that released today as there are Cylc Review fixes we need.


Given my above responses, can we not just get this in now.
The ping/pong is very simple (just send and test that the workflow RES and PUB), back-compat is already accounted for too.

> This is a very different issue to https://github.com/cylc/cylc-uiserver/issues/845, and would require a cylc-flow interface

Also, the ping/pong alone _would_ address #845 , because the restarted workflow means the existing-UIS/old connections don't work anymore... So it would reestablish the workflow at the UIS.

Another thing, it's important for a (above mentioned) problem found at ESNZ, so also want this in ASAP.

oliver-sanders:
This ping solution requires a cylc-flow release so we can't get it in and released today.

<br />

I think there are some crossed wires here. The issue in #845 is about the contact file *changing* between scans, the ping-pong mechanism cannot detect this (it's checking connection timeout not scanning the contact file) but your [other change to the workflows mgr](https://github.com/cylc/cylc-uiserver/pull/851#discussion_r3468512847) does. So I've cherry-picked this change for quick merge and release.

<br />

> Another thing, it's important for a (above mentioned) problem found at ESNZ, so also want this in ASAP.

I don't think this issue requires anything more than #856 to fix, here's my test to reproduce:

* Start GUI.
* Start workflow.
* Refresh UIS scan -> workflow appears in GUI sidebar
* `kill -9` the workflow.
* `cylc scan --ping`
* Refresh UIS scan -> workflow changes to stopped, data store disconnects

Example GUI log:

```
[I 2026-06-30 09:52:51.686 CylcUIServer] [data-store] register_workflow('~oliver.sanders/foo/run1', True)
[I 2026-06-30 09:52:51.703 CylcUIServer] [data-store] connect_workflow('~oliver.sanders/foo/run1', <dict>)
[D 2026-06-30 09:53:12.296 ServerApp] 200 POST /cylc/graphql (ba9b516b6cb94071b85cd08789d4fd26@127.0.0.1) 3.43ms
[D 2026-06-30 09:53:28.411 ServerApp] 200 POST /cylc/graphql (ba9b516b6cb94071b85cd08789d4fd26@127.0.0.1) 5.77ms
[I 2026-06-30 09:53:28.417 CylcUIServer] [data-store] disconnect_workflow('~oliver.sanders/foo/run1')
```

The detects the removal of the contact file and disconnects the workflow as expected.

We can prove that it's working correctly by restarting the workflow and requesting a UIS scan - the UIS will then successfully reconnect.

(note this is a subtly different scenario to #845 as the workflow is detected as stopped in one scan and restarted in another)

<br />

> This is the primary purpose of the ping/pong, to detect and refresh/reconnect after possible networking/ZMQ issues.. Not to detect crashed workflows, perhaps that can be another effort.

I've not yet heard of issues with ZMQ connections, have you got any details on this or approaches to reproduce or methods to test?

I've opened a stub issue for this problem, please add any details you can: https://github.com/cylc/cylc-uiserver/issues/857

A ping-pong seems like a reasonable solution to this problem. Out of interest, why did you go down the route of implementing a bespoke low-level interface for this rather than using the basic GraphQL request interface used by `cylc scan --ping`?

cylc/uiserver/workflows_mgr.py
oliver-sanders: I don't understand this change, stopped workflows should already be getting detected.

Could you give me an example where this is a problem to help me understand and review?

cylc/uiserver/workflows_mgr.py
oliver-sanders: The `/active` state transition will cause:

1. disconnect
2. unregister
3. register
4. connect

We want the disconnect/reconnect, but we don't want the unregister/register (this yields pruned/added deltas). The workflow hasn't been deleted (unregistered), we're just re-attempting the connection (which may well fail).

Note that if the connection does fail, it will activate the `detect_old_contact_file` logic via the `timeout_handler`, so this is equivalent to automatically running `detect_old_contact_file` in a loop 🤦.

dwsutherland:
Rebased on top of #856, will add a way of addressing the ~"irresponsive workflow connection reset loop" soon, and possibly change/include to account for the ZMQ options in #857

dwsutherland:
Making some progress...

Have added handling for irresponsive workflows, it will now mark the workflow as such then send and apply a status update delta, which will be undone by either a reconnect or on receiving a "pong" (which is now the status of the workflow)..

This is useful if the problem is beyond ZeroMQ socket options/handling, perhaps a network outage and/or irresponsive workflow for any other reason...

Which can be tested/seen/reproduced with:
```
(uis) sutherlander@cortex-hyper:cylc-uiserver$ git diff
diff --git a/cylc/uiserver/app.py b/cylc/uiserver/app.py
index 9abb52a..705ad25 100644
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -348,7 +348,7 @@ class CylcUIServer(ExtensionApp):
             This involves sending workflows a REQ/RES request which prompts
             a PUB response captured by our SUB.
         ''',
-        default_value=180.0
+        default_value=15.0
     )
     max_workers = Int(
         config=True,
diff --git a/cylc/uiserver/data_store_mgr.py b/cylc/uiserver/data_store_mgr.py
index cddd712..0117a7f 100644
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -36,6 +36,7 @@ from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
 import json
 from pathlib import Path
+import random
 import time
 from typing import TYPE_CHECKING, Dict, Optional, Set, cast
 
@@ -300,9 +301,13 @@ class DataStoreMgr:
             topic == 'ping'
             and w_id in self.workflows_mgr.workflows
         ):
+            if random.randint(1,10):
+                return
+            print('PONG')
             self.workflows_mgr.workflows[w_id]['pubsub_time'] = time.time()
             # workflow is now clearly responsive
             if w_id in self.workflows_mgr.irresponsive:
+                print('RESPONSIVE')
                 # Bring status up to date
                 status_data = json.loads(delta.decode('utf-8'))
                 self.create_status_delta(
diff --git a/cylc/uiserver/workflows_mgr.py b/cylc/uiserver/workflows_mgr.py
index ab04254..ba76b7f 100644
--- a/cylc/uiserver/workflows_mgr.py
+++ b/cylc/uiserver/workflows_mgr.py
@@ -160,7 +160,7 @@ class WorkflowsManager:  # noqa: SIM119
 
         # Connection checker threshold, twice the check interval and buffer
         # to allow for two check attempts.
-        self.conn_threshold = uiserver.connections_check_interval * 2 + 30
+        self.conn_threshold = uiserver.connections_check_interval * 2 + 5
         # Naughty bin of irresponsive workflows
         self.irresponsive = set()
 
@@ -288,12 +288,13 @@ class WorkflowsManager:  # noqa: SIM119
                     status='irresponsive',
                     status_msg='Workflow is not responding to the UI Server.',
                 )
+                print('IRRESPONSIVE')
                 # using `~` to indicate disconnect/reconnect only
                 # if successful, the DSMgr will overwrite the above delta.
-                if wid in active_before:
-                    yield (wid, '~active', 'active', flow)
-                else:
-                    yield (wid, '~inactive', 'active', flow)
+                #if wid in active_before:
+                #    yield (wid, '~active', 'active', flow)
+                #else:
+                #    yield (wid, '~inactive', 'active', flow)
 
             else:
                 # this flow is running
```
image

image
(ignore the websocket closed warnings, I just closed a tab while a delta was being sent or something)

Will tidy up a bit more next week (and add socket options ZeroMQ side of things)

cylc/uiserver/workflows_mgr.py
dwsutherland: So `self.get_workflows()` calls `data_store_mgr.get_workflows()` which returns `active`/`inactive` workflows according to the status, if the status is `stopped` then `inactive`..
However, the status can be `stopped` in the `data_store_mgr` (because a shutdown delta or irresponsive workflow) but not inactive in the `workflow_mgr` (because the scan happened after).... So in order for the `workflow_mgr` to process `active => inactive` properly (i.e. actually disconnect) we have to mark such workflows as `active_before` and exclude from `inactive_before` according to their state in the `workflow_mgr`...

dwsutherland: Have added more detail to comment.

PR #853, opened 2026-06-26
review: rejig sorting

### LINKED ISSUE #807: review: sort order test unstable
We're getting a lot of flaky-failures for Cylc Review sort order tests, cause unknown.

oliver-sanders:
Slight rejig of sorting mechanism.

Hopefully fixes #807

**Check List**

### DIFF
--- a/cylc/review/review.py
+++ b/cylc/review/review.py
@@ -537,11 +537,11 @@ class CylcReviewService:
             data["entries"].sort(key=lambda entry: entry["name"], reverse=True)
         elif order == "time_asc":
             data['entries'] = self.sort_workflows(
-                data['entries'], reverse=True
+                data['entries']
             )
         else:  # order == "time_desc"
             data['entries'] = self.sort_workflows(
-                data['entries']
+                data['entries'], reverse=True
             )
 
         data["of_n_entries"] = len(data["entries"])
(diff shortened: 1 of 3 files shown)


cylc/review/review.py
```diff
@@ -1141,32 +1141,35 @@ def sort_workflows(workflows, reverse=False):
...
-                if self['last_activity_time'] and other['last_activity_time']:
-                    return (
-                        self['last_activity_time']
-                        > other['last_activity_time']
-                    )
-                elif other['last_activity_time'] or self['last_activity_time']:
-                    return (
-                        self['last_activity_time'] or '0'
-                        > other['last_activity_time'] or '0'
-                    )
```
oliver-sanders, line 1156: It's possible this was just lacking brackets:

```
>>> None or '0' > None or '0'
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'str' and 'NoneType'
>>> (None or '0') > (None or '0')
False
```

...

cylc/review/review.py
```diff
@@ -1141,32 +1141,35 @@ def sort_workflows(workflows, reverse=False):
...
+
         """
-        class Workflow(dict):
+        class Workflow:
+
+            __slots__ = ('name', 'last_activity_time')
+
+            def __init__(self, item):
+                self.name = item['name']
+                self.last_activity_time = item['last_activity_time'] or '0'
```
oliver-sanders, line 1153: ... but it's simpler to handle the default up-front.

MetRonnie (approved):
Fixes the sort order issue we encountered with one user.

Just missing changelog entry.

I would also be tempted to make this change to make the sorting less confusing?

```diff
diff --git a/cylc/review/review.py b/cylc/review/review.py
index 15012ee..d7f8946 100644
--- a/cylc/review/review.py
+++ b/cylc/review/review.py
@@ -538,9 +538,9 @@ class CylcReviewService:
         elif order == "time_asc":
             data['entries'] = self.sort_workflows(
-                data['entries'], reverse=True
+                data['entries']
             )
         else:  # order == "time_desc"
             data['entries'] = self.sort_workflows(
-                data['entries']
+                data['entries'], reverse=True
             )
 
@@ -1154,11 +1154,11 @@ class CylcReviewService:
             def __lt__(self, other):
                 if self.last_activity_time == other.last_activity_time:
-                    return self.name < other.name
-                return self.last_activity_time > other.last_activity_time
+                    return self.name > other.name
+                return self.last_activity_time < other.last_activity_time
 
             def __gt__(self, other):
                 if self.last_activity_time == other.last_activity_time:
-                    return self.name > other.name
-                return self.last_activity_time < other.last_activity_time
+                    return self.name < other.name
+                return self.last_activity_time > other.last_activity_time
 
             def __eq__(self, other):
diff --git a/tests/unit/test_review.py b/tests/unit/test_review.py
index f7d6d96..eca2645 100644
--- a/tests/unit/test_review.py
+++ b/tests/unit/test_review.py
@@ -102,5 +102,6 @@ def test_sort_workflows(workflows, sorted_names):
     """Workflow object sorting"""
     assert [
-        w['name'] for w in CylcReviewService.sort_workflows(workflows)
+        w['name']
+        for w in CylcReviewService.sort_workflows(workflows, reverse=True)
     ] == sorted_names
 
@@ -124,5 +125,6 @@ def test_sort_workflows_plausible():
     ]
     assert [
-        i['name'] for i in CylcReviewService.sort_workflows(workflows)
+        i['name']
+        for i in CylcReviewService.sort_workflows(workflows, reverse=True)
     ] == [
         '08-suites-page-2-3',
```

oliver-sanders:
> I would also be tempted to make this change to make the sorting less confusing?

@MetRonnie, that diff gives a different sort order because you've flipped the order for `last_activity_time` but not `name`.

The code is purposely sorting one field ascending and the other descending because we want the most recent workflows to float to the top, but for workflows with the same activity time to be sorted alphabetically a-z not z-a.

oliver-sanders:
@ChrisPaulBennett, quick review for you.

To test, you can compare:

```
conda run -n cylc-dev cylc review start
```

Against:

```
CYLC_VERSION=7 cylc review start
```

With this change, they should match.

MetRonnie:
> @MetRonnie, that diff gives a different sort order because you've flipped the order for `last_activity_time` but not `name`.
> 
> The code is purposely sorting one field ascending and the other descending because we want the most recent workflows to float to the top, but for workflows with the same activity time to be sorted alphabetically a-z not z-a.

Diff corrected.

Currently it is confusing that the "greater than" dunder method actually returns "less than" for the time field, which is the primary sorting field.

oliver-sanders:
> Currently it is confusing that the "greater than" dunder method actually returns "less than" for the time field, which is the primary sorting field.

Yeah, but I think the reason it's that way is because the default (for Cylc Review workflow listings) is `reverse=True`, so it would be confusing either way.

Will change it to preference, but not to fussed (just nuke it all!).

ChrisPaulBennett (approved):
Yup, they match

oliver-sanders:
* Pushed changelog.
* Applied Ronnie's patch.

PR #856, opened 2026-06-29
workflows mgr: fix detection of restarted crashed workflow

### LINKED ISSUE #845: restarted workflow not detected

oliver-sanders:
This is the fix @dwsutherland [came up with](https://github.com/cylc/cylc-uiserver/pull/851#discussion_r3468512847) for #845. I've chipped this part off of #851 as that PR is also attempting to solve #781.

Summary:

* Closes #845
* There was pre-existing logic to detect workflow runs which had been stopped, cleaned and restarted between scans (effectively resilience against failure of the "shutdown delta" mechanism or internal race conditions).
* However, crashed workflows weren't getting detected because they are the same workflow run (i.e, the UUID is unchanged).
* The rectification is the same in both cases, so this extends the UUID detection method to also inspect the PID and HOST which can be used to identify a scheduler instance.

To test:

```console
$ # open Cylc GUI

$ cylc vip --host=<host1>
$ # refresh workflow listing in Cylc GUI
$ # * workflow should be listed as running on host1

$ kill -9 <workflow-pid>
$ # refresh workflow listing in Cylc GUI
$ # * workflow should be listed as running on host1
$ # * not technically true but the UIS don't know better (#781 will handle this one day)

$ cylc play --host=<host2>
$ # refresh workflow listing in Cylc GUI
$ # * before this change the workflow will be listed as running on **host1** - wrong
$ # * after this change the workflow should be listed as running on **host2** - right (and should reconnect successfully)
```

**Check List**

- [~] Tests are included (or explain why tests are not needed).

### DIFF
--- a/changes.d/856.fix.md
+++ b/changes.d/856.fix.md
@@ -1,0 +1,1 @@
+Fixed detection of workflows that were restarted after having crashed.
--- a/cylc/uiserver/tests/test_workflows_mgr.py
+++ b/cylc/uiserver/tests/test_workflows_mgr.py
@@ -142,7 +142,7 @@ async def test_workflow_state_changes(tmp_path, active_before, active_after):
 
 
 @pytest.mark.parametrize(
-    'active_before,active_after',
+    'active_before, active_after',
     product([True, False], repeat=2)
 )
 async def test_workflow_state_change_uuid(
@@ -167,8 +167,11 @@ async def test_workflow_state_change_uuid(
     wid = Tokens(user=wfm.owner, workflow='a').id
 
     wfm.workflows[wid] = {
+        # NOTE: the mocked scan will return "42" for each of these fields
         CFF.API: API,
-        CFF.UUID: '41'
+        CFF.UUID: '41',
+        CFF.PID: '41',
+        CFF.HOST: '41',
     }
 
     if active_before:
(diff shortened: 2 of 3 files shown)


MetRonnie (approved):
I've pushed a changelog

samuel-denton (approved):
LGTM

oliver-sanders:
One valid comment (already known), one dumb.

MetRonnie:
@oliver-sanders I have just realised, this is now causing data-store `register_workflow()` every time a workflow is restarted normally. I'm not sure that was intended?

oliver-sanders:
> I have just realised, this is now causing data-store register_workflow() every time a workflow is restarted normally. I'm not sure that was intended?

If it's also calling "unregister" then that's definitely a problem.

It's it's just "re-registering" the workflow (note the contact file has changed so this may be necessary) then that's fine.

MetRonnie:
Yes, `unregister()` as well. See https://github.com/cylc/cylc-uiserver/pull/863

dwsutherland:
>  it's just "re-registering" the workflow

Yes, contact file changes on restart, so it should (disconnect if not already) unregister then register (then connect using the new host/port)

oliver-sanders:
It should re-register (i.e, update the contact data in an updated delta), but it shouldn't unregister (that's a pruned delta which infers that the workflow has been deleted which it hasn't).

PR #862, opened 2026-06-30
Fixed workflows showing stopped in GUI 

### LINKED ISSUE #695: workflow showing as stopped after restart
When workflows are restarted from the GUI, they don't always appear as running, even if you request a scan. But if you reload the GUI they show up as running straight away.

Probably either:
* A missing updated delta from the UIS.
* An internal state problem in the UI.

~Transfer this issue to cylc-ui if it turns out to be the latter.~

I've struggled to reproduce this locally, however, it reproduces intermittently when workflows are started on remote hosts.

samuel-denton:
Closes #695 
Moved GUI subscription removal to after _entire_workflow_update so when GUI subscription loop catches this workflow and re-subscribes, it gets the most up-to-date data including states.

Bug seemed to be a race condition where a workflow was de-subscribed before its status was updated; the subscription loop in cylc-flow could detect this and re-subscribe the workflow (triggering an 'initial_burst' snapshot of the data store) while/before status was updated, and therefore getting old data.  

The fix is to move the line de-subscribing the workflow to after the data store has been updated.  

I have not found any issues caused by this, but I think this needs some testing as it's a core behaviour being changed.

**Check List**

- [ ] Tests are included (or explain why tests are not needed). - Not really sure how to test this, but it would be nice

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -161,8 +161,6 @@ class DataStoreMgr:
         if w_id in self.w_subs:
             return
 
-        self.delta_queues[w_id] = {}
-
         # Might be options other than threads to achieve
         # non-blocking subscriptions, but this works.
         self.executor.submit(
@@ -173,6 +171,8 @@ class DataStoreMgr:
             contact_data[CFF.PUBLISH_PORT]
         )
         successful_updates = await self._entire_workflow_update(ids=[w_id])
+
+        self.delta_queues[w_id] = {}
 
         if w_id not in successful_updates:
             # something went wrong, undo any changes to allow for subsequent
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -161,8 +161,12 @@ class DataStoreMgr:
         if w_id in self.w_subs:
             return
 
-        # Might be options other than threads to achieve
-        # non-blocking subscriptions, but this works.
+        # Start the ZMQ subscription (in its own thread) first so that
+        # any scheduler deltas published while the entire workflow update
+        # (below) is in progress are not missed. These deltas wait,
+        # since _update_workflow_data() blocks until last_updated > 0.
+        # (Might be options other than threads to achieve
+        # non-blocking subscriptions, but this works.)
         self.executor.submit(
             self._start_subscription,
             w_id,
@@ -170,8 +174,12 @@ class DataStoreMgr:
             contact_data[CFF.HOST],
             contact_data[CFF.PUBLISH_PORT]
         )
+        # Populate the data store with the entire workflow state, setting
+        # last_updated > 0 so any queued deltas from step 1 can be applied.
         successful_updates = await self._entire_workflow_update(ids=[w_id])
 
+        # Only now clear the GraphQL subscribers' delta queues, so that they
+        # get a fresh "initial burst" snapshot of this up-to-date data store.
         self.delta_queues[w_id] = {}
 
         if w_id not in successful_updates:
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -163,8 +163,8 @@ class DataStoreMgr:
 
         # Start the ZMQ subscription (in its own thread) first so that
         # any scheduler deltas published while the entire workflow update
-        # (below) is in progress are not missed. These deltas wait,
-        # since _update_workflow_data() blocks until last_updated > 0.
+        # (below) is in progress are not missed. These deltas wait until
+        # _update_workflow_data() is done by looking for last_updated > 0.
         # (Might be options other than threads to achieve
         # non-blocking subscriptions, but this works.)
         self.executor.submit(
--- a/changes.d/862.fix.md
+++ b/changes.d/862.fix.md
@@ -1,0 +1,1 @@
+Fixed a bug causing some workflows to show as stopped in the GUI after a restart.


MetRonnie:
Can you reproduce the bug on 1.9.x? I cannot anymore...

Edit: might be because of https://github.com/cylc/cylc-uiserver/pull/856#issuecomment-4858697422, looks like an unintentional effect of that change has resulted in avoiding this race condition as well?

samuel-denton:
> Can you reproduce the bug on 1.9.x? I cannot anymore...
> 
> Edit: might be because of [#856 (comment)](https://github.com/cylc/cylc-uiserver/pull/856#issuecomment-4858697422), looks like an unintentional effect of that change has resulted in avoiding this race condition as well?

~~I was able to reproduce it just less consistently and it often seems to fix itself if left for a few minutes.  I can test that again tho and see if it's consistent.~~

~~Ah, not sure I had #856 merged when I was testing. Ill test again today~~

I can reproduce the bug and show this PR fixes it on the latest commit to 1.9.x

MetRonnie:
This does seem to fix the bug (when the commit is rebased onto #863 - just to be sure). Hats off to you because I have now sunk 2 days into looking at the data store manager & resolvers code and I'm still scratching my head!

I'm not sure the "initial_burst" is entirely relevant as I added debug logging on 1.9.x and saw initial bursts coming through for workflows that did not correctly start showing as running, as well as lack of initial bursts for workflows that did correctly start showing as running.

Currently we set an empty entry in the `delta_queues` dict during `register_workflow()` (which makes sense, it needs to be initialised for the workflow). As well as during `connect_workflow()` (which you have moved a few lines down), but it is unclear why this is needed in the first place?

Need @dwsutherland to take a look and explain this

samuel-denton:
> This does seem to fix the bug (when the commit is rebased onto #863 - just to be sure). Hats off to you because I have now sunk 2 days into looking at the data store manager & resolvers code and I'm still scratching my head!
> 
> I'm not sure the "initial_burst" is entirely relevant as I added debug logging on 1.9.x and saw initial bursts coming through for workflows that did not correctly start showing as running, as well as lack of initial bursts for workflows that did correctly start showing as running.
> 
> Currently we set an empty entry in the `delta_queues` dict during `register_workflow()` (which makes sense, it needs to be initialised for the workflow). As well as during `connect_workflow()` (which you have moved a few lines down), but it is unclear why this is needed in the first place?
> 
> Need @dwsutherland to take a look and explain this

I do not understand the bug or fix that well, it was a bit of an accident.  I was still trying to understand what each part of that function was doing and playing with the order of operations since that sort of inconsistent behaviour is often a race condition and that worked.  I then spent a few hours trying to work out why which is the explanation I gave :D 
I have since spent a while trying to better understand it.

The way I ended up working out what ```self.delta_queues[w_id] = {}``` actually did admittedly was asking co-pilot, which told me it was "clearing any GUI subscription registrations for a workflow, which would be picked up by the subscription loop in cylc-flow causing a re-subscription and an update to status, triggering an 'initial_burst'".  I think I agree with that summary having followed the loop in ```cylc-flow/cylc/flow/network/resolvers.py::BaseResolvers.subscribe_delta()``` but it might not be 100% and might be different terminology to what we use.
I think here:
```py
                for w_id in w_ids:
                    if w_id in self.data_store_mgr.data:
            -->         if sub_id not in delta_queues[w_id]:
                            delta_queues[w_id][sub_id] = deltas_queue
                            # On new yield workflow data-store as added delta
                            if args.get('initial_burst'):
                                delta_store = create_delta_store(
                                    workflow_id=w_id)
                                delta_store[DELTA_ADDED] = (
                                    self.data_store_mgr.data[w_id])
                                delta_store[DELTA_ADDED][
                                    WORKFLOW
                                ].reloaded = True
                                deltas_queue.put(
                                    (w_id, 'initial_burst', delta_store))
```

I have convinced myself this is correct having spent a long time going line by line.  Previously the data given by the initial_burst on re-subscription could have been stale, where now it will always be the latest data, since the data store may be outdated until _entire_workflow_update() has ran.  I think this is roughly correct:

```py
async def connect_workflow(self, w_id, contact_data):
    # 1. Start scheduler subscription thread, receives
    #    deltas from the scheduler but blocks until the data store
    #    is populated (waits for last_updated > 0, set by step 2).
    self.executor.submit(self._start_subscription, ...)

    # 2. Populate the data store: fetches the full workflow state from
    #    the scheduler and replaces self.data[w_id].
    #    Sets last_updated > 0
    await self._entire_workflow_update(ids=[w_id])

    # 3. Clear GUI subscriber registrations which triggers a
    #    re-register and snapshot the data store as a fresh initial_burst.
    self.delta_queues[w_id] = {}
```

If the initial_burst is triggered before _entire_workflow_update(), it may be fetching old data depending how long the update takes.

MetRonnie:
Ping @dwsutherland

dwsutherland:
The `self.delta_queues[w_id] = {}` in `register_workflow()` will push contact data essentially.. GraphQL subscriptions will do this by pushing an `initial_burst` when it finds `sub_id` is not an entry: https://github.com/cylc/cylc-flow/blob/9c45ad1bfb2dde80ef43b96895fd2cd42b74fa02/cylc/flow/network/resolvers.py#L556-L563

```py
. . .
                for w_id in w_ids:
                    if w_id in self.data_store_mgr.data:
                        if sub_id not in delta_queues[w_id]:
                            delta_queues[w_id][sub_id] = deltas_queue
                            # On new yield workflow data-store as added delta
                            if args.get('initial_burst'):
                                delta_store = create_delta_store(
                                    workflow_id=w_id)
. . .
```

The `self.delta_queues[w_id] = {}` in `connect_workflow()` will wipe out previous GraphQL subscription `sub_id` queue reference(s) for this workflow (for all subscriptions to this workflow), then they are added in that same above loop block causing a new `initial_burst`.

This PR makes sense.. Because the populating of the workflow at the UI Server (i.e. the data sync `flow => UIS`) intentionally subscribes first and then does the `_entire_workflow_update` in `connect_workflow()`:
https://github.com/cylc/cylc-uiserver/blob/9950599f06cd46a9690fd536191f19df609d8dcb/cylc/uiserver/data_store_mgr.py#L183-L192

It does this so it doesn't miss deltas that come in during the `_entire_workflow_update`.. These deltas won't make sense to the UI if they are pushed before the `_entire_workflow_update` push.. So we remove reference to the current GraphQL sub queues before.. 

However (as we've noted above), if the GraphQL sub sets the queue and initial burst after this "wipe out" and before `_entire_workflow_update` then you will get a blank workflow in the UI..

So resetting the GraphQL subscription queues after syncing the newly restarted workflow makes more sense.. Having non-matching deltas sent before and initial-burst after is more preferable than having useless initial-burst before and non-matching deltas forever (or until all the patchwork of deltas are coherent to the UI) after..

dwsutherland (approved):
Yes, this will work, as explained 👍

MetRonnie:
Why are "mismatched" deltas coming in in between starting the subscription and receiving the entire workflow update?

And another question: what is the idea behind the `delta_queues` dict in the first place? From debugging, every entry in the dict is identical for all workflows - a single-item dict where the `sub_id` key is the same for all workflows and the queue value is the same queue object for all workflows

image

dwsutherland:
> Why are "mismatched" deltas coming in in between starting the subscription and receiving the entire workflow update?

Because the ZeroMQ subscription is run in a separate thread. This is why we start the subscription:
```
 self.executor.submit( 
     self._start_subscription,
.
.
.
```
before `entire_workflow_update`, in case deltas come in while we're processing the requested workflow dump (so they aren't missed, and can be applied/sent after).
 
> And another question: what is the idea behind the `delta_queues` dict in the first place?

With a GraphQL subscription, the async gen needs a queue of deltas to work through..
Because the ZeroMQ subscriptions are run in separate threads we need a thread safe way of applying these workflow deltas to the data-store and peddling these off to whatever GraphQL subscriptions are interested.

> every entry in the dict is identical for all workflows -  

This would only be true if there's only one GraphQL subscription that subscribes to all workflows.

> a single-item dict where the `sub_id` key is the same for all workflows

If you have multiple GraphQL subscriptions you'd have multiple `sub_id`, and if some subscriptions are only interested in workflow `foo` then their `sub_id` will only be in `delta_queues["foo"]` not in `delta_queues["bar"]`..

> and the queue value is the same queue object for all workflows

The queues are unique to `sub_id`/the-graphql-sub..

In this way `delta_queues` is a (nice) way to have the GraphQL subscriptions manage what and when deltas they receive and send, independent of the data-store manager.. 
It also allows for the discussed mechanism of the data-store manager (DSMgr) signaling, to those GraphQL subscriptions who are interested, to re-register your queues (and `initial_burst` if the user has flagged this option).. 

If you were to say "why doesn't the DSMgr just push it directly?", then, aside from the separate thread problem, the DSMgr would have to not only manage the data-store, but also manage which GraphQL subscriptions are interested in which workflows/data (...etc). I'd rather let the GraphQL resolver manage this, it's the one that knows the details of the user request/sub/doc.

MetRonnie:
Thanks for the explanations, we're rather lacking in knowledge of this area of the codebase so it's really good to get some understanding :+1:


Going back to this point

> It does this so it doesn't miss deltas that come in during the `_entire_workflow_update`.. These deltas won't make sense to the UI if they are pushed before the `_entire_workflow_update` push.. So we remove reference to the current GraphQL sub queues before..
>
> However (as we've noted above), if the GraphQL sub sets the queue and initial burst after this "wipe out" and before `_entire_workflow_update` then you will get a blank workflow in the UI..

It sounds like the correct order should therefore be:

```python
successful_updates = await self._entire_workflow_update(ids=[w_id])

self.delta_queues[w_id] = {}

self.executor.submit(
    self._start_subscription,
    w_id,
    contact_data['name'],
    contact_data[CFF.HOST],
    contact_data[CFF.PUBLISH_PORT]
)
```

Seeing as `_entire_workflow_update` only sets an entry `self.data[w_id]` and does not seem to have anything to do with the deltas?

From my testing, I could not reproduce the bug with this order.

dwsutherland:
> It sounds like the correct order should therefore be

The only problem is you may miss would-be/incoming deltas between `_entire_workflow_update` and `self._start_subscription`.. That's why it's ordered the other way.

As it is, the deltas wait some time for the dump to be applied:
https://github.com/cylc/cylc-uiserver/blob/ca394eb2280f0deddba2b5cfc3fe761d7fe205db/cylc/uiserver/data_store_mgr.py#L286-L295

So deltas can be applied or discarded according to the time stamp on the dump (if the deltas time is after or before respectively).

There appears to be some concurrency between `connect_workflow` and the GraphQL sub/resolver, otherwise we wouldn't have this problem (unlike ZeroMQ subscription, it's not on a different thread (I think), just uses `asyncio` tools)..

I think the way this PR proposes is fine, the initial-burst/dump after will clean it up on the front end ... 

And **actually**, come to think of it, because of the above referenced delta delay/loop (waiting for the dump to be applied), and the fact that `last_updated` should already be zero, there shouldn't be any rubbish deltas being applied/sent.. So I think we're good.  (and I should walk back the related bit in the previous comments)

MetRonnie:
> The only problem is you may miss would-be/incoming deltas between `_entire_workflow_update` and `self._start_subscription`.. That's why it's ordered the other way.

I thought you were saying that deltas before the entire workflow update are mismatched, after all that is why you are resetting the delta queue after the entire workflow update?

dwsutherland:
> I thought you were saying that deltas before the entire workflow update are mismatched, after all that is why you are resetting the delta queue after the entire workflow update?

Ah sorry, they would be mismatched with the UI (if sent), but it looks like I had already handled that with the wait loop so not sent.

MetRonnie:
So if a delta comes through before the entire workflow update, it waits for the entire workflow update to have happened first before it will be sent to the UI. But then would the delta not be outdated, as the change it is telling the UI about was already included in the entire workflow update?

dwsutherland:
> So if a delta comes through before the entire workflow update, it waits for the entire workflow update to have happened first before it will be sent to the UI. 

Yes.. But more before the entire workflow update is processed.

> But then would the delta not be outdated, as the change it is telling the UI about was already included in the entire workflow update?

Not all deltas may be outdated, if the entire workflow update (dump) is created and sent and then a delta immediately pushed afterwards then the dump may still be being processed by the UIS when the UIS subscriber receives the delta (remember it's running on a separate thread and processing the dump involves loops ...etc). Also deltas are `PUB` and dump `RES`, which use to be on a separate thread but may still have concurrency.

And there's still a comparison of the data-store/dump and delta timestamps (set as a field on the flow end), with old deltas being discarded.

MetRonnie (approved):
I've pushed some code comments, @dwsutherland please merge if the comments are accurate

cylc/uiserver/data_store_mgr.py
```diff
@@ -161,17 +161,25 @@ async def connect_workflow(self, w_id, contact_data):
         if w_id in self.w_subs:
             return
 
-        # Might be options other than threads to achieve
-        # non-blocking subscriptions, but this works.
+        # Start the ZMQ subscription (in its own thread) first so that
+        # any scheduler deltas published while the entire workflow update
+        # (below) is in progress are not missed. These deltas wait,
+        # since _update_workflow_data() blocks until last_updated > 0.
```
dwsutherland:
```suggestion
        # (below) is in progress are not missed. These deltas wait until
        # _update_workflow_data() is done by looking for last_updated > 0.
```
It's not so much that `_update_workflow_data()` blocks (it's on a separate thread to the delta application), it's more that it releases the deltas by setting `last_updated`.

dwsutherland: Well.. Maybe `_update_workflow_data` should set `last_updated` to zero just before the request, usually it already is via `_update_contact` and I'm not sure of a situation where that hasn't run before connecting, so probably fine..

PR #863, opened 2026-07-02
Fix logic from #856

MetRonnie:
See https://github.com/cylc/cylc-uiserver/pull/856#issuecomment-4858697422

The change made in that PR treated changes to the PID or host as equivalent to changes to the UUID but that is not the case.
- UUID changed - workflow was cleaned and started
  - disconnect, unregister, reregister, reconnect
- PID/host changed - workflow was *re*started
  - disconnect, reconnect

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No changelog entry needed as minimal impact to users
- [x] Docs not needed

### DIFF
--- a/cylc/uiserver/data_store_mgr.py
+++ b/cylc/uiserver/data_store_mgr.py
@@ -36,15 +36,28 @@ from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
 from pathlib import Path
 import time
-from typing import TYPE_CHECKING, Dict, Optional, Set, cast
+from typing import (
+    TYPE_CHECKING,
+    Dict,
+    cast,
+)
 
+from cylc.flow.data_store_mgr import (
+    ALL_DELTAS,
+    DATA_TEMPLATE,
+    DELTAS_MAP,
+    EDGES,
+    WORKFLOW,
+    apply_delta,
+    create_delta_store,
+    generate_checksum,
+)
 from cylc.flow.exceptions import WorkflowStopped
 from cylc.flow.id import Tokens
 from cylc.flow.network.server import PB_METHOD_MAP
-from cylc.flow.network.subscriber import WorkflowSubscriber, process_delta_msg
-from cylc.flow.data_store_mgr import (
-    EDGES, DATA_TEMPLATE, ALL_DELTAS, DELTAS_MAP, WORKFLOW,
-    apply_delta, generate_checksum, create_delta_store
+from cylc.flow.network.subscriber import (
+    WorkflowSubscriber,
+    process_delta_msg,
 )
 from cylc.flow.workflow_files import (
     ContactFileFields as CFF,
@@ -56,7 +69,10 @@ from cylc.flow.workflow_status import WorkflowStatus
 from .utils import fmt_call
 from .workflows_mgr import workflow_request
 
+
 if TYPE_CHECKING:
+    from collections.abc import Container
+
     from cylc.flow.data_messages_pb2 import PbWorkflow
 
 
@@ -144,7 +160,7 @@ class DataStoreMgr:
         self._purge_workflow(w_id)
 
     @log_call
-    async def connect_workflow(self, w_id, contact_data):
+    async def connect_workflow(self, w_id: str, contact_data: dict) -> None:
         """Initiate workflow subscriptions.
 
         Call this when a workflow has started.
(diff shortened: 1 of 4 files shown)


samuel-denton (approved):
LGTM, logic is clear and the test covers the gap well

cylc/uiserver/tests/test_workflows_mgr.py
```diff
@@ -207,6 +209,56 @@ async def test_workflow_state_change_uuid(
...
 
+@pytest.mark.parametrize(
+    'pid, host, change_expected',
+    [
+        pytest.param('42', '42', False, id='no-change'),
+        pytest.param('1', '42', True, id='pid-change'),
+        pytest.param('42', '1', True, id='host-change'),
+    ],
+)
+async def test_workflow_state_change__killed(
```
samuel-denton, line 220: Is the double underscore intended?  I can see the pattern in a few other test files but it's not clear if those are accidental either.  I don't dislike using it to point to a state thats being tested, especially if its a set of tests going through different states, but it doesn't seem consistent if that is intended.

EG:
test_resolvers.py::test_clean__error(()
test_resolvers.py::test__schema_opts_to_api_opts()
test_data_store_mgr::test_entire_workflow_update__stopped_workflow()

MetRonnie: There isn't consistency on this but I tend to separate the name of the function being tested and the word(s) identifying the test case with a double underscore. (I'm probably not 100% consistent on that myself, though!)

If there's a double underscore at the start like `test__schema_opts_to_api_opts()` then it's probably because the function being tested starts with a single underscore i.e. `def _schema_opts_to_api_opts()`. Again there isn't consistency on this

cylc/uiserver/data_store_mgr.py
```diff
@@ -36,15 +36,29 @@
 from copy import deepcopy
 from pathlib import Path
 import time
-from typing import TYPE_CHECKING, Dict, Optional, Set, cast
+from typing import (
+    TYPE_CHECKING,
+    Container,
```
oliver-sanders: I'm guessing this covers `list[x]`, `tuple[x]`, `set[x]`, `dict[x]`, etc (but not generators which yield `x`) which makes sense.

Should probably use this more often.

But, from the docs:

> Deprecated alias to [collections.abc.Container](https://docs.python.org/3/library/collections.abc.html#collections.abc.Container).
>
> -- https://docs.python.org/3/library/typing.html#typing.Container

cylc/uiserver/workflows_mgr.py
```diff
@@ -215,62 +236,65 @@ async def _workflow_state_changes(self):
...
-                    # PID and HOST are (almost) unique to each scheduler
-                    # invocation
-                    or flow[CFF.PID] != self.workflows[wid].get(CFF.PID)
+                    # across reload/restart.
+                    # This workflow has been cleaned & started since last scan
+                    if wid in active_before:
+                        yield (wid, '/active', 'active', flow)
+                    else:
+                        yield (wid, '/inactive', 'active', flow)
+                    continue
```
oliver-sanders, line 252: I think this was simpler as `if/elif/else` logic than with all the `continue`s (but won't block for this).

cylc/uiserver/workflows_mgr.py
```diff
@@ -215,62 +236,65 @@ async def _workflow_state_changes(self):
...
-                if wid in active_before:
-                    yield (wid, '/active', 'active', flow)
-                else:
-                    yield (wid, '/inactive', 'active', flow)
-
+                ):
+                    # Process ID or host changes means workflow must have
+                    # restarted (contact file was left behind so we didn't
+                    # detect it as inactive earlier)
+                    yield (wid, 'active', 'active', flow)
```
oliver-sanders, line 260: :+1:

oliver-sanders (approved):

cylc/uiserver/data_store_mgr.py
MetRonnie: I'm aware of the deprecation, thought it seemed simplest to stick with `typing` for the foreseeable future, but it's probably slightly quicker to execute if switching to `collections.abc` I suppose

cylc/uiserver/workflows_mgr.py
MetRonnie: The logic requires the use of `continue` here at least https://github.com/cylc/cylc-uiserver/blob/10c635764c7e4ca6a797314787395b1d386d72a1/cylc/uiserver/workflows_mgr.py#L243-L261

(otherwise would need to repeat `if wid in self.workflows and ...` twice which I find rather unreadable in comparison)

PR #866, opened 2026-07-08
Remove --force-remote command

ChrisPaulBennett:
Now that the cat log changes (https://github.com/cylc/cylc-flow/pull/7309) are merged, we can now remove the `--force-remote` command from the ui-server

**Check List**


oliver-sanders (commented):
There are two cat-log commands run from the UIS, so we would expect a two line diff.

Note:

```
$ cylc cat-log --help
...
  -o, --force-remote    View remote logs remotely even if they have been
                        retrieved to the workflow host (default False).
```

ChrisPaulBennett:
I'll create another ticket on cylc-flow to remove the force-remote option.
But I've double checked, I can only see one use of `force-remote` in UIS. Where is the other one?

oliver-sanders:
The `--force-remote` option can remain in cylc-flow, we just don't need to use it here any more.

oliver-sanders (approved):
Plz add a changelog entry for this PR.

changes.d/866.fix.md
```diff
@@ -0,0 +1,2 @@
+Removed the `--force-remote` flag from Cylc Review `cat-log` calls so that log retrieval no longer forces remote file checking.
+
```
MetRonnie:
```suggestion
Removed the `--force-remote` flag from Cylc UI `cat-log` calls so that the log view no longer forces remote file checking.
```

MetRonnie (commented):
Just got a suggestion to improve the accuracy of the changelog entry

MetRonnie (approved):

PR #867, opened 2026-07-14
Fix logging rotation

### LINKED ISSUE #742: Log rotation not working correctly
The log rotation implemented in #376 appears to be flawed.
Logic was added which creates `log` as a symlink to `01-uiserver.log` (in `~/.cylc/uiserver/log/`) which is rotated at each start up.
So, we end up with a directory containing something like this:
```
01-uiserver.log
02-uiserver.log
03-uiserver.log
04-uiserver.log
05-uiserver.log
log -> /home/user/.cylc/uiserver/log/01-uiserver.log
```
However, the traitlets logging config is setup to use the file `log` with log rotation based on file size.
Therefore, if the log file gets large enough to trigger rotation (tested by reducing the size limit) we get this:
```
01-uiserver.log
02-uiserver.log
03-uiserver.log
04-uiserver.log
05-uiserver.log
log
log.1 -> /home/user/.cylc/uiserver/log/01-uiserver.log
```
After this, the log rotation at startup no longer works and output just gets appended to the file `log`.

### LINKED ISSUE #743: Log file issue with parallel runs
Users may have multiple UI server's running at once (e.g. one instance via Cylc Hub and another via `cylc gui --new` to test a new version). This can cause issues with the UI server log file.

To demonstrate the problem, run one instance of the UI server (`cylc gui --new`) and leave it running. It's log output appears in `01-uiserver.log`. Then start up and shutdown another UI server - you'll find the log output from the original server is now going to `02-uiserver.log`. Repeat this another 4 times and you'll find the log output from the original server is now going to a deleted file (depending on your filesystem, you may find it showing as a `.nfs` file).

I think a simple improvement would be to include the hostname in the log file name (the problem is most likely to occur when using multiple hosts with shared home directories). It may not be worth worrying about the case where you run multiple UI servers on the same host.

I suggest we look at this at the same time as #742.

MetRonnie:
Supersedes and contains two of the bug fixes from #789

Closes #742, closes #743

Fixed logging issues:
- Separated log files for UI servers run by `cylc gui` and JupyterHub - the latter now uses `hubapp-log` instead of `log`.
- Log file rotation was not always working correctly. Log file names now rotate as `log.1`, `log.2` etc. (or `hubapp-log.1`, `hubapp-log.2` etc.).

**Check List**

- [x] Tests are included 
- [x] Changelog entry included 
- [x] No docs needed

### DIFF
--- a/cylc/uiserver/logging_util.py
+++ b/cylc/uiserver/logging_util.py
@@ -14,59 +14,36 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from contextlib import suppress
-from glob import glob
-import logging
-import os
+from logging.handlers import RotatingFileHandler
 from pathlib import Path
 
-from typing import List
 
-from cylc.uiserver.app import USER_CONF_ROOT
+# Global var to track whether logging has been started before.
+# This is needed as Traitlets may reconfigure logging several times
+# within the lifetime of the application, which creates new instances
+# of the handler.
+log_started = False
 
 
-class RotatingUISFileHandler(logging.handlers.RotatingFileHandler):
-    """Rotate logs on ui-server restart"""
+class RotatingUISFileHandler(RotatingFileHandler):
+    """Rotate logs on ui-server restart or when reaching maxBytes."""
 
-    LOG_NAME_EXTENSION = "-uiserver.log"
+    def __init__(self, filename: str, *args, **kwargs):
+        global log_started
+        if log_started:
+            do_initial_rollover = False
+        else:
+            log_started = True
+            do_initial_rollover = Path(filename).is_file()
 
-    def __init__(self):
-        self.file_path = Path(USER_CONF_ROOT / "log").expanduser()
+        # Ensure the log directory exists:
+        Path(filename).parent.mkdir(parents=True, exist_ok=True)
 
-    def on_start(self):
-        """Set up logging"""
-        self.file_path.mkdir(parents=True, exist_ok=True)
-        self.delete_symlink()
-        self.update_log_archive()
-        self.setup_new_log()
+        super().__init__(filename, *args, **kwargs)
+        if do_initial_rollover:
+            self.doRollover()
 
-    def update_log_archive(self):
-        """Ensure log archive retains only 5 logs"""
-        log_files = sorted(glob(os.path.join(
-            self.file_path, f"[0-9]*{self.LOG_NAME_EXTENSION}")), reverse=True)
-        while len(log_files) > 4:
-            os.unlink(log_files.pop(0))
-        # rename logs, logs sent in descending order to prevent conflicts
-        self.rename_logs(log_files)
-
-    def rename_logs(self, log_files: List[str]):
-        """Increment the log number by one for each log"""
-        for file in log_files:
-            log_num = int(Path(file).name.partition('-')[0]) + 1
-            new_file_name = Path(
-                f"{self.file_path}/{log_num:02d}{self.LOG_NAME_EXTENSION}"
-            )
-            Path(file).rename(new_file_name)
-
-    def delete_symlink(self):
-        """Deletes an existing log symlink."""
-        symlink_path = Path(self.file_path / 'log')
-        if symlink_path.exists() and symlink_path.is_symlink():
-            symlink_path.unlink()
-
-    def setup_new_log(self):
-        """Create log"""
-        log = Path(self.file_path / f'01{self.LOG_NAME_EXTENSION}')
-        log.touch()
-        symlink_path = Path(self.file_path / 'log')
-        with suppress(OSError):
-            symlink_path.symlink_to(log)
+        # Remove any old naming scheme log files
+        for old_log in Path(filename).parent.glob('[0-9][0-9]-uiserver.log'):
+            with suppress(OSError):
+                old_log.unlink()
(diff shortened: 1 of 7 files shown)


cylc/uiserver/logging_util.py
```diff
@@ -13,60 +13,25 @@
...
-        with suppress(OSError):
-            symlink_path.symlink_to(log)
+class RotatingUISFileHandler(RotatingFileHandler):
+    """Rotate logs on ui-server restart or when reaching maxBytes."""
+
+    # Class attribute to track whether logging has been started before.
+    # This is needed as Traitlets may reconfigure logging several times
+    # within the lifetime of the application, which creates new instances
+    # of the handler.
+    started = False
```
oliver-sanders: This is a little icky, but I'm guessing we have no access to application lifecycle information (or even the application itself) at this point so needs must.

It might be clearer to implement this as a `global` variable as it otherwise looks like the pattern of setting variable default on the class to be shadowed on the application.

oliver-sanders (commented):
This generally looks good. The standalone and multi-user files are now neatly separated and the rollover is working nicely - no more blank files or other nonsense.

I'm happy with the removal of the symlink, getting it to rename the file instead is perfectly reasonable.

I spotted one issue, it will fall over with traceback if the `~/.cylc/uiserver/log` directory doesn't exist when the server is started. We should `mkdir -p` that at `__init__` time.

This does change the filenames to remove the `.log` extension which is unfortunate as this naming is now inconsistent with other Cylc logs. I'm guessing it changed because that's the logging handler default, but I think we can override the rotation naming via the `rotation_filename` method or `namer` attr?

I think as it stands, this will leave any pre-existing `<x>-uiserver.log` files kicking around (as they are now out of the rotation), which is annoying. I think we could solve this by restoring the original file naming (which would pull the old files back into the rotation), or potentially by shimming something into the rollover code to wipe out files with the previous name (if there's a desire to rename the file for other reasons).

MetRonnie:
> This does change the filenames to remove the `.log` extension which is unfortunate as this naming is now inconsistent with other Cylc logs. I'm guessing it changed because that's the logging handler default, but I think we can override the rotation naming via the `rotation_filename` method or `namer` attr?

The `*.1`, `*.2` filename pattern is [hard-coded](https://github.com/python/cpython/blob/027a8336891cf7dab6a437a5d4f9213a6a7828cd/Lib/logging/handlers.py#L166) into the `RotatingFileHandler` unfortunately, so it is not possible to alter without re-implementing the `doRollover()` method

MetRonnie:
> I spotted one issue, it will fall over with traceback if the `~/.cylc/uiserver/log` directory doesn't exist when the server is started. We should `mkdir -p` that at `__init__` time.

Done

oliver-sanders:
> The *.1, *.2 filename pattern is [hard-coded](https://github.com/python/cpython/blob/027a8336891cf7dab6a437a5d4f9213a6a7828cd/Lib/logging/handlers.py#L166) into the RotatingFileHandler unfortunately, so it is not possible to alter without re-implementing the doRollover() method

This seems to do the trick:

```python
import logging
import logging.handlers
from pathlib import Path
from time import time


class MyRotatingFileHandler(logging.handlers.RotatingFileHandler):
    def rotation_filename(self, default_name):
        """
        Modify the filename of a log file when rotating.

        This is provided so that a custom filename can be provided.

        The default implementation calls the 'namer' attribute of the
        handler, if it's callable, passing the default name to
        it. If the attribute isn't callable (the default is None), the name
        is returned unchanged.

        :param default_name: The default name for the log file.
        """
        name, number = default_name.rsplit(".", 1)
        path = Path(name)
        return str(path.parent / f"{number}-{path.name}")


LOG = logging.getLogger("a")

LOG.addHandler(MyRotatingFileHandler("logfile.log", maxBytes=100, backupCount=2))
LOG.setLevel(logging.INFO)

LOG.info(f"Hello world {time()} ...........................................")
```

But we do also have `cylc.flow.loggingutil.RotatingFileHandler` which it might make sense to use (not sure)?

MetRonnie:
Ah right, from a quick look at the `RotatingFileHandler` it appeared hard-coded but it was actually just passing in the filepath and number as a single string `<filepath>.<number>` instead of separately for some reason.

Your change would mean the first file is `uiserver.log`, then `01-uiserver.log`. Rather than try and work around this I am minded to stick with the current implementation, which also has the advantage of sorting single UIS and hub app logs separately.

oliver-sanders:
> Your change would mean the first file is uiserver.log, then 01-uiserver.log

I'm not sure what the problem with that would be (but haven't looked in depth).

<br />

> Rather than try and work around this I am minded to stick with the current implementation

We do need to change *something*, because if we don't pull pre-existing log files back into the rotation, they will persist indefinitely.


<br />

> which also has the advantage of sorting single UIS and hub app logs separately.

We can still manage the UIS and hubapp logs separately, configuring the naming pattern doesn't impact this?

MetRonnie:
> We do need to change _something_, because if we don't pull pre-existing log files back into the rotation, they will persist indefinitely.

The something I went with was your other suggestion at the time:

> or potentially by shimming something into the rollover code to wipe out files with the previous name

https://github.com/cylc/cylc-uiserver/blob/5b05c100dd1ac0530621ab5b475e1bd62ca408e8/cylc/uiserver/logging_util.py#L46-L49

> > which also has the advantage of sorting single UIS and hub app logs separately.
> 
> We can still manage the UIS and hubapp logs separately, configuring the naming pattern doesn't impact this?

```
$ ls
01-hubapp.log
01-uiserver.log
02-hubapp.log
02-uiserver.log
...
```
vs
```
$ ls
hubapp-log.1
hubapp-log.2
log.1
log.2
```

MetRonnie:
Also note from the original issue https://github.com/cylc/cylc-uiserver/issues/742 that if we went with `NN-uiserver.log` we would have the opposite problem of old `log.N` files left over. The `NN-uiserver.log` files are currently empty and the `log.N` are the files that actually contain the recent logs, so in that sense it is preferable to keep `log.N`.

ChrisPaulBennett (approved):
Sorry for the delay, I've never seen this code before and took me a while to understand what's going on here.

LGTM

PR #868, opened 2026-07-14
Fix `c.CylcUIServer.log_level` not working in jupyter config

MetRonnie:
Supersedes and contains one of the fixes from #789

Overriding `c.CylcUIServer.log_level` in `jupyter_config.py` was not working.

**Check List**

- [x] Tests are updated
- [x] Changelog entry included 
- [x] Docs not needed

### DIFF
--- a/changes.d/868.fix.md
+++ b/changes.d/868.fix.md
@@ -1,0 +1,1 @@
+Fixed a bug where overriding `c.CylcUIServer.log_level` in `jupyter_config.py` was not working.
--- a/cylc/uiserver/app.py
+++ b/cylc/uiserver/app.py
@@ -79,6 +79,7 @@ from pathlib import (
     Path,
     PurePath,
 )
+from pprint import pformat
 import sys
 from textwrap import dedent
 from typing import (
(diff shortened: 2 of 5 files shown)


MetRonnie:
(flaky cylc review test failure unrelated: https://github.com/cylc/cylc-uiserver/issues/807)

oliver-sanders:
Copying the context from the previous PR...

The issue with this approach is that all logging is supposed to be set by the logging config, this controls everything level, handlers, adapters, streams, etc as an atomic bundle which gets applied in one go. The logging config is a reactive trait which should be able to be edited and reapplied on the fly - modifying the entire setup when it is.

This PR works by hacking the logging (as set up by the logging config) in a way that will break its reactivity, but might corrupt its intention (config depending).

We should really be controlling this via the logging config rather than our application-level. I haven't yet invested the time into working out how that should be done.

Note, the `--debug` flag sets one value in that config which is all it is intended to do, it isn't really meant to cover user-defined logging config (only default console stream), but I do understand why you think it would be helpful if it did.

MetRonnie:
> The logging config is a reactive trait which should be able to be edited and reapplied on the fly - modifying the entire setup when it is.

"Should" being the operative word here. But there is no way for a user or admin to edit the UI Server log level in the logging config, without this workaround.

> This PR works ... in a way that will break [the logging config's] reactivity, but might corrupt its intention (config depending).

Can you explain this, I don't get what you're saying.

> We should really be controlling this via the logging config rather than our application-level. I haven't yet invested the time into working out how that should be done.

Is there a strong reason this workaround cannot be merged for now, until such a time as someone figures out the "correct" way to do this? (I'm happy to add a TODO comment to the code explaining that what I've done is a workaround.)

PR #869, opened 2026-07-14
Fix new error in mypy 2.3

MetRonnie:
> cylc/uiserver/handlers.py:359: error: Return type "Awaitable[None] | None" of "execute" incompatible with return type "Coroutine[Any, Any, Any]" in supertype "cylc.uiserver.graphql.tornado.TornadoGraphQLHandler"  [override]


samuel-denton (approved):
LGTM

PR #864, opened 2026-07-06
Use `jq` to improve Cylc Review tests

MetRonnie:
Make test failures less cryptic. Does not fix the flaky failure of `04-time-sort` (#807)

[jq](https://jqlang.org/manual/) is a useful command line tool for parsing JSON.

Edit: as it happened, the flaky test failed on this occasion: https://github.com/cylc/cylc-uiserver/actions/runs/28813257297/job/85446223206?pr=864#step:12:452

Before:
```
Traceback (most recent call last):
  File "<stdin>", line 21, in <module>
ValueError: ("[('entries', 2, 'name'), 'qux']", 'bar')
```

After:
```diff
--- -	2026-07-06 18:18:05
+++ /dev/fd/63	2026-07-06 18:18:05
@@ -1,5 +1,5 @@
 duration_run_desc
 baz
 foo
-qux
 bar
+qux
```

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] No dependency changes 
- [x] Tests are included 
- [x] Changelog entry not needed 
- [x] Docs not needed
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

Issue #846, opened 2026-06-18
review: remove Cylc 7 support


oliver-sanders:
Cylc review currently supports Cylc 7 workflows.

Cylc 7 is now long-retired, we will remove back-compat in due course but the time-frame is yet to be agreed.

MetRonnie:
We don't currently have any tests for Cylc 7 workflow DB support 😬

PR #872, opened 2026-07-23
Fix flaky test

### LINKED ISSUE #807: review: sort order test unstable

ChrisPaulBennett:
Closes #807 I believe the flaky test was caused by the `bar` job taking longer than the 2 secs due to server overload, causing the sorting to be different. I've increased the time difference between the two.
I'm not 100% sure this is the cause, but I can't get the test to fail anymore.

**Check List**


MetRonnie:
> I believe the flaky test was caused by the `bar` job taking longer than the 2 secs due to server overload, causing the sorting to be different

Yes it's likely to be the cause. However I outlined a change to eliminate any chance of flakiness (and speed up the test considerably): https://github.com/cylc/cylc-uiserver/pull/864#issuecomment-4902523157. It would be preferable to go down that route if not too inconvenient to attempt?

ChrisPaulBennett:
> > I believe the flaky test was caused by the `bar` job taking longer than the 2 secs due to server overload, causing the sorting to be different
> 
> Yes it's likely to be the cause. However I outlined a change to eliminate any chance of flakiness (and speed up the test considerably): [#864 (comment)](https://github.com/cylc/cylc-uiserver/pull/864#issuecomment-4902523157). It would be preferable to go down that route if not too inconvenient to attempt?

I'll take a look

MetRonnie:
Current test failures (also on master) are believed to be due to `suite.rc` support having been removed in cylc-flow. ~Addressing the failure would tackle https://github.com/cylc/cylc-uiserver/issues/846~ Edit: I misunderstood, we need to fix `suite.rc` support in Cylc Review, not remove it.

ChrisPaulBennett:
The test isn't anything to do cylc 7 back compat, so I've just ported it to Cylc 8

MetRonnie (approved):
Many thanks 🫡

PR #877, opened 2026-08-06
Added a depth limit when iterating through user suite files and restricted walking in _cylc_install/source

### LINKED ISSUE #829: review: recursively traverses any arbitrary symlinks in the cylc-run directory
### Description

Cylc review can hang for a long time while it goes through symlinks in the cylc-run dir, looking for workflows.

For example, if a workflow _source_ dir contains a symlink to a location with thousands of dirs (because cylc review traverses the `_cylc_install/source -> /path/to/source-dir` symlink and then traverses the aforementioned symlink).

Also, this can lead to a (finite) loop, leaving you with a mess on your cylc-review page:
```
~/cylc-run
`-- wflow
    |-- blargh -> ~/cylc-run
    `-- flow.cylc
```

### Expected Behaviour

Cylc review definitely should not traverse `_cylc_install/source`. But it should also probably respect `global.cylc[install]max depth`

Scott-Owen-James:
closes #829 

**Check List**

### DIFF
--- a/cylc/review/review.py
+++ b/cylc/review/review.py
@@ -42,6 +42,7 @@ from time import (
 )
 import traceback
 from urllib.parse import quote
+from pathlib import Path
 
 import cherrypy
 import jinja2
@@ -51,6 +52,7 @@ import markupsafe
 from cylc.flow import __version__ as CYLC_VERSION
 from cylc.flow.task_state import TASK_STATUSES_ORDERED
 from cylc.flow.workflow_files import WorkflowFiles
+from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.review.review_dao import (
     TASK_STATUS_GROUPS,
     CylcReviewDAO,
@@ -444,6 +446,22 @@ class CylcReviewService:
         except jinja2.TemplateError:
             traceback.print_exc()
 
+    # Filters out any os.walk results longer than max_scan_depth
+    @staticmethod
+    def safe_walking(user_suite_dir_root):
+        max_scan_depth = glbl_cfg().get(['install', 'max depth'])
+        for dirpath, dnames, fnames in os.walk(
+            user_suite_dir_root, followlinks=True
+        ):
+            relpath = Path(dirpath).relative_to(Path(user_suite_dir_root
+                                                     )).parts
+            if (len(relpath) > max_scan_depth):
+                continue
+            if any((relpath[x] == "_cylc_install" and relpath[x + 1] ==
+                    "source") for x in range(0, len(relpath) - 1)):
+                continue
+            yield (dirpath, dnames, fnames)
+
     @cherrypy.expose
     def suites(
         self,
@@ -505,8 +523,8 @@ class CylcReviewService:
             "share",
             "work"
         ]
-        for dirpath, dnames, fnames in os.walk(
-            user_suite_dir_root, followlinks=True
+        for dirpath, dnames, fnames in self.safe_walking(
+            user_suite_dir_root
         ):
             if dirpath != user_suite_dir_root and (
                 any(name in dnames or name in fnames for name in sub_names)
--- a/cylc/review/review.py
+++ b/cylc/review/review.py
@@ -41,6 +41,7 @@ from time import (
 )
 import traceback
 from urllib.parse import quote
+from pathlib import Path
 
 import cherrypy
 import jinja2
@@ -50,6 +51,7 @@ import markupsafe
 from cylc.flow import __version__ as CYLC_VERSION
 from cylc.flow.task_state import TASK_STATUSES_ORDERED
 from cylc.flow.workflow_files import WorkflowFiles
+from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.review.review_dao import (
     TASK_STATUS_GROUPS,
     CylcReviewDAO,
@@ -443,6 +445,23 @@ class CylcReviewService:
         except jinja2.TemplateError:
             traceback.print_exc()
 
+    # Filters out any os.walk results longer than max_scan_depth
+    @staticmethod
+    def safe_walking(user_suite_dir_root):
+        max_scan_depth = glbl_cfg().get(['install', 'max depth'])
+        for dirpath, dnames, fnames in os.walk(
+            user_suite_dir_root, followlinks=True
+        ):
+            relpath = Path(dirpath).relative_to(Path(user_suite_dir_root
+                                                     )).parts
+            if (
+                (len(relpath) > max_scan_depth) or
+                WorkflowFiles.Install.DIRNAME in relpath
+            ):
+                dnames[:] = []
+                continue
+            yield (dirpath, dnames, fnames)
+
     @cherrypy.expose
     def suites(
         self,
@@ -504,8 +523,8 @@ class CylcReviewService:
             "share",
             "work"
         ]
-        for dirpath, dnames, fnames in os.walk(
-            user_suite_dir_root, followlinks=True
+        for dirpath, dnames, fnames in self.safe_walking(
+            user_suite_dir_root
         ):
             if dirpath != user_suite_dir_root and (
                 any(name in dnames or name in fnames for name in sub_names)
--- a/cylc/review/review.py
+++ b/cylc/review/review.py
@@ -455,10 +455,10 @@ class CylcReviewService:
         ):
             relpath = Path(dirpath).relative_to(Path(user_suite_dir_root
                                                      )).parts
-            if (len(relpath) > max_scan_depth):
-                continue
-            if any((relpath[x] == "_cylc_install" and relpath[x + 1] ==
-                    "source") for x in range(0, len(relpath) - 1)):
+            if (
+                (len(relpath) > max_scan_depth) or
+                WorkflowFiles.Install.DIRNAME in relpath
+            ):
                 continue
             yield (dirpath, dnames, fnames)
 
--- a/cylc/review/tests/functional/08-suites-page.t
+++ b/cylc/review/tests/functional/08-suites-page.t
@@ -2,7 +2,7 @@
 # THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
 # Copyright (C) Earth Sciences New Zealand & British Crown (Met Office)
 # & Contributors.
-# 
+#
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
--- a/cylc/review/tests/functional/08-suites-page.t
+++ b/cylc/review/tests/functional/08-suites-page.t
@@ -1,7 +1,8 @@
 #!/bin/bash
 # THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
-# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
-# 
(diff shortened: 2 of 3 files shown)


tests/unit/test_review.py
```diff
@@ -133,3 +135,36 @@ def test_sort_workflows_plausible():
...
+            os.mkdir(directory)
+
+        depth = glbl_cfg().get(['install', 'max depth'])
+        assert len(CylcReviewService.safe_walking("TestDirectory")) == depth
+        with contextlib.suppress(Exception):
+            shutil.rmtree("TestDirectory")
+    except Exception as e:
+        with contextlib.suppress(Exception):
+            shutil.rmtree("TestDirectory")
+        raise e
```
oliver-sanders: Good job catching all these exceptions and triggering the cleanup logic with the `raise e` to preserve the `AssertionError`!

But there's good news here, `pytest` can provision temporary directories and handle the cleanup logic for us. https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

Scott-Owen-James: Implemented the recommendation.

cylc/review/review.py
```diff
@@ -444,6 +445,22 @@ def taskjobs(
...
 
+    # Filters out any os.walk results longer than max_scan_depth
+    @staticmethod
+    def safe_walking(user_suite_dir_root):
+        walkList = []
+        max_scan_depth = glbl_cfg().get(['install', 'max depth'])
+        for dirpath, dnames, fnames in os.walk(
+            user_suite_dir_root, followlinks=True
+        ):
+            if (len(dirpath.replace(user_suite_dir_root, "").split("/")) >
```
oliver-sanders: Use `os.path.split` rather than `.split('/')`.

There are various reasons for using `os.path.split`, one extreme example is that `/` in Paths is only a convention.

You might want to consider using `pathlib.Path` which has a `relative_to` function which can handle the `user_suite_dir_root` problem too.

cylc/review/review.py
```diff
@@ -444,6 +445,22 @@ def taskjobs(
...
+                max_scan_depth or
+                    "_cylc_install/source" in dirpath or
+                    dnames == "_cylc_install/source"):
```
oliver-sanders: Avoid direct string comparison of a path.

You might want to parse (i.e, the `.split` thing) the path first, you could then check if the last two parts of the path are `_cylc-install` and `source` rather than doing a direct string comparison.

cylc/review/review.py
```diff
@@ -444,6 +445,22 @@ def taskjobs(
...
+                continue
+            walkList.append((dirpath, dnames, fnames))
+        return walkList
```
oliver-sanders: Small point, it's more efficient to `yield` than to `list.append ; return` in this sort of situation.

With the `list.append` approach, we have to build a list with one entry for each item.

With the yield approach there's no list, the `for` loop is provided with each item, one at a time.

cylc/review/review.py
```diff
@@ -448,18 +449,20 @@ def taskjobs(
...
-        walkList = []
         max_scan_depth = glbl_cfg().get(['install', 'max depth'])
         for dirpath, dnames, fnames in os.walk(
             user_suite_dir_root, followlinks=True
         ):
-            if (len(dirpath.replace(user_suite_dir_root, "").split("/")) >
-                max_scan_depth or
-                    "_cylc_install/source" in dirpath or
-                    dnames == "_cylc_install/source"):
+            relpath = os.path.split(Path(dirpath
```
oliver-sanders: We don't have to `os.path.split` a `Path` object, that's something `pathlib` does for us:

```python
>>> from pathlib import Path
>>> Path('a/b/c')
PosixPath('a/b/c')
>>> Path('a/b/c').parts
('a', 'b', 'c')
```

cylc/review/review.py
```diff
@@ -448,18 +449,20 @@ def taskjobs(
...
+                                         ).relative_to(Path(user_suite_dir_root
+                                                            )))[0]
+            if (len(relpath) > max_scan_depth):
                 continue
-            walkList.append((dirpath, dnames, fnames))
-        return walkList
+            listConfig = [(relpath[x] == "_cylc_install" and relpath[x + 1] ==
+                          "source") for x in range(0, len(relpath) - 1)]
+            if True in listConfig:
+                continue
```
oliver-sanders: The Python `any` function comes in handy here.

```python
if any(x for x in ...):
    continue
```

cylc/review/review.py
```diff
@@ -444,6 +446,22 @@ def taskjobs(
...
+    def safe_walking(user_suite_dir_root):
+        max_scan_depth = glbl_cfg().get(['install', 'max depth'])
+        for dirpath, dnames, fnames in os.walk(
+            user_suite_dir_root, followlinks=True
+        ):
+            relpath = Path(dirpath).relative_to(Path(user_suite_dir_root
+                                                     )).parts
+            if (len(relpath) > max_scan_depth):
+                continue
+            if any((relpath[x] == "_cylc_install" and relpath[x + 1] ==
```
oliver-sanders, line 460: It's `_cylc-install` rather than `_cylc_install`.

This *Should* prevent the walk from looking inside the `_cylc-install` directory. So do we need the `source` check too?

Scott-Owen-James: The issue specified the path _cylc_install/source, I assumed because source may be too common to blacklist and we might want things in _cylc-install? as for the dash rather than an underscore, can easily rectify.

oliver-sanders: It's `_cylc-install`. To avoid the potential for typos, we hardcode all the file and path names in `cylc.flow.workflow_files` so we can use them as constants rather than writing them out every time:

```python
>>> from cylc.flow.workflow_files import WorkflowFiles
>>> WorkflowFiles.Install.DIRNAME
'_cylc-install'
```

Which get used like so:

https://github.com/cylc/cylc-flow/blob/c2f8e7ab3a7254c65bf0663566476f3239cc6bf4/cylc/flow/workflow_files.py#L605-L608

---

There is no need for us to scan inside of the `_cylc-install` directory as it cannot contain any workflows, so this should do the trick. I wouldn't expect the `_cylc-install/source` file to be scanned if we have successfully prevented it from looking inside of this directory, so I suspect we don't need the secondary check.

ChrisPaulBennett (approved):
Good use of `yield`.
Addresses the issue quite nicely.

oliver-sanders (changes requested):
The approach is good, but this needs a little tweak before it solves the issue.

Here's a little test I put together to try out this change. I created this workflow in my cylc-run directory, and pointed `source` to someone else's cylc-run directory:

```
my-test
├── _cylc-install
│   └── source -> /home/users/other.user/cylc-run
└── run1
    └── flow.cylc
```

When I run this with the previous version of `cylc review`, I can see the other user's workflows listed as `my-test/_cylc-install/source/<workflow-id>`. This replicates the original problem.

But, when I run `cylc review` on this branch, the page doesn't load and just hangs there. Note, I picked a user with a fair few workflows to try this out.

I added this `print` statement to see what it was doing and found that it was looking through the other user's workflows:

```diff
diff --git a/cylc/review/review.py b/cylc/review/review.py
index f246ae4..74ec158 100644
--- a/cylc/review/review.py
+++ b/cylc/review/review.py
@@ -452,6 +452,7 @@ class CylcReviewService:
         for dirpath, dnames, fnames in os.walk(
             user_suite_dir_root, followlinks=True
         ):
+            print(f'{dirpath=} {dnames=}')
             relpath = Path(dirpath).relative_to(Path(user_suite_dir_root
                                                      )).parts
             if (
```

Scott-Owen-James:
> The approach is good, but this needs a little tweak before it solves the issue.
> 
> Here's a little test I put together to try out this change. I created this workflow in my cylc-run directory, and pointed `source` to someone else's cylc-run directory:
> 
> ```
> my-test
> ├── _cylc-install
> │   └── source -> /home/users/other.user/cylc-run
> └── run1
>     └── flow.cylc
> ```
> 
> When I run this with the previous version of `cylc review`, I can see the other user's workflows listed as `my-test/_cylc-install/source/<workflow-id>`. This replicates the original problem.
> 
> But, when I run `cylc review` on this branch, the page doesn't load and just hangs there. Note, I picked a user with a fair few workflows to try this out.
> 
> I added this `print` statement to see what it was doing and found that it was looking through the other user's workflows:
> 
> ```diff
> diff --git a/cylc/review/review.py b/cylc/review/review.py
> index f246ae4..74ec158 100644
> --- a/cylc/review/review.py
> +++ b/cylc/review/review.py
> @@ -452,6 +452,7 @@ class CylcReviewService:
>          for dirpath, dnames, fnames in os.walk(
>              user_suite_dir_root, followlinks=True
>          ):
> +            print(f'{dirpath=} {dnames=}')
>              relpath = Path(dirpath).relative_to(Path(user_suite_dir_root
>                                                       )).parts
>              if (
> ```

Was afraid this might happen, the problem with the solution is that it still uses os.walk(), at which point it goes through every directory and just filters out the ones it doesn't want, meaning it will stall if that search is too large. You can more easily stall it by adding the line:
> print(len(list(os.walk(user_suite_dir_root, followlinks=True))))
at the start.

To resolve this I'll likely need to replace os.walk somehow?

cylc/review/review.py
```diff
@@ -505,8 +523,8 @@ def suites(
             "share",
             "work"
         ]
-        for dirpath, dnames, fnames in os.walk(
-            user_suite_dir_root, followlinks=True
+        for dirpath, dnames, fnames in self.safe_walking(
+            user_suite_dir_root
```
oliver-sanders, line 527: > To resolve this I'll likely need to replace os.walk somehow?

This problem is solvable using `os.walk`. Take a look through the code within this for loop for inspiration.

Scott-Owen-James: I'm still a little confused, are you referring to the problem in that you can walk to other people's directories, or it hanging because it's trying to read an extremely populous repository?

oliver-sanders:
LGTM

@ChrisPaulBennett, could you re-review, make sure to jam in some `print` statements to check the directories that the `os.walk` is iterating over (rather than the ones the safe walking method yields) to make sure it's not scanning the bits of the FS we don't want it to.

@Scott-Owen-James, the issue is against 1.9.x, but this PR is against master. Please could you rebase this branch, when rebasing backwards in time we have to delete the newer commits we no longer want, give me a yell and I'll walk you through it.

oliver-sanders:
@Scott-Owen-James, once you've rebased this branch onto 1.9.x (I think we've done that), you'll need to change the base branch of the PR to match.

To do this, scroll to the top of this PR, press the pencil icon next to the title, then change master to the appropriate branch.

Once that's done, the conflict should vanish.

Finally set the milestone for this PR.

Scott-Owen-James:
> @Scott-Owen-James, once you've rebased this branch onto 1.9.x (I think we've done that), you'll need to change the base branch of the PR to match.
> 
> To do this, scroll to the top of this PR, press the pencil icon next to the title, then change master to the appropriate branch.
> 
> Once that's done, the conflict should vanish.
> 
> Finally set the milestone for this PR.

Done

oliver-sanders:
@ChrisPaulBennett: Here's the instructions to reproduce - https://github.com/cylc/cylc-uiserver/pull/877#pullrequestreview-5018790487

oliver-sanders (approved):

ChrisPaulBennett (approved):
Appears to be behaving.

PR #878, opened 2026-08-06
cat-log tail-end mode

samuel-denton:
Trio of PRs to add a new tail mode to cat-log and make use of it in the GUI
 - cylc-flow: https://github.com/cylc/cylc-flow/pull/7414
 - cylc-uiserver: https://github.com/cylc/cylc-uiserver/pull/878
 - cylc-ui: https://github.com/cylc/cylc-ui/pull/2620

The three PR's will close 
 - https://github.com/cylc/cylc-uiserver/issues/802

Changes will be documented in cylc-doc: 
 - https://github.com/cylc/cylc-doc/pull/965

-----------------------------------------------------------------------------------

Added new mode for tailing the log file and altered the truncation warning to warn either the start or end is cut off.

The new mode relies on ```cat-log -t --tail-lines=MAX_LINES``` for the start position (MAX_LINES lines from end) where MAX_LINES is set in the UI.  It can pick up new lines followed from the end of the file.  

Out of scope for this PR, but I went through and removed any instances of typing.Optional from the 2 files this PR touches and removed the imports.

**Check List**

### DIFF
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -489,9 +489,7 @@ class Services:
 
                     # read in the log lines and add them to the buffer
                     line_count += 1
-                    # buffer.append(line)
-
-                    buffer.append(f"{line_count - 1}: {line}") # REMEMBER to revert this
+                    buffer.append(line)
 
                     if mode == TAIL_END and line_count - 1 == max_lines:
                         # we received exactly MAX_LINES lines -> the *start* of
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -52,6 +52,7 @@ from cylc.flow.data_store_mgr import WORKFLOW
 from cylc.flow.exceptions import CylcError
 from cylc.flow.id import Tokens
 from cylc.flow.network.resolvers import BaseResolvers
+from cylc.flow.scripts.cat_log import TAIL, TAIL_FROM_START
 from cylc.flow.scripts.clean import CleanOptions, run
 from cylc.flow.util import natural_sort_key
 
@@ -374,7 +375,9 @@ class Services:
                 await queue.put(exc)
 
     @classmethod
-    async def cat_log(cls, id_: Tokens, app: 'CylcUIServer', info, file=None):
+    async def cat_log(
+        cls, id_: Tokens, app: 'CylcUIServer', info, file=None, mode=TAIL
+    ):
         """Calls `cat log`.
 
         Used for log subscriptions.
@@ -382,7 +385,12 @@ class Services:
         cmd: List[str] = [
             'cylc',
             'cat-log',
-            '--mode=tail',
+            f'--mode={mode}',
+        ]
+        if mode == TAIL:
+            # only tail mode reads a fixed number of lines from the file
+            cmd.append(f'--tail-lines={MAX_LINES}')
+        cmd += [
             '--prepend-path',
             id_.id,
         ]
@@ -448,8 +456,9 @@ class Services:
 
                 else:
                     # there *are* lines to read from the cat-log process
-                    if line_count > MAX_LINES:
-                        # we have read beyond the line count
+                    if mode == TAIL_FROM_START and line_count > MAX_LINES:
+                        # we have read beyond the line count -> the *end* of
+                        # the file is truncated in tail-from-start mode
                         yield {'lines': buffer}
                         yield {
                             'error': (
@@ -479,7 +488,22 @@ class Services:
 
                     # read in the log lines and add them to the buffer
                     line_count += 1
-                    buffer.append(line)
+                    # buffer.append(line)
+
+                    buffer.append(f"{line_count - 1}: {line}") # REMEMBER to revert this
+
+                    if mode == TAIL and line_count - 1 == MAX_LINES:
+                        # we received exactly MAX_LINES lines -> the *start* of
+                        # the file is (probably) truncated in tail mode
+                        yield {'lines': list(buffer)}
+                        buffer.clear()
+                        yield {
+                            'error': (
+                                'The start of this file has been truncated'
+                                f' because it is over {MAX_LINES} lines long.'
+                            )
+                        }
+
                     if len(buffer) >= 75:
                         yield {'lines': list(buffer)}
                         buffer.clear()
@@ -639,13 +663,15 @@ class Resolvers(BaseResolvers):
         info: 'GraphQLResolveInfo',
         _command: str,
         ids: List[Tokens],
-        file=None
+        file=None,
+        mode=TAIL,
     ):
         async for ret in Services.cat_log(
             ids[0],
             self.app,
             info,
-            file
+            file,
+            mode,
         ):
             yield ret
 
@@ -701,6 +727,7 @@ async def stream_log(
     command='cat_log',
     id: str,  # noqa: required to match schema arg name
     file=None,
+    mode=TAIL,
     **kwargs: Any,
 ) -> AsyncGenerator[Any, None]:
     """Cat Log Resolver
@@ -717,6 +744,7 @@ async def stream_log(
         info,
         command,
         [tokens],
-        file
+        file,
+        mode,
     ):
         yield item
--- a/cylc/uiserver/resolvers.py
+++ b/cylc/uiserver/resolvers.py
@@ -52,7 +52,12 @@ from cylc.flow.data_store_mgr import WORKFLOW
 from cylc.flow.exceptions import CylcError
 from cylc.flow.id import Tokens
 from cylc.flow.network.resolvers import BaseResolvers
-from cylc.flow.scripts.cat_log import TAIL, TAIL_FROM_START
+from cylc.flow.scripts.cat_log import (
+    MIXED,
+    MIXED_TRUNCATION_PREFIX,
+    TAIL,
+    TAIL_FROM_START,
+)
 from cylc.flow.scripts.clean import CleanOptions, run
 from cylc.flow.util import natural_sort_key
 
@@ -376,20 +381,23 @@ class Services:
 
     @classmethod
     async def cat_log(
-        cls, id_: Tokens, app: 'CylcUIServer', info, file=None, mode=TAIL
+        cls, id_: Tokens, app: 'CylcUIServer', info, file=None, mode=TAIL,
+        max_lines: Optional[int] = None,
     ):
         """Calls `cat log`.
 
         Used for log subscriptions.
         """
+        # the maximum number of log lines to display before truncating
+        max_lines = max_lines or MAX_LINES
         cmd: List[str] = [
             'cylc',
             'cat-log',
             f'--mode={mode}',
         ]
-        if mode == TAIL:
-            # only tail mode reads a fixed number of lines from the file
-            cmd.append(f'--tail-lines={MAX_LINES}')
+        if mode in (TAIL, MIXED):
+            # these modes read a fixed number of lines from the file
+            cmd.append(f'--tail-lines={max_lines}')
         cmd += [
             '--prepend-path',
             id_.id,
@@ -456,16 +464,11 @@ class Services:
 
                 else:
                     # there *are* lines to read from the cat-log process
-                    if mode == TAIL_FROM_START and line_count > MAX_LINES:
+                    if mode == TAIL_FROM_START and line_count > max_lines:
                         # we have read beyond the line count -> the *end* of
                         # the file is truncated in tail-from-start mode
                         yield {'lines': buffer}
-                        yield {
-                            'error': (
-                                'This file has been truncated because'
-                                f' it is over {MAX_LINES} lines long.'
-                            )
-                        }
+                        yield {'truncated': 'end'}
                         break
 
                     line = await queue.get()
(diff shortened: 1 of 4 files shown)


cylc/uiserver/resolvers.py
```diff
@@ -479,7 +488,22 @@ async def cat_log(cls, id_: Tokens, app: 'CylcUIServer', info, file=None):
...
+                        # we received exactly MAX_LINES lines -> the *start* of
+                        # the file is (probably) truncated in tail mode
+                        yield {'lines': list(buffer)}
+                        buffer.clear()
+                        yield {
+                            'error': (
+                                'The start of this file has been truncated'
+                                f' because it is over {MAX_LINES} lines long.'
+                            )
+                        }
```
samuel-denton: Currently, it will give this truncation warning if viewing the log file live when it reaches MAX_LINES, however it will not actually truncate until the page is refreshed, navigation changes or the tail/head mode is toggled.  If we decide to pop the first lines off once MAX_LINES is reached, this is fine as is, but if not, we may want to find a way around this.

samuel-denton:
Test failing due to import not yet present in cylc-flow master.  Should resolve once the relevant cylc-flow PR (https://github.com/cylc/cylc-flow/pull/7414) is merged.

cylc/uiserver/resolvers.py
```diff
@@ -374,15 +378,25 @@ async def enqueue(
                 await queue.put(exc)
 
     @classmethod
-    async def cat_log(cls, id_: Tokens, app: 'CylcUIServer', info, file=None):
+    async def cat_log(
+        cls, id_: Tokens, app: 'CylcUIServer', info, file=None, mode=TAIL,
+        max_lines: Optional[int] = None,
```
MetRonnie, line 383: (Our min supported Python version is 3.12, so we don't use the deprecated `typing.Optional` etc anymore. NB there are still uses of it dotted around because we haven't gotten round to removing them yet)
```suggestion
        max_lines: int | None = None,
```

cylc/uiserver/resolvers.py
```diff
@@ -448,15 +462,11 @@ async def cat_log(cls, id_: Tokens, app: 'CylcUIServer', info, file=None):
...
+                        # we have read beyond the line count -> the *end* of
+                        # the file is truncated in tail (from-start) mode
                         yield {'lines': buffer}
-                        yield {
-                            'error': (
-                                'This file has been truncated because'
-                                f' it is over {MAX_LINES} lines long.'
-                            )
-                        }
+                        yield {'truncated': 'end'}
```
MetRonnie, line 469: (Pre-existing) The UI be able to handle both these coming in the same message, which would be more efficient
```suggestion
                        yield {
                            'lines': buffer,
                            'truncated': 'end',
                        }
```

cylc/uiserver/resolvers.py
```diff
@@ -480,6 +490,14 @@ async def cat_log(cls, id_: Tokens, app: 'CylcUIServer', info, file=None):
                     # read in the log lines and add them to the buffer
                     line_count += 1
                     buffer.append(line)
+
+                    if mode == TAIL_END and line_count - 1 == max_lines:
+                        # we received exactly MAX_LINES lines -> the *start* of
+                        # the file is (probably) truncated in tail-end mode
+                        yield {'lines': list(buffer)}
+                        buffer.clear()
+                        yield {'truncated': 'start'}
```
MetRonnie, line 499: I think the buffer clearing is only meant to happen if the buffer size is exceeded.

As the file is truncted, we should break here to end the GraphQL subscription.
```suggestion
                        yield {
                            'lines': buffer,
                            'truncated': 'start',
                        }
                        break
```

cylc/uiserver/resolvers.py
samuel-denton: Ah ok, I hadn't realise that was deprecated. Ill correct that and look for others.

samuel-denton: Found and corrected 5-6 examples of this.

samuel-denton:
Technically out of scope for this PR, but I went through and removed any instances of typing.Optional from the 2 files this PR touches and removed the imports.

cylc/uiserver/resolvers.py
samuel-denton: Nice, I didn't realise this worked.
I've found and updated 3 instances of it.

cylc/uiserver/resolvers.py
samuel-denton: Yep, this isn't needed.

samuel-denton: Ah wait actually it is needed.  The buffer will be under the max buffer size at the point we truncate, so it needs its own flush to prevent lines remaining in the buffer.  I think we would get duplicated log lines if this was removed.

samuel-denton:
The test failures don't make sense to me:

> ./cylc/uiserver/tests/test_resolvers.py:409:16: F821 undefined name 'timeout'
> ./cylc/uiserver/tests/test_resolvers.py:455:16: F821 undefined name 'timeout'

There is no timeout usage on those lines of that file.  Not sure if somehow the CI is checking out the wrong branch?
There is use of timeout in that file, offset by 11 lines for both warnings (398 and 444), but they look correct to me.

samuel-denton:
> The test failures don't make sense to me:
> 
> > ./cylc/uiserver/tests/test_resolvers.py:409:16: F821 undefined name 'timeout'
> > ./cylc/uiserver/tests/test_resolvers.py:455:16: F821 undefined name 'timeout'
> 
> There is no timeout usage on those lines of that file. Not sure if somehow the CI is checking out the wrong branch? There is use of timeout in that file, offset by 11 lines for both warnings (398 and 444), but they look correct to me.

Ah, found it.  There has been a commit to that file on master altering timeout:
https://github.com/cylc/cylc-uiserver/commit/109098805e99bbe053cea9d4ddc05860799df34a

So I need to merge master back in and update to asyncio.timeout.

samuel-denton:
Sorted flake8 tests, now failing a pytest due to an import from cylc-flow not existing.  I guess because it's not on master yet.
So I think I need to get my cylc-flow PR merged first, then re-run these tests.

cylc/uiserver/resolvers.py
```diff
@@ -610,6 +620,13 @@ async def cat_log(cls, id_: Tokens, app: 'CylcUIServer', info, file=None):
                     # read in the log lines and add them to the buffer
                     line_count += 1
                     buffer.append(line)
+
+                    if mode == TAIL_END and line_count - 1 == max_lines:
```
MetRonnie: Minor style thing: this is a bit of a "Yoda expression"
```suggestion
                    if mode == TAIL_END and line_count == max_lines + 1:
```

cylc/uiserver/resolvers.py
MetRonnie: Ah my bad, we shouldn't break as this is not the end of the subscription!

cylc/uiserver/resolvers.py
```diff
@@ -818,7 +839,7 @@ def kill_process_tree(
 
 
 async def list_log_files(
-    root: Optional[Any],
+    root: Any | None,
```
MetRonnie: (Not sure why this was `Optional[Any]` to begin with. `Any | None` is no different from `Any`.)

cylc/uiserver/schema.py
```diff
@@ -84,11 +84,11 @@
 
 
 async def mutator(
-    root: Optional[Any],
+    root: Any | None,
     info: 'GraphQLResolveInfo',
     *,
     command: str,
-    workflows: Optional[List[str]] = None,
+    workflows: List[str] | None = None,
```
MetRonnie:
```suggestion
    workflows: list[str] | None = None,
```

cylc/uiserver/schema.py
```diff
@@ -562,7 +562,7 @@ def run_task_query(conn, workflow):
 
 def _status_to_state(
     status: str
-) -> Tuple[Optional[int], Optional[int], Optional[bool]]:
+) -> Tuple[int | None, int | None, bool | None]:
```
MetRonnie:
```suggestion
) -> tuple[int | None, int | None, bool | None]:
```

cylc/uiserver/schema.py
```diff
@@ -966,6 +973,25 @@ class Logs(graphene.ObjectType):
...
+                f'"{TAIL_END}" (follow the last lines from the end of the '
+                'file).'
+            ),
+        ),
+        max_lines=graphene.Argument(
+            graphene.Int,
+            required=False,
+            description=(
+                'The maximum number of log lines to display before'
+                ' truncating the file.'
```
MetRonnie, line 992:
```suggestion
                f' truncating the file in {TAIL_END} mode.'
```

samuel-denton: max_lines applies to both modes.  

- Live workflows:
    - In TAIL mode (end truncated) it requests whats already present and then follows, up to max_lines, then stops
    - In TAIL_END (start truncated) it does the same, unless the log is already over max_lines, then it loads the last max_lines number of lines and then follows (currently without limit)
- Complete workflows:
    - In TAIL mode it just loads the first max_lines lines
    - In TAIL_END mode it just loads the last max_lines lines

cylc/uiserver/resolvers.py
samuel-denton: Ah yeah of course ```Any | None``` is the same as ```Any```...
I've corrected a few instances of that.

cylc/uiserver/schema.py
samuel-denton: That has reminded me that I was going to point [this](https://github.com/cylc/cylc-ui/pull/2620#issuecomment-5778606294) out. Discussion on TAIL_END not currently being bound by any limit on live workflows.

MetRonnie: That contradicts what is in your cylc-flow PR, I think
```python
    parser.add_option(
        "--tail-lines",
        help=(
            "For the tail-end mode, the number of lines to show from the"
            f" end of the file (default {DEFAULT_TAIL_LINES})."
            " Has no effect in other modes."
        ),
```

https://github.com/cylc/cylc-flow/blob/042392950af7410345a6a745df712a8bf08a806c/cylc/flow/scripts/cat_log.py#L398-L404

samuel-denton: Thats the --tail-lines flag from cat-log in cylc-flow.

However in the uiserver, max_lines is a different concept, used by both modes.

MetRonnie: Ah I was looking at https://github.com/cylc/cylc-uiserver/blob/b4899044e1618f908dcf3ac0a94db98f5317d871/cylc/uiserver/resolvers.py#L525-L527

but forgot about https://github.com/cylc/cylc-uiserver/blob/b4899044e1618f908dcf3ac0a94db98f5317d871/cylc/uiserver/resolvers.py#L594

cylc/uiserver/tests/test_resolvers.py
```diff
@@ -389,6 +390,100 @@ async def test_cat_log_timeout(workflow_run_dir, app, fast_sleep):
...
+            max_lines=3,
+        )
+        async for response in ret:
+            responses.append(response)
+            if response.get('truncated') == 'end':
+                info.context['sub_statuses'][3] = 'stop'
+            await asyncio.sleep(0)
+
+    lines = [
+        line
```
MetRonnie, line 424:
```suggestion
        line.rstrip('\n')
```

cylc/uiserver/tests/test_resolvers.py
```diff
@@ -389,6 +390,100 @@ async def test_cat_log_timeout(workflow_run_dir, app, fast_sleep):
...
+        response.get('truncated')
+        for response in responses
+        if response.get('truncated') is not None
+    ]
+
+    assert truncated == ['end']
+    assert len(lines) == 3
+    assert any('alpha' in line for line in lines)
+    assert any('beta' in line for line in lines)
+    assert any('gamma' in line for line in lines)
```
MetRonnie, line 438:
```suggestion
    assert lines == ['alpha', 'beta', 'gamma']
```

cylc/uiserver/tests/test_resolvers.py
```diff
@@ -389,6 +390,100 @@ async def test_cat_log_timeout(workflow_run_dir, app, fast_sleep):
...
+            max_lines=3,
+        )
+        async for response in ret:
+            responses.append(response)
+            if response.get('truncated') == 'start':
+                info.context['sub_statuses'][4] = 'stop'
+            await asyncio.sleep(0)
+
+    lines = [
+        line
```
MetRonnie, line 470:
```suggestion
        line.rstrip('\n')
```

cylc/uiserver/tests/test_resolvers.py
```diff
@@ -389,6 +390,100 @@ async def test_cat_log_timeout(workflow_run_dir, app, fast_sleep):
...
+        response.get('truncated')
+        for response in responses
+        if response.get('truncated') is not None
+    ]
+
+    assert truncated == ['start']
+    assert len(lines) == 3
+    assert any('three' in line for line in lines)
+    assert any('four' in line for line in lines)
+    assert any('DONE' in line for line in lines)
```
MetRonnie, line 484:
```suggestion
    assert lines == ['three', 'four', 'DONE']
```

samuel-denton: Nice improvements, applied to both tests.

PR #879, opened 2026-08-10
Locate cylc review tests in `cylc/review/` & fix lack of pytest

MetRonnie:
Does what it says on the commits.

**Check List**

- [x] Is tests - no changelog or docs needed
- [x] On 1.9.x as only affects tests and fixes tests not running when they should


hjoliver:
Looks fine, but can you explain why the `get_hostname` function in `test_header` isn't needed anymore?
```bash
#     get_hostname
#         Wraps logic to deal with differences between Darwin and Gnu Linux
#         $(hostname) on Github actions runners.
```

MetRonnie:
Not sure how old that was, but it doesn't exist in the cylc-flow `test_header`, and may predate the DNS issue (https://github.com/actions/runner-images/issues/8649).

Note for some reason the hostname of the MacOS runners are of the format `sat12-bq167-16cc0bc0-aab0-415b-a8f9-21a292138f0b-EA59D065D350.local`. There was a bit of logic that was trying to strip out the `.local` but I've removed that on the grounds it doesn't really matter, it's just for the `host` field in the JSON responses and the title in the main landing page of cylc review. Tests now passing.

oliver-sanders:
Why are we moving the functional tests into the Python package?

(These files are not pip-installable and are not located within cylc-flow in this way)

MetRonnie:
The tests for the `cylc.uiserver` package are located inside there. So this does the same for `cylc.review`.

The functional tests are not included in the build but [setuptools seems to deliberately include `tests/test*.py`](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html) for some reason (and has been doing this for `cylc/uiserver/tests/` the whole time: https://github.com/cylc/cylc-uiserver/actions/runs/31413200718/job/93535927685?pr=879#step:5:19). Using `prune` in the manifest file will hopefully do the trick (edit: yes, confirmed fixed in latest commit)

oliver-sanders:
I really don't get why we would want to put the functional tests inside the lib (in difference to how we do it in cylc-flow)?!

The `cylc/uiserver` unit tests are inside the lib for historical reasons (I.E, I never bothered to move them out). We could move the review tests to match `cylc/uiserver`, or we could move the `cylc/uiserver` tests to match `cylc/flow` and `cylc/rose` for consistency and no need to think about packaging?

So don't quite get the purpose of this, but if you're keen to do it, ok?

cylc/review/tests/functional/00-basic.t
```diff
@@ -50,12 +50,10 @@ grep_ok 'HTTP/.* 200 OK' "${TEST_NAME}.stdout"
 TEST_NAME="${TEST_NAME_BASE}-200-curl-root-json"
 run_ok "${TEST_NAME}" curl "${TEST_CYLC_WS_URL}/?form=json"
 
-get_hostname
-
 cylc_ws_json_greps "${TEST_NAME}.stdout" "${TEST_NAME}.stdout" \
     "[('cylc_version',), '$(cylc version | cut -d' ' -f 2)']" \
     "[('title',), 'Cylc Review']" \
-    "[('host',), '${HOSTNAME}']"
+    "[('host',), '$(hostname)']"
```
oliver-sanders, line 56: In metomi-rose we had to change this to `hostname -f` to get tests to pass on GHA AND local devices (due to differing network config).

https://github.com/metomi/rose/pull/2985/changes/4f918b882e7451ac476723440ccd1b2bcd838913

I'm not sure what we're actually testing here, but there's a chance it might be more portable to use `hostname -f` (it if actually works).

oliver-sanders: Ach, yes, the issue was not the `-f` the issue was a case transformation:

```diff
- sat12-dp153-ea219eec-38ec-4e56-9817-ea2534f89ef2-963cf56a9844
+ sat12-dp153-ea219eec-38ec-4e56-9817-ea2534f89ef2-963CF56A9844
```

https://github.com/cylc/cylc-uiserver/actions/runs/31407628290/job/93517699941#step:11:47

Which is a bit nuts, and likely a GH DNS issue :(

MetRonnie: I think it is meant to be the short hostname, as is displayed here:

image

MetRonnie:
> I really don't get why we would want to put the functional tests inside the lib (in difference to how we do it in cylc-flow)?!
> 
> The `cylc/uiserver` unit tests are inside the lib for historical reasons (I.E, I never bothered to move them out). We could move the review tests to match `cylc/uiserver`, or we could move the `cylc/uiserver` tests to match `cylc/flow` and `cylc/rose` for consistency and no need to think about packaging?
> 
> So don't quite get the purpose of this, but if you're keen to do it, ok?

I'm only keen to ensure we have consistency and are running all the tests (which we were not - we were missing the cylc review doctests, unit and integration tests!).

I weighed up moving the `cylc/uiserver/tests` to `tests/unit` but thought that it was nicer to make use of the existing separation of the `cylc/uiserver` and `cylc/review` directories, seeing as what is being tested are two completely different libraries in a shared monorepo.

cylc/review/tests/functional/00-basic.t
oliver-sanders: I'm not _sure_ it's the review-host, I think it might be for displaying the Cylc server a workflow is running on - just with some strange workflow titles?

MetRonnie: It is the review service host (try looking at the curl URLs in your browser for our site cylc review service)

oliver-sanders (approved):
Tests pass locally and on CI.

You might want to rename `cylc/review/tests/functional/cylc-review` to `cylc/review/tests/functional`.

MetRonnie:
Merging having done that

